PR #2, opened 2026-07-07
fix(tui): Keep picker columns aligned on waiting rows

epilande:
## Overview

Makes the row 1 attention label (pending tool, `Permission`, `Plan`) a budgeted part of the row layout and adds honest ellipsis truncation to the path:branch cell, so waiting rows no longer break the picker's column grid.

## Motivation

A session entering `waiting` gained a trailing label that no other row had. The label was injected outside the shared column structure, so flexbox took its width from the only shrinkable cell: the path. Three visible artifacts followed: the waiting row's path stopped lining up with its siblings, paths hard-clipped mid-word with no truncation marker (`epilande/ccmux` rendered as `epilandccmu`), and a dead whitespace band sat between the prompt and the label because the prompt's width budget didn't know the label existed.

## Screenshots

**Before**

Before

**After**

After

### Layout

- **SessionItem.tsx** - Renders the attention label and subagent count in a cell sized to exactly its own content; adds a `maxProjectLen` budget for the path cell; feeds the per-row label width into `maxPromptLen` so the prompt truncates right where the label begins and unlabeled rows keep the full width; the path cell gives its flex squeeze to the prompt in inline mode so identity survives at the message's expense

### Pure helpers

- **session-columns.ts** - Adds `getAttentionLabel` (moved from SessionItem), `subagentCountLabel`, `trailingLabelsWidth`, a 12-char label cap, and `fitProjectCell`, which fits prefix/dirname/branch to a width budget and marks every cut with `…` (or the existing `~` for branches) so the path can never silently drop characters

### Tests

- **SessionList.test.tsx** - Agent column alignment across waiting + idle rows; idle prompts extend into the space a sibling's label occupies
- **SessionItem.test.tsx** - Label ellipsis past the cap; long paths at narrow widths render `…` instead of mid-word clips
- **session-columns.test.ts** - Unit coverage for the shrink cascade's edge cases (exact fit, prefix drop, dirname floor, zero/negative budgets)

## Breaking Changes

None

## Testing

- [x] Unit tests pass (2232 tests, 110 files)
- [x] Live picker verified in a detached tmux session at 200/120/80 cols, both `inline` and `row2` prompt display modes
- [x] A real agent driven into `waiting [permission]` to exercise the label path end to end

### DIFF
--- a/src/tui/components/SessionItem.test.tsx
+++ b/src/tui/components/SessionItem.test.tsx
@@ -235,6 +235,39 @@ describe("SessionItem", () => {
     expect(frame).toContain("2 Agent");
   });
 
+  it("ellipsizes an attention label longer than the cap", async () => {
+    const frame = await renderItem(
+      {
+        session: mockEnrichedSession({
+          status: "waiting",
+          attentionType: "permission",
+          pendingTool: "Bash(git status --porcelain)",
+        }),
+      },
+      160,
+    );
+    // Capped to 12 chars with an ellipsis; the full tool string never shows.
+    expect(frame).toContain("…");
+    expect(frame).not.toContain("Bash(git status --porcelain)");
+  });
+
+  it("ellipsizes a long path instead of clipping it mid-word", async () => {
+    // At a narrow width the dirname no longer fits; it must show `…`, not a
+    // silent hard clip like "claude-tool".
+    const frame = await renderItem(
+      {
+        session: mockEnrichedSession({
+          cwd: "/Users/epilande/Code/epilande/claude-toolkit",
+          gitBranch: "main",
+        }),
+      },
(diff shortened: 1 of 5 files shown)


epilande:
**E2E demo**: one agent is driven `idle -> working -> waiting` while its sibling stays idle. When the `waiting` state lands, the pending-tool label (`Bash`) renders in its own budgeted cell, the `path:branch` column stays aligned across both rows, and the idle row's long prompt keeps flowing with `…` through the space the label occupies on the waiting row (no reserved dead band). Recorded at 160 cols, filtered to two throwaway demo sessions.

https://github.com/user-attachments/assets/b2584e6c-3798-4309-ba72-857d0a584254

PR #4, opened 2026-07-08
fix(daemon): Ensure config dir exists before opening daemon log

sperano:
Saw your post on reddit, seems like your tool is exactly what i need.  But it crashes on a fresh install (no `~/.config/ccmux/` yet), `spawnDaemonBackground` crashed with `ENOENT` when opening `ccmux.log`:

```
Starting daemon...
ENOENT: no such file or directory, open '/Users/eric/.config/ccmux/ccmux.log'
```

Create the log file's directory (recursive) before `openSync`, mirroring what `redirectStdioToLogFile` in `log-redirect.ts` already does.

### DIFF
--- a/src/daemon/lifecycle.ts
+++ b/src/daemon/lifecycle.ts
@@ -1,4 +1,12 @@
-import { readFileSync, unlinkSync, existsSync, openSync, closeSync } from "fs";
+import {
+  readFileSync,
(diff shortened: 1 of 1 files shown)


epilande (approved):
Oops can't believe I missed this. Thanks for fixing! LGTM!

PR #8, opened 2026-07-08
fix(daemon): use `lsof -Ffn` so cwd resolution works on lsof 4.99+

### LINKED ISSUE #7: [Bug]: cwd resolution fails on lsof 4.99+ (`-Fn` omits `fcwd`), daemon tracks 0 sessions
## Summary

On systems where a newer `lsof` (4.99+) is first on the daemon's `PATH` — e.g. the Nix-provided `lsof`, or Homebrew's — ccmux resolves **no** process working directories, so **no session ever binds to a tmux pane** and the daemon tracks 0 sessions. `ccmux debug` shows every process under "Untracked" with `CWD lookup failed (lsof returned no result)`.

## Root cause

`batchGetProcessCwds` in `src/daemon/processes.ts` runs:

```ts
Bun.spawn(["lsof", "-p", pidList, "-Fn"], …)
```

and parses the output expecting an `fcwd` fd-type line followed by the `n<path>` line:

```ts
} else if (line === "fcwd") {
  isCwd = true;
} else if (isCwd && line.startsWith("n") && currentPid !== null) {
  results.set(currentPid, line.slice(1));
  …
```

But `-Fn` selects **only** the `n` (name) field. Whether the `f` (fd) field is *also* emitted depends on the lsof version:

- **lsof 4.99+** (e.g. Nix `lsof-4.99.6`): `-Fn` emits `p…` and `n…` lines only — **no `fcwd`** → `isCwd` never flips true → every cwd resolves to `undefined`.
- **older builds** (e.g. macOS system `/usr/sbin/lsof`): `-Fn` still includes the `f` field, so `fcwd` is present and parsing works.

That version split is why this only bites some setups.

### Reproduction

```console
$ /usr/sbin/lsof -p <pid> -Fn | grep -c '^fcwd'        # macOS system lsof
1
$ /nix/store/…/lsof-4.99.6/bin/lsof -p <pid> -Fn | grep -c '^fcwd'   # Nix lsof 4.99.6
0
$ /nix/store/…/lsof-4.99.6/bin/lsof -p <pid> -Ffn | grep -c '^fcwd'  # with the f field selected
1
```

Impact: `ccmux debug` → `Tracked Sessions (0 from daemon)`, every entry `CWD lookup failed (lsof returned no result)`, and the TUI shows nothing even though hooks fire and markers are written.

## Fix

Select the fd field explicitly with `-Ffn` (correct on both old and new lsof):

```ts
Bun.spawn(["lsof", "-p", pidList, "-Ffn"], …)
```

The other lsof call (`getLsofLines` in `src/daemon/index.ts`, used for `.jsonl` session-id resolution) only reads `n` lines and is unaffected.

Verified locally: this took the daemon from 0 tracked sessions to all live sessions correctly bound to their panes.

I have this fix in a branch and can send a PR.

bcotton:
Fixes #7.

## Problem

`batchGetProcessCwds` runs `lsof -p <pids> -Fn` and keys the working dir off an `fcwd` fd-type line. **lsof 4.99+** (e.g. the Nix build, or any setup where a newer lsof is first on `PATH`) emits *no* fd-type lines for a bare `-Fn`, so `fcwd` never appears, every process cwd resolves to `undefined`, and no session ever binds to a tmux pane — `ccmux debug` shows `Tracked Sessions (0 from daemon)` with every process under `CWD lookup failed (lsof returned no result)`. Older builds (macOS system lsof) include the `f` field even with `-Fn`, which is why it only bites some setups.

```console
$ /nix/store/…/lsof-4.99.6/bin/lsof -p <pid> -Fn  | grep -c '^fcwd'   # 0
$ /nix/store/…/lsof-4.99.6/bin/lsof -p <pid> -Ffn | grep -c '^fcwd'   # 1
$ /usr/sbin/lsof            -p <pid> -Fn  | grep -c '^fcwd'           # 1
```

## Fix

Select the fd field explicitly with `-Ffn` — correct on both old and new lsof. The `.jsonl`-matching call in `getLsofLines` only reads `n` lines and is unaffected.

Verified locally: took the daemon from 0 tracked sessions to all live sessions correctly bound to their panes.

### DIFF
--- a/src/daemon/processes.ts
+++ b/src/daemon/processes.ts
@@ -197,7 +197,13 @@ export async function discoverAgentProcesses(
 
 /**
  * Batch get working directories for multiple processes in a single lsof call
- * Parses lsof -Fn output format: p<pid>, fcwd, n<path>
+ * Parses lsof -Ffn output format: p<pid>, fcwd, n<path>
+ *
+ * The `f` field selector is required: lsof 4.99+ (e.g. the Nix build) emits no
+ * fd-type lines for a bare `-Fn`, so `fcwd` never appears and every cwd lookup
+ * silently fails (0 sessions ever get a cwd → nothing binds to a pane). Older
+ * builds (macOS system lsof) include `f` even with `-Fn`, which is why this
+ * only bites on some setups. `-Ffn` is correct on both.
  */
 async function batchGetProcessCwds(
   pids: number[],
@@ -208,7 +214,7 @@ async function batchGetProcessCwds(
   try {
     const pidList = pids.join(",");
     DaemonPerf.incSubprocessSpawn("lsof-cwds");
-    const proc = Bun.spawn(["lsof", "-p", pidList, "-Fn"], {
+    const proc = Bun.spawn(["lsof", "-p", pidList, "-Ffn"], {
       stdout: "pipe",
       stderr: "pipe",
     });


epilande (approved):
LGTM Thanks for fixing!

PR #9, opened 2026-07-08
feat: watch multiple Claude config dirs (CLAUDE_CONFIG_DIR) in one instance

### LINKED ISSUE #6: [Feature]: Watch multiple Claude config dirs (CLAUDE_CONFIG_DIR) in one instance
## Summary

ccmux only ever reads `~/.claude/projects`. Any Claude Code session started with a non-default `CLAUDE_CONFIG_DIR` (e.g. a separate personal/work account) writes its session transcripts to a *different* projects tree and is therefore invisible to ccmux — even though its hooks fire and its session-pid marker is written correctly.

I'd like ccmux to be able to watch **multiple Claude config/projects directories from a single instance**, the same way it already fans out across multiple agents.

## Repro

```bash
# wrapper for a second Claude account
CLAUDE_CONFIG_DIR=~/.claude-personal claude
```

- The SessionStart hook fires and writes `~/.config/ccmux/session-pids/claude-<id>.json` (verified). ✅
- But the transcript lands in `~/.claude-personal/projects/<encoded-cwd>/<id>.jsonl`, which ccmux never scans, so the session never surfaces in the TUI. ❌

## Root cause

`src/lib/config.ts`:

```ts
export const CLAUDE_DIR = join(homedir(), ".claude");        // hardcoded
export const PROJECTS_DIR = join(CLAUDE_DIR, "projects");
export const CODEX_DIR = process.env.CODEX_HOME ?? join(homedir(), ".codex");  // codex already honors an env override
```

There is no `CLAUDE_CONFIG_DIR` handling anywhere in the codebase, and `PROJECTS_DIR` is a single path baked into the Claude log adapter and the daemon.

## Proposed design — fan out Claude log roots like agents

An agent in ccmux is a `{ LogAdapter + LogWatcher }` pair feeding a shared `SessionManager` (`src/daemon/index.ts`). A second Claude config dir is not a new *agent* — it's the *same* agent with a second *log root*: same adapter, parser, `sessionFilePattern`, hook `agent_type`, and process matcher; only the root path differs.

So: instantiate **one `ClaudeLogAdapter` + `LogWatcher` per configured Claude projects dir**, all feeding the same `SessionManager`. `LogWatcher` needs no changes.

### Config surface (mirrors the existing `agents` overrides in `ccmux.json`)

```jsonc
// ~/.config/ccmux/ccmux.json
{ "claudeConfigDirs": ["~/.claude", "~/.claude-personal"] }   // default: ["~/.claude"]
```

Optionally an opt-in `"claudeAutoDiscoverConfigDirs": true` that globs `~/.claude*/projects`. I'd keep auto-discovery behind a flag rather than implicit.

### Touchpoints

- `src/lib/config.ts` — `PROJECTS_DIR` (single) → resolve a list of Claude project dirs (falling back to `[~/.claude/projects]`). Precedent: `CODEX_DIR`/`CODEX_HOME`.
- `src/daemon/adapters/claude/log-adapter.ts` — take the projects dir as a constructor arg instead of the hardcoded `logDirGlob = PROJECTS_DIR`.
- `src/daemon/index.ts` — build one adapter+watcher per dir; `Promise.all` their `.ready`/`.start()`. `buildLogPath()` and the `existsSync(PROJECTS_DIR)` guard in `migrateExistingSessions()` resolve across the dir list.
- `ccmux status`/`ccmux debug` — report all watched Claude roots.

### Edge cases (both already safe)

- **ID collisions:** Claude session IDs are UUIDs and `SessionManager` is keyed by session-id, so the same cwd across two accounts coexists fine.
- **Shared hook marker:** hooks write `agent_type:"claude"` for both; `HookManager.getAdapter("claude")` is process/marker-based, not dir-based, so one shared hook adapter stays correct. Transcript correlation happens by session-id via whichever watcher discovers the `.jsonl`.

## Workaround today

Symlink `~/.claude-personal/projects` → `~/.claude/projects` so both accounts write transcripts into the single tree ccmux watches (merges histories, which is the downside this feature avoids).

I'm happy to send a PR implementing the above.

bcotton:
Closes #6.

Lets one ccmux instance track Claude sessions from **multiple config dirs** — e.g. a work account in `~/.claude` and a personal one launched with `CLAUDE_CONFIG_DIR=~/.claude-personal`. Today ccmux only watches `~/.claude/projects`, so the second account's sessions never surface even though their hooks fire and markers are written.

## Approach

A second config dir isn't a new *agent* — it's the same adapter pointed at a second log root. So this fans out Claude log roots the way ccmux already fans out agents:

- `resolveClaudeConfigDirs()` / `resolveClaudeProjectDirs()` gather `~/.claude` plus any dirs from the new **`claudeConfigDirs`** preference and the `CLAUDE_CONFIG_DIR` env var (deduped, order-preserving).
- The daemon stands up one `ClaudeLogAdapter` + `LogWatcher` per projects tree, all feeding the shared `SessionManager`. The primary watcher stays authoritative for marker-driven, path-agnostic `processPath` routing; `buildLogPath` probes every tree.
- Default behavior is unchanged when nothing extra is configured.

## Hook install fan-out (per your review note)

`ccmux setup` hardcoded `~/.claude` for hooks. Now `install()`/`uninstall()` fan out across every configured config dir, each writing its own `hooks/` scripts and `settings.json` entries. `isInstalled()` still reflects the primary `~/.claude` (it drives the daemon's single global runtime mode — keying it off an unconfigured extra dir could flip the whole daemon to no-hooks), and a new `describeInstallAnomalies()` warns at startup / in `ccmux setup` status when a configured dir is still missing hooks.

## Config

```jsonc
// ~/.config/ccmux/ccmux.json
{ "claudeConfigDirs": ["~/.claude-personal"] }
```
```bash
ccmux config set claudeConfigDirs '["~/.claude-personal"]'
ccmux setup --agent claude   # installs hooks into every configured dir
ccmux daemon restart
```

## Tests

`bun run typecheck` clean. New unit tests for `resolveClaudeProjectDirs` and for hook install/uninstall/`isInstalled` fan-out across dirs. Full suite green except the same pre-existing env-only failures on `main` (native chokidar watch timing + missing generated plugin bundles). Verified end-to-end: a daemon with `claudeConfigDirs` set watches both trees and tracks a live `claudep` session whose `logPath` resolves under `~/.claude-personal/projects`.

Note: builds on the same base as #8 (the `lsof -Ffn` cwd fix); they're independent and can merge in either order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ ccmux config list
 | `groupBy`           | `project`, `cwd`, `session`, `window`, `none`                                | `project`          | How sessions are grouped in the TUI                             |
 | `promptDisplay`     | `inline`, `row2`, `off`                                                      | `inline`           | Prompt display: inline on row 1, its own row, or hidden         |
 | `backgroundAgents`  | `true`, `false`                                                              | `true`             | Show Claude background agents as rows (daemon restart required) |
+| `claudeConfigDirs`  | array of paths                                                               | `[]`               | Extra Claude config dirs to watch beyond `~/.claude` (daemon restart required; see [Multiple Claude Config Dirs](#-multiple-claude-config-dirs)) |
 | `searchPaneContent` | `true`, `false`                                                              | `true`             | Include captured pane content in TUI search                     |
 | `persistent`        | `true`, `false`                                                              | `false`            | Keep picker open after switching sessions (dashboard mode)      |
 | `sidebar.width`     | `10`–`80`                                                                    | `30`               | Sidebar pane width in columns                                   |
@@ -364,6 +365,31 @@ An unknown base name falls back to the default theme; an invalid hex value or un
 
 > [!NOTE]
 > ccmux paints no background fill, so theme colors sit on your terminal's own background. The light palettes (`catppuccin-latte`, `tokyo-night-day`, `gruvbox-light`, `rose-pine-dawn`) assume a light terminal; pair them with a light background. Every other palette assumes a dark one.
+
+### 🗂️ Multiple Claude Config Dirs
+
+Claude Code writes its session transcripts to `$CLAUDE_CONFIG_DIR/projects` (default `~/.claude/projects`). If you run more than one Claude account — e.g. a work login in `~/.claude` and a personal one launched with `CLAUDE_CONFIG_DIR=~/.claude-personal` — sessions from the non-default account land in a separate `projects` tree that ccmux doesn't watch by default, so they never show up in the TUI even though their hooks fire correctly.
+
+Add the extra config dirs to `claudeConfigDirs` and ccmux watches every `<dir>/projects` tree from a single daemon, the same way it fans out across agents:
+
+```jsonc
+// ~/.config/ccmux/ccmux.json
+{
+  "claudeConfigDirs": ["~/.claude-personal"]
+}
+```
+
+```bash
+ccmux config set claudeConfigDirs '["~/.claude-personal"]'
+ccmux daemon restart
+```
+
+Notes:
+
+- `~/.claude` is always watched; entries here are **additional** Claude config dirs (their `projects` subdir is watched). Paths may start with `~`.
+- The `CLAUDE_CONFIG_DIR` environment variable, if set, is honored too and added automatically.
+- Sessions are keyed by their (globally unique) session ID, so the same project opened under two accounts coexists without collision.
+- Install the Claude hooks (`ccmux setup`) for each config dir you want authoritative session matching in — hooks are resolved per `CLAUDE_CONFIG_DIR`.
 
 ## 🔗 Session Matching with Hooks
 
--- a/README.md
+++ b/README.md
@@ -384,12 +384,18 @@ ccmux config set claudeConfigDirs '["~/.claude-personal"]'
 ccmux daemon restart
 ```
 
+Then install hooks into every configured dir so sessions from each account are matched authoritatively:
+
+```bash
+ccmux setup --agent claude   # fans out to ~/.claude and every claudeConfigDirs entry
+```
+
 Notes:
 
 - `~/.claude` is always watched; entries here are **additional** Claude config dirs (their `projects` subdir is watched). Paths may start with `~`.
 - The `CLAUDE_CONFIG_DIR` environment variable, if set, is honored too and added automatically.
 - Sessions are keyed by their (globally unique) session ID, so the same project opened under two accounts coexists without collision.
-- Install the Claude hooks (`ccmux setup`) for each config dir you want authoritative session matching in — hooks are resolved per `CLAUDE_CONFIG_DIR`.
+- `ccmux setup` installs the hook scripts and `settings.json` entries into **all** configured Claude dirs. Configure `claudeConfigDirs` first, then run setup. If you add a dir later, re-run setup — the daemon warns at startup about any configured dir still missing hooks.
 
 ## 🔗 Session Matching with Hooks
 
--- a/README.md
+++ b/README.md
@@ -258,21 +258,21 @@ ccmux config get <key>
 ccmux config list
 ```
 
-| Key                 | Values                                                                       | Default            | Description                                                     |
-| :------------------ | :--------------------------------------------------------------------------- | :----------------- | :-------------------------------------------------------------- |
-| `iconStyle`         | `dot`, `emoji`, `nerdfont`, `none`                                           | `dot`              | Status icon style                                               |
-| `theme`             | `catppuccin-*`, `tokyo-night*`, `dracula`, `gruvbox-*`, `nord`, `rose-pine*` | `catppuccin-mocha` | TUI color theme (resolved at launch; see [Theme](#-theme))      |
-| `showPreview`       | `true`, `false`                                                              | `false`            | Show preview panel on launch                                    |
-| `previewWidth`      | `20`–`80`                                                                    | `40`               | Preview panel width (percentage)                                |
-| `command`           | any non-blank string                                                         | `claude`           | CLI command used for session restart                            |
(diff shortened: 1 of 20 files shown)


src/daemon/index.ts
```diff
@@ -624,8 +642,46 @@ export class Daemon {
...
+  // The primary `~/.claude/projects` watcher is built in the constructor;
+  // this only adds the extras.
+  private setupExtraClaudeWatchers(preferences: Preferences): void {
+    this.claudeProjectDirs = resolveClaudeProjectDirs(
+      preferences.claudeConfigDirs,
+    );
+    for (const dir of this.claudeProjectDirs) {
+      if (dir === PROJECTS_DIR) continue;
+      const adapter = new ClaudeLogAdapter(this.sessionManager, dir);
+      this.extraClaudeWatchers.push(
```
epilande, line 677: The extra watchers never get registered in `logWatchers` (keyed by `agentType`), so marker events all route to the primary via `getLogWatcher("claude")`. A secondary-dir session that was idle at boot gets its transcript unwatched, and on resume its marker hits the primary watcher and no-ops, so it stays invisible until a daemon restart. `handleMarkerRemoved` has the same issue and leaks the extra adapter's subagent watch.

Let's fan `handleMarkerAdded`/`handleMarkerRemoved` across `claudeWatchers()` (or key routing by config dir) so the owning watcher re-arms. One fix covers both.

src/lib/config.ts
```diff
@@ -9,6 +9,55 @@ export const PROJECTS_DIR = join(CLAUDE_DIR, "projects");
...
+ * `CLAUDE_CONFIG_DIR` env var (matches Claude's own resolution), then the
+ * `claudeConfigDirs` preference. Paths may start with `~`. The result is
+ * absolute and de-duplicated, order-preserving, so the default behavior is
+ * unchanged when nothing extra is configured.
+ */
+export function resolveClaudeConfigDirs(configDirs?: string[]): string[] {
+  const dirs = [CLAUDE_DIR];
+  const extraConfigDirs = [
+    ...(process.env.CLAUDE_CONFIG_DIR ? [process.env.CLAUDE_CONFIG_DIR] : []),
+    ...(configDirs ?? []),
```
epilande: `getPreferences()` returns unvalidated JSON, so a `"claudeConfigDirs": "~/.claude-personal"` typo (string, not array) gets spread into characters, and `install()` then writes `hooks/` + `settings.json` into `~`, `/`, and cwd. Let's guard the type:

```suggestion
    ...(Array.isArray(configDirs)
      ? configDirs.filter((d) => typeof d === "string")
      : []),
```

src/daemon/adapters/claude/hook-adapter.ts
```diff
@@ -85,9 +109,9 @@ export class ClaudeHookAdapter implements HookAdapter {
     }
 
     let settings: ClaudeSettings = {};
-    const settingsExisted = existsSync(SETTINGS_FILE);
+    const settingsExisted = existsSync(settingsFile);
     if (settingsExisted) {
-      settings = JSON.parse(readFileSync(SETTINGS_FILE, "utf-8"));
+      settings = JSON.parse(readFileSync(settingsFile, "utf-8"));
```
epilande, line 121: nit: This parse is unguarded inside the per-dir loop (same in `uninstallFromDir`), so one malformed `settings.json` throws out of the whole fan-out, leaving orphan scripts and skipping later dirs plus the marker cleanup below. Let's wrap each per-dir call in try/catch and continue.

README.md
```diff
@@ -365,6 +366,37 @@ An unknown base name falls back to the default theme; an invalid hex value or un
...
+
+```jsonc
+// ~/.config/ccmux/ccmux.json
+{
+  "claudeConfigDirs": ["~/.claude-personal"]
+}
+```
+
+```bash
+ccmux config set claudeConfigDirs '["~/.claude-personal"]'
```
epilande: nit: `claudeConfigDirs` isn't a `config set` key (array keys aren't CLI-settable, like `agents`), so this errors with `Unknown key`. The JSON block above already covers it, so let's drop the line:

```suggestion
```

src/daemon/index.ts
```diff
@@ -624,8 +642,46 @@ export class Daemon {
...
 
+  // A session's transcript may live under any watched Claude config dir
+  // (e.g. a `~/.claude-personal` account), so probe each `projects` tree and
+  // return the first that exists. Falls back to the primary tree's path when
+  // none is found yet (the file may not have been written), preserving the
+  // original single-root behavior.
   private buildLogPath(cwd: string, sessionId: string): string {
-    return join(PROJECTS_DIR, encodeProjectPath(cwd), `${sessionId}.jsonl`);
+    const rel = join(encodeProjectPath(cwd), `${sessionId}.jsonl`);
+    for (const dir of this.claudeProjectDirs) {
```
epilande: nit: This first-existing-wins probe locates second-account transcripts but is untested, so a regression would silently mislocate logs. Let's extract it into a pure helper and unit-test first-existing + fallback.

epilande (commented):
Thanks for the contribution! Left a few comments, the main one to sort before merging is the marker routing in `index.ts`. Overall looking good.

epilande:
Let's also fix CI, looks like there's unused exports

bcotton:
Addressed the above. Let me know if there are other issues. Loving the integration with tmux!

epilande (approved):
LGTM Thanks!

epilande:
Hey @bcotton, quick heads up: I've renamed the `claudeConfigDirs` preference to `additionalClaudeConfigDirs` in #12 so the name makes the behavior obvious that `~/.claude` is always watched and these are extra dirs on top. Apologies for the churn

PR #16, opened 2026-07-12
feat(tui): Float toasts as a top-right overlay

epilande:
## Summary

Rebuilds the toast as a floating card in the top-right corner, macOS notification style. It used to render as a flex row between the footer and the help overlay, which shifted the entire layout down a line every time a toast appeared and disappeared. It also blended into the app background (same `base` color, `subtext` foreground), making it easy to miss.

The new toast:

- Is absolutely positioned, so showing/hiding it never reflows the UI underneath
- Sits one row below the top edge and one column in from the right
- Uses the raised `surface` background with a muted `overlay` border, so it reads as a card without shouting
- Renders as the last child of the root box, so it paints above every other overlay (help, confirm dialog, context menus)

All colors come from the semantic theme palette, so it adapts to custom themes.

## Testing

- [x] `bun test` passes (2380 tests)
- [ ] Verified end to end with a real agent session (not applicable, TUI-only change)

## Notes for reviewers

Verified in a detached tmux session at 120x28, 100x25, and 50x20, triggering a real toast via the <kbd>f</kbd> (Hide Idle) toggle:

```
 ● Sessions (3/33) [active]
 ▼ antigravity-support (1)                        │ antigravity-support (┌──────────────┐
▎1 ◐ work anti…:work~+  okay sounds    cc  13s    │ ◐ 1 working          │ Hide Idle ON │
 ────────────────────────────────────────────────│ ─────────────────────└──────────────┘
```

- A flat borderless variant with half-block caps (`▄`/`▀`) for a half-row top offset was prototyped and rejected: the cap width math depends on `message.length`, which breaks with wide characters and `maxWidth` truncation, and the caps blank out half-cells behind them
- Toasts remain single-instance with no severity variants (errors and status share one style), same as before; a `variant` prop is a possible follow-up


epilande:
Screenshot of the shipped toast (top-right overlay, triggered via the demo hook during development):
toast-border

PR #19, opened 2026-07-12
fix(daemon): Show working while background subagents run

epilande:
## Summary

Sessions whose lead spawned background agents (the `Agent` tool, `taskKind: in_process_teammate`) rendered as **done**/idle in the picker while the agents were still working. The lead genuinely ends its turn after spawning (the tool acks in milliseconds), so by its own transcript it is idle, and ccmux had no other signal: three gaps compounded.

- **Discovery**: the status machine only recognized blocking `Task` tool_uses, so `hasActiveSubagent` was never set and the `subagents/` dir was never watched. The adapter now attaches on write activity in the subagents dir, re-evaluated every reconciler scan via a new `LogAdapter.onReconcileTick`. Parse-driven evaluation alone provably misses the window: parent parses stop at `end_turn`, and the first live e2e run failed exactly there.
- **Identification**: the subagent filename regex only matched bare-hex agent IDs; named agents (`agent-a<name>-<hex>.jsonl`) were silently ignored.
- **Completion**: background agents never write a terminal `end_turn` (observed live: `stop_sequence` and `null`), so a `capStaleSubagents` reconciler sweep infers completion from silence (`SUBAGENT_STALE_TIMEOUT_MS`, 3 min, shared by the attach probe and seed cap to prevent attach/teardown loops).

`getEffectiveStatus` lifts an idle parent to `working` while any subagent works, so every consumer (status badge, grouping, previews) renders the truth. Blocking `Task` subagents keep their existing lifecycle: their logs end with `end_turn`, so they self-evict immediately with no staleness wait.

## Testing

- [x] `bun test` passes (2400 tests; 14 new across the adapter, reconciler sweep, `getEffectiveStatus`, and StatusBadge)

E2E detail: restarted the daemon from this branch, drove a real Claude session that spawned two background agents and ended its turn. The picker showed `◐ working` with a `2 Agent` badge while the lead sat idle at its prompt, then settled back to `done` after the agents finished and the stale sweep evicted them. A screencast of the full arc is attached below.

## Notes for reviewers

- The `docs/architecture.md` "Subagent tracking (Claude)" section documents the two subagent lifecycles and the four design points; it is the fastest way to review the intent.
- Known tradeoff: silence-based completion means a finished background agent keeps the row at `working` for up to 3 minutes. `stop_reason: stop_sequence` looks like a viable fast-path terminal marker (observed live), queued as a follow-up.
- Also queued for a UI/UX pass: the group preview panel counts raw status (shows `1 idle` while the row is lifted to `working`), and a subagent sitting in one long Bash call derives `waiting` from its unresolved tool_use even under bypassPermissions (pre-existing derivation quirk, now visible since these agents are tracked at all).


epilande:
E2E screencast (sped up): the lead session spawns two background agents and ends its turn. The row stays `working` with a `2 Agent` badge while the agents run (previously it read `done`), then settles back to `done` after they finish and the stale sweep evicts them.

https://github.com/user-attachments/assets/46501b13-2df5-4011-8deb-186bda3ec001

PR #27, opened 2026-07-16
feat(notifications): Expand Claude notification actions

### LINKED ISSUE #25: [Feature]: Expand Claude notification actions (Reply on finished, deny with feedback, plan approval)
### Problem / motivation

PR #24 makes every notification line carry distinct information, but the action set is still v2-era: permission waits get Approve/Deny, Claude question waits get Reply, and everything else has nothing. That leaves value on the table:

- **Finished** is the highest-traffic notification and carries no actions. The pane is sitting at an idle composer, so "type your follow-up from the notification" is the most useful action of all and the cheapest to add.
- **Permission** denial can't carry steering feedback. Claude's permission prompt has a "No, and tell Claude what to do differently" path, which is exactly what replying to a permission notification means.
- **plan_approval** waits get neither buttons nor a context body. It is the only notification kind with no context at all.

### Proposed solution

Three features, one commit each, all Claude-scoped. Builds on the content layout from PR #24.

**Core design decision:** reuse the existing action ids (`approve`/`deny`/`answer`) everywhere and let `handleNotificationAction` branch on the live, token-validated session state instead of introducing new action ids.

- The macOS helper's category set is fixed and label-baked: `responderCategories()` (**notifier/Sources/main.swift:154-165**) registers exactly three shapes (`approve,deny`, `reply`, `approve,deny,reply`) with hard-coded labels. Every shape needed here is already registered, so this requires zero Swift changes and no helper release, as long as the ids stay `approve`/`deny`/`answer`.
- The staleness token is the same-turn identity: `statusChangedAt` bumps only on a status edge, the notifier stamps it into the payload, the helper echoes it opaquely. If the token matches, the session is still on the exact edge the notification described, so the handler can safely derive the reply kind from live `status`/`attentionType`.
- Caveat to document in the handler: a permission to question flip within `waiting` does not bump the token. Benign for Claude (both branches use the Escape prelude), but worth a comment.

#### 1. Reply on "finished" notifications (commit 1, land first)

- **src/daemon/notifier.ts** - attach `payload.reply = { id: "answer", label: "Reply" }` in the `finished` branch (439-444) when the agent def opts in via a new `notificationActions.replyOnFinished`
- **src/daemon/notification-action.ts** - the `answer` staleness gate (128-140) additionally accepts `idle` with a matching token; the idle branch sends NO prelude (Escape at Claude's idle composer clears a typed draft, and double-Escape opens the history-rewind picker), just `sendText(pane, text, pressEnter)`
- **src/lib/agents.ts** + **src/lib/preferences.ts** - plumb the new key through the Claude def (413-417), the `AgentDef`/`AgentConfig` types, and the field-by-field copy in `mergeAgentConfig` (292-296), which silently drops any key not listed there
- No backend changes: `payload.reply` already flows through the ccmux-notifier argv and D-Bus (`NotificationReplied` maps to `answer`)

Open questions:

- A half-typed draft in the composer merges with the reply text and Enter submits the combined line. Undetectable from ccmux; accept and document (mirrors Approve's risk posture).
- A reply starting with `/` triggers Claude's slash-command palette; `sanitizeReply` does not guard this. Decide: strip, prefix, or document.

#### 2. Reply on permission notifications = deny with feedback (commit 2)

- **src/daemon/notifier.ts** - the permission branch (456-465) also attaches Reply when the def declares a new `notificationActions.permissionReplyPrelude` (Claude: `["Escape"]`). Reply sits alongside Approve/Deny; the combined macOS category is already registered.
- **src/daemon/notification-action.ts** - the `answer` gate accepts `waiting` + `permission`; that branch sends `permissionReplyPrelude`, settles, then text + Enter. Semantics: this IS a deny (the tool is cancelled), with the text arriving as the next user message that Claude treats as steering feedback.
- **src/lib/agents.ts** - load-bearing comment mirroring 405-412: Escape at the permission prompt is UNVERIFIED across prompt variants (Bash approval, Edit/Write diff approval, permission-rule variant, MCP tools); the e2e pass must confirm Escape cancels to the composer before merge.
- Decided: the button label ships as "Reply" with the deny semantics documented in the README (a "Deny + reply" label would require a new category shape and a helper release; the body already shows the pending command, so the context is clear).

#### 3. Actions + context body for plan_approval (commit 3)

- **src/lib/agents.ts** - new `planApprove`/`planDeny`/`planReplyPrelude` keys. Claude defaults need e2e discovery first: the ExitPlanMode picker options vary (approve vs approve-with-auto-accept vs "keep planning"), so do not guess whether digit quick-select submits. Decided: `planApprove` targets the PLAIN approve option, never auto-accept (a notification press must not silently grant broad write permission); e2e discovers the actual keystroke. `planDeny: ["Escape"]` as the starting point to verify.
- **src/daemon/notifier.ts** - a `plan_approval` branch in `buildPayload` next to 456-465, attaching approve/deny/reply per def presence
- **src/daemon/notification-action.ts** - approve/deny/answer gates accept `waiting` + `plan_approval` (token-matched) and select the plan key lists
- **src/daemon/notify-context.ts** - a `plan_approval` branch in `buildNotificationContext` (365-371): pane-first `extractPlanPrompt`, anchored on the last "would you like to proceed" terminator (`TERMINATOR_RE` already matches it), clamped with the waiting caps (4 lines / 300 chars). Plan boxes run long, so verify whether `PANE_CAPTURE_LINES = 30` is deep enough. Stretch: fall back to the `ExitPlanMode` `tool_use.input.plan` off the transcript tail.

Open question: a plan wait can be stored as `permission` during the marker-fresher window. Consider widening `NotificationContext.reclassifyAs` to include `"plan_approval"` when the capture contains `/.claude/plans/`, so delivery renders plan actions and the "Plan ready for review" subtitle. Verify first whether the `Notification` hook actually fires for ExitPlanMode; if not, this is moot.

#### Testing

- **src/daemon/notifier.test.ts** (stamping describe, ~981) - finished reply opt-in/opt-out, permission actions + reply together, plan_approval actions, reclassify-to-plan if implemented
- **src/daemon/notification-action.test.ts** - answer-on-idle happy path (no prelude keys sent), answer-on-idle stale token 409 + re-notify, answer-on-permission prelude, answer-on-permission without the def key 409, plan approve/deny key selection; update (not delete) the reject test at ~363
- **src/daemon/notify-context.test.ts** - `extractPlanPrompt` fixtures from a real captured plan pane
- **src/lib/agents.test.ts** - new keys survive `mergeAgentConfig` and user overrides
- **src/lib/notify.test.ts** / **notify-delivery.test.ts** - actions + reply combined on one payload across backends

#### Docs

- **docs/architecture.md** notifications section (scope sentence, action gates, context bodies, reclassify if widened)
- **README.md** notifications section, including a NOTE that Reply on a permission notification denies the pending tool
- **docs/agent-adapters.md** Claude caveats: record the verified plan-picker and permission-Escape findings with the Claude version

#### Verification requirement (merge gate)

Mandatory real-agent e2e pass, per the precedent in `agents.ts:405-412`: drive a real Claude in a detached tmux session, POST the callbacks to `/notification-action` with the live `statusChangedAt`, and assert the pane advanced. Must cover: reply at the idle composer (confirm plain text + Enter submits and that the Escape prelude is indeed harmful there), Escape-then-text at each permission prompt variant, the plan picker approve digit and Escape behavior, and one stale-press 409 path.

### Relevant agent(s)

Claude Code

### Additional context

Depends on PR #24 (builds directly on its title/subtitle/body layout). Commit 1 lands first so it can be peeled off if e2e verification of 2 or 3 surfaces a prompt-variant problem.

Follow-up: #26 extends actions to non-Claude agents and depends on the capability shape introduced here.

epilande:
## Overview

Expands Claude notification actions with Reply on finished notifications, Reply-as-deny-with-feedback on permission prompts, and Approve/Deny/Reply on plan approvals, hardened by a pane-authoritative press gate that keys the picker actually on screen.

Resolves #25

## Motivation

- Finished is the highest-traffic notification and carried no actions, even though the pane sits at an idle composer where a typed follow-up is the cheapest, most useful action of all.
- Denying a permission couldn't carry steering feedback, even though Claude's own prompt has a "tell Claude what to do differently" path.
- Plan approvals had no buttons and no context body, and the verification pass proved the stored plan/permission classification is unreliable at exactly the moment a notification fires, so the press path needed to stop trusting it.

All action shapes reuse the existing `approve`/`deny`/`answer` ids, so the macOS helper's registered category set is untouched: zero Swift changes, no helper release.

### Action surface

- **src/daemon/notifier.ts** - Stamps Reply on finished (`replyOnFinished`), question (`replyOnQuestion`, replacing the hardcoded Claude check), permission (`permissionReplyPrelude`), and plan waits (`planReplyPrelude`); plan waits get Approve/Deny from the separate `planApprove`/`planDeny` keys. Buttons are both-or-none so payload shapes always match the helper's registered categories, and paneless sessions get no actions at all.

### Press gate

- **src/daemon/notification-action.ts** - The flat state gate becomes a staleness-token check plus a pure `resolveActionPlan` that decides how each press executes in the live state. For agents whose plan picker differs from their permission prompt, the captured pane is authoritative for the plan/permission split: both the marker (`pending_tool: null`) and the log (`ExitPlanMode` tool_use nondeterministically deferred) can be silent, so keying from stored state sent `1` (auto mode) at plan pickers and could send `2` ("don't ask again") at permission prompts. Approve/Deny fail closed when no prompt is classifiable. A liveness guard refuses to type when the pane's foreground process is a shell (a reply after the agent exits would execute as a command), and replies starting with `/` get a leading space so they can't trigger the slash-command palette.

### Prompt classification

- **src/daemon/pane-classify.ts** - New `classifyClaudePromptPane`, bottom-anchored on the last prompt terminator so stale plan chrome above a fresh prompt can't misclassify it; now the single owner of the prompt terminator regex.

### Notification context

- **src/daemon/notify-context.ts** - Plan waits get a context body: transcript `input.plan` first, pane extraction (anchored on the "Here is Claude's plan:" header, deeper 60-line capture) when the transcript is silent. `reclassifyAs` widens to `plan_approval` so a plan wait stored as a permission renders the right subtitle and buttons. A currency guard rejects superseded plans, and the duplicated transcript-tail pipeline is shared.

### Config plumbing

- **src/lib/agents.ts** - New `notificationActions` keys (`replyOnQuestion`, `replyOnFinished`, `permissionReplyPrelude`, `planApprove`, `planDeny`, `planReplyPrelude`) with keystrokes verified against a real agent (Claude Code 2.1.211), including the auto-mode footgun note: plan approve is `2`, never `1`. The merge copy becomes a whole-object spread.
- **src/lib/preferences.ts** - `AgentConfig` mirror for the new keys.

### Docs

- **README.md** - Reply surfaces, the deny semantics of permission Reply, and a migration NOTE that a `notificationActions` override is a whole-object replace.
- **docs/architecture.md** - Gate matrix, press-time guards, and context-body updates.
- **docs/agent-adapters.md** - Verified plan-picker findings (option order, digit-submits, Escape behavior, marker/log silence) on Claude Code 2.1.211.

## Breaking Changes

None. A user override of `agents.claude.notificationActions` keeps its documented whole-object-replace semantics, and now also controls the reply surfaces; omitted keys disable them (see the README NOTE).

## Testing

- [x] Full suite: 2794 pass / 0 fail, typecheck clean
- [x] Real-agent e2e (Claude Code 2.1.211, isolated branch daemon): reply at the idle composer, slash guard, deny-with-feedback across Bash, Edit/Write, and MCP prompt variants, plain approve, stale-press 409
- [x] Plan-approve safety case caught live and re-verified fixed: with the session stored as `{permission, pendingTool: null}`, POST approve now sends `2` (footer flips to "manual mode on", auto mode NOT enabled); the inverse press right after a plan approval sends `1` with no persistent rule created
- [x] Liveness guard e2e: reply against a killed agent's pane returns 409 with nothing typed into the shell
- [x] Reviewed by a multi-agent verification pass plus an independent Codex cross-check; all confirmed findings fixed in the two hardening commits


epilande:
E2E merge-gate evidence (real Claude Code 2.1.211, isolated branch daemon via `CCMUX_PORT=2270`, verified against HEAD a724ada):

- **Plan approve, the safety case**: with the session stored as `{waiting, permission, pendingTool: null}` at POST time (the common live-plan state), approve sends key `2` and the footer flips to "manual mode on". Pre-fix, the identical state sent key `1` and enabled auto mode.
- **Inverse direction**: a Bash approval right after a plan approval sends key `1` (approve-once); a second identical command re-prompted, proving no "don't ask again" rule was created.
- **Liveness guard**: after killing the agent process (pane foreground becomes `zsh`, session retained as idle), a Reply returns 409 with nothing typed into the shell.
- **No false positives**: idle reply on a healthy pane returns 200 and submits; slash-prefixed replies submit as text without opening the command palette.
- **Offer side**: a live plan-picker capture classifies as `plan_approval` and yields the plan body from the pane, so the notification reads "Plan ready for review" with Approve/Deny/Reply even while `pendingTool` is null.
- First-pass coverage also verified: deny-with-feedback across Bash, Edit/Write, and MCP prompt variants, and the stale-token 409 path.

PR #29, opened 2026-07-18
feat(notifications): Add attention-generation staleness token

### LINKED ISSUE #28: [Enhancement]: Attention-generation token for stale notification presses across waiting-to-waiting swaps
### Problem / motivation

The notification staleness token is `statusChangedAt`, which bumps only on status edges (`SessionManager.updateSession`). When one wait resolves and a NEW wait begins within a single scan (the daemon observes `waiting -> waiting`), the token still matches, so a press on the stale notification executes against a prompt the user never saw.

PR #27 already neutralizes the dangerous cross-type arms for Claude: the pane-authoritative press gate classifies the live picker and keys it correctly, and Approve/Deny fail closed when no prompt is classifiable. What remains is the same-type swap:

- Notification fires for Bash prompt A. A resolves in the pane; Bash prompt B appears within one scan tick. A press on A's notification passes the token check AND the pane classification (both prompts classify as `permission`), so it approves B sight-unseen.
- The same shape exists for two consecutive plan waits.

This class pre-exists PR #27 for permission waits (the v2 risk posture in the README covers it: buttons only send the mapped keystroke). It is filed as a follow-up because closing it fully needs a token change, not another guard.

### Proposed solution

Fold an attention generation into the echoed staleness token so a press must match the exact wait, not just the exact status edge:

- Bump a per-session `attentionChangedAt` (or a monotonic generation counter) on ANY `attentionType`/`pendingTool` change inside `SessionManager.updateSession`, including `waiting -> waiting` swaps observed in one reconcile.
- Stamp it into the notification payload next to `statusChangedAt` (the macOS helper echoes the `--payload` blob opaquely, so no Swift change or helper release is needed; the D-Bus path constructs the callback in-process).
- Require an exact match in `handleNotificationAction` for `approve`/`deny`/`answer`.

Caveats to design around:

- The generation must NOT bump on a delivery-time-only reclassification (permission -> question/plan rendered from the pane), or every reclassified notification would 409 its own buttons.
- A `waiting -> waiting` swap that the daemon never observes (both edges inside one scan, no marker events) still cannot bump any counter. The pane guard from PR #27 remains the last line for those; document the residual window.

### Relevant agent(s)

All agents with `notificationActions` (Claude today, #26 extends the surface).

### Additional context

Surfaced by the multi-agent review + Codex cross-check on PR #27 (finding: "the staleness token can't see intra-waiting attention swaps across a missed edge"). Depends on the action plumbing from #25/PR #27.

epilande:
## Summary

The notification staleness token was `statusChangedAt`, which bumps only on status edges. When one wait resolves and a new wait begins while status never leaves `waiting` (a waiting -> waiting swap), the token still matched, so a press on the stale notification could act on a prompt the user never saw. The PR #27 pane-authoritative gate only neutralizes cross-type swaps (both prompts classifying as `permission` pass it).

This adds `Session.attentionGeneration`, a monotonic per-session counter bumped solely in `SessionManager.updateSession` whenever the attention identity (`attentionType` or `pendingTool`) changes. It is stamped into the notification payload next to `statusChangedAt`, threaded through both backends (the macOS helper echoes the `--payload` blob opaquely, so no Swift change or helper release is needed; the D-Bus path passes it in-process), and `handleNotificationAction` requires an exact match for `approve`/`deny`/`answer`. A mismatch, or a missing field from a notification fired by an older daemon build, fails closed with 409 plus a fresh "state changed" re-notification. `default` (focus) presses are not gated.

Design notes:

- A counter instead of the issue's suggested timestamp: marker events apply with `skipDebounce`, so two updates can land in the same millisecond and an ISO timestamp could collide.
- The delivery-time reclassification caveat from the issue is satisfied structurally: `buildPayload`'s `reclassifyAs` never calls `updateSession`, so it cannot bump the generation.
- Pane reuse (`resetPaneTrackedSessionState`) deliberately does not reset the counter, keeping it monotonic so a stale cross-run press always mismatches.
- Residual window (documented in `docs/architecture.md`): the generation only moves on an observed field change. A swap whose edges both land inside one reconcile scan with no marker events, or a field-identical swap (same `attentionType` AND `pendingTool`, e.g. two consecutive Bash permission prompts, since `pendingTool` is the tool name) with a missed intermediate `working` edge, bumps nothing. The pane-authority gate remains the last line of defense there; full closure would need a per-wait identity in the marker payload itself.

## Related issue

Closes #28

## Testing

- [x] `bun test` passes (2825 pass / 0 fail, 15 new tests)
- [ ] Verified the affected TUI surface (picker / sidebar) in a detached tmux session (if this touches the TUI, columns, layout, theming, or the daemon → TUI data path)

New tests cover: generation init and bump semantics in `updateSession` (bumps on attentionType or pendingTool change, single bump when both change, no bump on status-only change, bumps on a waiting -> waiting pendingTool swap), the approve/deny/answer exact-match gate (match succeeds, mismatch 409s and re-notifies, missing input fails closed, `default` unaffected), payload stamping, `--payload` serialization, D-Bus threading, and server callback parsing (non-number treated as absent).

No TUI surface is touched (a picker render-check under the branch build passed anyway).

E2E verified against a real Claude session through an isolated daemon running the branch code, with the session at a live Bash permission wait (generation 1): a stale-generation approve press (`attentionGeneration: 0`, current `statusChangedAt`) returned 409 with the prompt visibly untouched, a missing-generation press failed closed with 409, and a current-generation press returned 200 with the approve keystroke landing and Claude proceeding to working. The subsequent wait advanced the daemon's generation 1 -> 2, the exact waiting -> waiting movement the token exists to track. A screen recording of the run is attached in a comment below.

## Notes for reviewers

- The only bump site is `updateSession`; nothing else touches the counter.
- Follow-up candidate spotted while implementing (not addressed here): the `Notifier` dedup key is `statusChangedAt` (ISO ms), so a same-millisecond waiting -> waiting swap could suppress the banner for the new wait entirely. The press-side gate still protects against a stale press; only the fresh banner is at risk.


epilande:
ccmux-verify-pr e2e recording (claude, notification press path via isolated branch daemon on :2271). Verifies: the attentionGeneration staleness gate from #28: a stale-generation approve press returns 409 and leaves the prompt untouched, a missing-generation press fails closed with 409, and a current-generation press approves the live prompt. Exercises notification-action.ts, server.ts callback parsing, and the notifier payload stamping end to end with a real Claude session.

https://github.com/user-attachments/assets/5c91cc81-45e8-40e7-8437-c127b611c19c

Issue #26, opened 2026-07-16
[Feature]: Notification actions for non-Claude agents


epilande:
### Problem / motivation

Notification actions are Claude-only today. Claude is the only built-in agent with `notificationActions` (**src/lib/agents.ts:413-417**), and the Reply gate hard-codes `agentType === "claude"` (**src/daemon/notifier.ts:466-474**). Codex, OpenCode, Cursor, Gemini, and Antigravity permission waits already fire notifications, they just carry no buttons.

There is also a real pre-existing gap: brand-new custom agents cannot receive `notificationActions` at all. The custom-agent construction branch in `getAgents` (**src/lib/agents.ts:769-808**) never maps `notificationActions` (nor `ambiguousPermissionMarker`) from the config, despite `src/lib/preferences.ts:60-71` documenting that custom agents can opt in. Only overrides of built-in agent names work today, via `mergeAgentConfig`.

### Proposed solution

Per-agent commits adding verified `notificationActions` defaults. Every keystroke is e2e-verified against the installed agent version before it is encoded, with the verified binary version stamped into the def comment (precedent: **src/lib/agents.ts:405-412**, which made e2e confirmation the merge gate for Claude's keys).

Backend and helper impact is zero, verified: the macOS helper resolves categories from a fixed shape set keyed only on the reused `approve`/`deny`/`answer` ids (**notifier/Sources/main.swift:113-165**), and D-Bus builds actions dynamically from the payload. `handleNotificationAction` is already fully def-driven, so the per-agent work is choosing verified defaults, not new architecture.

#### Commit 0: generalize the gates + custom-agent fix

- **src/daemon/notifier.ts** - replace the `agentType === "claude"` Reply gate (466-474) with def-presence (attach Reply for a question wait when the agent's `notificationActions` declares question-reply support). This is the only agent-name hard-coding in the action pipeline; the Claude-only gates in `notify-context.ts` cover context BODIES, not actions, and stay as a non-goal. Pick one capability convention (e.g. `answerPrelude` present, with `[]` meaning "reply supported, no prelude") and document it in the `notificationActions` doc block.
- **src/lib/agents.ts** - map `notificationActions` (and `ambiguousPermissionMarker`) in the custom-agent construction branch (769-808), reusing the same normalization as `mergeAgentConfig`.

#### Commit 1: OpenCode

- Detectable waits: `permission` via terminalRules and authoritatively via the plugin's `permission.asked` marker. No question detection.
[...]

PR #31, opened 2026-07-19
feat(agents): Add GitHub Copilot CLI as a built-in agent

epilande:
## Summary

Adds GitHub Copilot CLI (`@github/copilot`, the standalone agentic CLI, not the legacy `gh copilot` extension) as a built-in agent with full native tracking: a hook adapter driving the marker pipeline, a real-time log adapter over Copilot's `events.jsonl`, and terminal rules as the cascade fallback. Copilot sessions get authoritative session identity, instant state transitions, `--resume` support, prompt history, and `ccmux invoke copilot`.

Copilot turned out to be one of the best-instrumented agents ccmux supports:

- `~/.copilot/session-state/<uuid>/events.jsonl` flushes incrementally in real time, and `permission.requested` lands on disk while the approval dialog is still up (verified live on v1.0.71), so the log source alone can catch a permission wait. Claude's transcript defers permission-gated `tool_use` until after resolution; Copilot has no such gap.
- Copilot ships a Claude-grade hooks system with an observational `notification` event (`notification_type: permission_prompt` carries the pending command text), so markers flip to `waiting_permission` with context the moment the dialog opens.
- The hooks install is a drop-in: two ccmux-owned files in the auto-discovered `~/.copilot/hooks/` dir. Install/uninstall never touches user-owned config.

- **feat(agents)** - `copilot` agent definition: basename-anchored process match, terminal rules for the permission and folder-trust dialogs plus the `esc interrupt` working footer, `copilot --resume {id}`, invoke mode (`-p {prompt} --allow-all-tools`), and native-id discovery via the `session-state/<uuid>/session.db` fd Copilot keeps open (its `events.jsonl` is append-and-close)
- **feat(hooks)** - `CopilotHookAdapter` + marker script registering observational events only (`sessionStart`, `userPromptSubmitted`, `notification`, `agentStop`, `sessionEnd`). The deciding `permissionRequest` hook is deliberately never registered: its output can allow/deny the pending tool call. The script derives pid/tty from `$PPID` (Copilot is the hook's direct parent) and treats `session-start` on an existing marker as identity-refresh only, since `userPromptSubmitted` fires before the deferred `sessionStart`
- **feat(daemon)** - `CopilotLogAdapter` mapping `user.message`/`assistant.turn_start` to working, `permission.requested` to waiting (permission), and `assistant.turn_end`/`session.shutdown`/`abort` to idle, wired alongside the Claude/Codex watchers with a per-scan marker link step
- **fix(agents)** - Match only the real SEA binary, never its node wrapper (details in the notes below)
- **docs** - Copilot section in `docs/agent-adapters.md` (adapter caveats, file paths), README setup/matching sections, AGENTS.md agent lists

## Testing

- [x] `bun test` passes (2830 tests, includes new hook-adapter, hook-scripts, log-adapter, terminal-rule, and agent-def coverage)

E2e against real Copilot CLI v1.0.71 (npm install, live GitHub auth): a session in a detached tmux pane cycles idle → working → waiting (`attentionType: permission`, `pendingTool: Command`) → idle in the picker, with the marker file observed carrying the correct pid, tty, session UUID, transcript path, and permission context at each step, and removed on `/exit`. A fresh session shows `nativeSessionId`, `logPath`, and `lastPrompt` all populated and stable across scans. Picker rendering verified at wide (agent column `Copilot`) and narrow (`cp`, lowercase like every other short code) breakpoints. Screencast of the full state cycle attached below.

## Notes for reviewers

- The daemon-level e2e caught a bug unit tests could not: the npm/mise wrapper (`node .../bin/copilot`) spawns the real SEA binary as a child on the same pane tty. Matching the wrapper via `commandPatterns` double-detected the pane, flip-flopping the session's pid every scan and tripping the pane-reuse identity reset, which cleared `nativeSessionId`/`logPath`/`lastPrompt` as fast as enrichment wrote them while status (re-derived by the cascade) masked the churn. The fix drops `commandPatterns` entirely; the real binary always exists as a child of any wrapper form, so basename matching stays complete and becomes unique.
- Copilot fires `sessionStart` at the first prompt submission, not at UI launch, and can emit `userPromptSubmitted` before it. The pre-prompt phase (including the folder-trust dialog) is covered by terminal rules only; the marker script never lets a late `session-start` downgrade a fresher state.
- Deliberate non-goals, queued as follow-ups: `notificationActions` for Copilot's numbered permission dialog (refs #26), a brand color in `agentColorFor` (falls back to overlay, same as cursor/antigravity), and `COPILOT_HOME` relocation support beyond the marker script's transcript path.


epilande:
E2e screencast (screen recording, sped up): the session picker filtered to a live Copilot CLI session cycling idle → working → waiting (permission, pending tool "Command") → idle after the command is approved.

https://github.com/user-attachments/assets/8b23a5de-0af1-4f58-bcc1-0e58f457086a

PR #49, opened 2026-07-24
fix(daemon): Wire picker x on background rows to claude stop

### LINKED ISSUE #48: Picker `x` silently no-ops on background rows; wire it to `claude stop`
## Problem

Pressing `x` (kill session) on a background-agent row (`trackingMode: "background"`, the paneless `claude --bg` rows) does nothing visible. Two layers are broken:

1. **The daemon refuses the kill by design but the refusal is invisible.** `handleKillSession` returns a 400 for background rows because their worker pid is owned by Claude's supervisor and ccmux must not SIGTERM it (`src/daemon/server.ts:1089-1096`). That policy is correct.
2. **The TUI swallows the error.** `killOrCancelSession` in `src/tui/App.tsx` fires the POST and never reads the response, so the 400 (and any other kill failure) is silently dropped. The user presses `x`, confirms the dialog, and the row just stays.

Observed instance: two background workers finished their useful work but ended their turn with a question, sat in `waiting` for a week, and could not be removed from the picker. `/clear` plus detaching from the attach view doesn't help either: row existence is driven purely by roster membership in `~/.claude/daemon/roster.json`, which ccmux mirrors read-only (`src/daemon/sources/claude-background.ts`).

## Proposal

Map `x` on a background row to Claude's own sanctioned stop path, keeping ccmux's read-only ownership model intact:

1. **Add `backgroundStopCommand` to `AgentDef`** (`src/lib/agents.ts`), following the existing `resumeCommand` precedent. Optional, something like `(id: string) => string[]`. Defined only for Claude today: `["claude", "stop", id]`. No harness literals in the server.
2. **Daemon**: `handleKillSession`'s background branch resolves the agent via the existing `getAgentByType` accessor and spawns the command (PATH-resolved), awaits exit, and maps a nonzero exit to a 500 carrying stderr. If the agent defines no `backgroundStopCommand`, keep a clear 400 ("background session is read-only; this agent has no stop command"). Never SIGTERM the worker pid directly.
3. **TUI**: `killOrCancelSession` checks the response and surfaces non-OK errors through the existing error display. This fixes error surfacing for all kill failures, not just background rows.
4. **Row removal stays event-driven**: `claude stop` makes the worker exit, the supervisor drops it from `roster.json`, and the Background Source's roster watcher removes the row. The kill handler does not remove the session itself.

Deliberately out of scope:

- `kill-all` / `kill-group` keep excluding background rows (`src/daemon/server.ts:1251`). Mass-stopping background agents from a group header is a bigger, more surprising action; per-row `x` covers the need.
- `claude rm` must NOT be used: it deletes the session **and its worktree** (unpushed work included). `claude stop` keeps the conversation resumable via `claude attach`, which matches what `x` means for every other row.

## Verification

- Server tests: background branch success + failure (mocked spawn), no-stop-command agent, and the untouched non-background path.
- TUI test: kill error surfaces in the error display.
- Live e2e (required): spawn a real `claude --bg` worker, confirm the row appears, press `x` in the picker, verify `claude stop` runs, the short drops out of `roster.json`, and the row disappears within a few seconds. Also verify the conversation remains resumable via `claude attach <id>`. Per AGENTS.md, capture picker output in a detached tmux session for any TUI-visible change.

## Repro (current behavior)

1. `claude --bg -p "do something, then ask me a question"` in some repo
2. Open `ccmux picker`, select the background row (`◇`), press `x`, confirm
3. Nothing happens; no error is shown; the row remains until the worker is stopped out-of-band

epilande:
## Summary

Pressing `x` on a background-agent row (the paneless `claude --bg` rows, `trackingMode: "background"`) silently did nothing. Two layers were broken: the daemon refused the kill with a 400 by design (the worker pid is owned by Claude's supervisor, so ccmux must never SIGTERM it), and the TUI never read the kill response, so the refusal was invisible and the row could not be removed at all.

This maps `x` on a background row to Claude's own sanctioned stop path while keeping ccmux's read-only ownership model intact:

- New optional `AgentDef.backgroundStopCommand` (following the `resumeCommand` precedent), defined only for Claude as `claude stop <short>`. No harness literals in `server.ts`; background rows' session id is the daemon short, so it's passed straight through.
- `handleKillSession`'s background branch resolves the agent via `getAgentByType` and spawns the command PATH-resolved, mapping a nonzero exit to a 500 carrying stderr. Agents without a stop command keep a clear 400. The handler never removes the row itself: removal stays event-driven (the supervisor drops the short from `roster.json` and the Background Source's roster watcher removes the row).
- `killOrCancelSession` in the TUI now reads the response and surfaces any non-OK result (and network failures) through the existing toast, fixing silent error-dropping for all kill paths, not just background rows.

Because removal is event-driven, the row outlives the request by a beat. Two follow-ons keep that window from reading as a failure:

- New optional `AgentDef.backgroundStopAlreadyGone` (for Claude, `/\bNo job matching\b/i`). A second `x` inside the removal window re-runs `claude stop` against a short the supervisor has already dropped, which exits 1; matching that stderr reports success instead of a 500, so a stop that landed is never reported as a failure. (A stop against a still-known but already-stopped job exits 0 on its own, so only the aged-out case needed handling.)
- A successful background stop toasts `Stopping agent...`. Every other row dies with its pane, which is feedback enough; a background row does not, and without confirmation `x` reads as having done nothing and invites the retry above.

The row context menu also gains `Kill` for background rows, so the mouse path matches the keyboard `x` that this PR wires up. `Restart` stays excluded, since that genuinely is a pane-session concept.

## Related issue

Closes #48

## Testing

- [x] `bun test` passes (2991 tests; 5 new server tests covering background stop success/nonzero-exit/already-gone/no-stop-command plus a SIGTERM regression for normal sessions, and 3 new TUI tests covering the failure toast, the background success toast, and the absence of that toast on a normal kill)
- [x] Verified end to end with a real agent session (if this touches detection, state, the daemon, or the CLI). Every change was exercised live against a real `claude --bg` worker with the daemon restarted from this branch; see the breakdown under "Notes for reviewers".

## Notes for reviewers

Live e2e with the daemon running this branch's build: spawned a real `claude --bg` worker, its `◈ waiting` row appeared in the picker, pressed `x` and confirmed the dialog; the spawned `claude stop` made the short drop out of `roster.json` and the row disappeared within a few seconds with no error toast. Afterward `claude attach <short>` resumed the conversation with full context, confirming stop (unlike rm) keeps the session resumable.

The follow-ons added since were re-verified against a daemon restarted from this branch (both the daemon and the picker running this build):

- **Happy path**: spawned a real `claude --bg` worker, pressed `x` on its row and confirmed. The `Stopping agent...` toast rendered, `roster.workers` went empty, the row disappeared, and no error toast appeared.
- **Context menu `Kill`**: verified live on a real background row. The capture below shows the background-only branch (`Attach agent` / `Open agent view`) now carrying `Kill`, with `Restart` still absent.
- **`backgroundStopAlreadyGone`**: verified live, with a control. With a stubbed `claude` whose `stop` exits 1 printing `No job matching '<id>'...`, `POST /sessions/<short>/kill` returned `200 {"success":true}` while the worker was demonstrably still in the roster (so the 200 came from the pattern branch, not an actual stop). The control, an identical exit 1 with non-matching stderr, still returned `500 {"error":"Failed to stop background session: supervisor unreachable"}`. So the branch discriminates on the pattern rather than swallowing nonzero exits.

The pattern itself is taken from the real CLI: `claude stop <unknown-short>` exits 1 with `No job matching '<short>'. Run 'claude agents' to list running sessions.` on stderr, while `claude stop <known-but-stopped-short>` exits 0 with empty stderr, which is why only the aged-out case needed handling.

Context menu on a real background row, showing the added `Kill` and the deliberately absent `Restart`:

```
 3 ◈ waiting scratc┌────────────────────┐ay hello in one short sentence, then ask me…
 ──────────────────│ Attach agent enter │────────────────────────────────────────────
 ▼ Cameo (1)       │ Open agent view    │
 4 ● idle    xcode/│ Kill             x │…
 ──────────────────│ Review diff      d │────────────────────────────────────────────
 ▼ ccmux (2)       └────────────────────┘
```

Picker during the run:

```
 ▼ bg-e2e-repo (1)
 3 ◈ waiting scratchpad/bg-e2e-repo:main  Say hello in one short sentence, then ask me a clarifying question…  Question   Claude   v2.1.218  25s

                    ┌────────────────────────────────────────────────┐
                    │                 Kill Session?                  │
                    │                                                │
                    │                  bg-e2e-repo                   │
                    │                                                │
                    │               Y confirm N cancel               │
                    └────────────────────────────────────────────────┘
```

Scope notes: `kill-all` deliberately keeps excluding background rows, so the bulk sweep leaves long-running background work running. `kill-group` is a client-side fan-out over the group's rows, so it stops them exactly like single-row `x` does; that asymmetry is intentional and documented in `docs/architecture.md`. `claude rm` is never used: it deletes the session's worktree, while `claude stop` keeps the conversation resumable via `claude attach`, which matches what `x` means for every other row.

One aside: the issue's repro command `claude --bg -p "<prompt>"` is rejected by current Claude (`--bg` conflicts with `--print`); the prompt must be positional (`claude --bg "<task>"`).


epilande:
E2E recording of this change (screen recording, sped up). The daemon in the clip runs this branch's build against an isolated session list, so only the test row is visible. A real `claude --bg` worker is dispatched off screen: its background row appears and settles in `waiting`, then `x` opens the kill dialog and `y` confirms. The daemon runs `claude stop <short>`, the supervisor drops the short from `roster.json`, and the row disappears through the roster watcher, with no direct SIGTERM and no immediate removal by the kill handler. Afterward the roster is empty and the conversation remains resumable via `claude attach`. The `App.tsx` error-toast path is unit-tested and not shown here, since the success path produces no toast.

https://github.com/user-attachments/assets/2bf70506-cce2-4aa0-9ec2-edb7699d61db

PR #51, opened 2026-07-25
feat(search): Rank picker search results by relevance

### LINKED ISSUE #50: [Feature]: Rank picker search results by relevance
## Problem

Search unions four sources (fuzzy metadata, prompt index, pane content, transcript) but never ranks the results. Matching rows keep the picker's normal sort: `waiting` first, then recency (`src/tui/store.ts:580-582`). fuzzysort computes a relevance score for metadata matches and it is discarded (`src/tui/store.ts:532-536`).

The result: searching for a session by its project name can surface an unrelated session at the top because that session is `waiting` and its transcript happened to contain the query once. The strongest match is often buried mid-list.

## Current behavior

- Membership is a boolean OR union across sources; no source tiering or field weighting (`src/tui/store.ts:573-578`)
- A project-name hit ranks identically to a deep transcript hit
- Group names are only searchable by accident: under project/cwd grouping the group key doubles as the project field, but under session/window grouping the tmux session/window name is not a search field at all (`src/tui/utils/grouping.ts:103-132`)
- Groups stay pin-order/alphabetical during search (`src/tui/utils/grouping.ts:197-222`)
- Transcript matches return in file order, first 5 found, with no preference for user prompts or recent turns (`src/daemon/transcript-search.ts:221-251`)
- `paneMatch` is computed but never rendered, so pane-matched rows give no clue why they matched (`src/tui/utils/grouping.ts:46`)

## Proposal

While a query is active, order results by a composite relevance score. Empty-query ordering stays exactly as it is today.

Session score:

- **Identity tier (highest)**: group key, `project`, `gitBranch`. Reuse the fuzzysort `.score` already computed; add the group key as a match field so session/window names (and renamed groups, #42) are searchable
- **Location tier**: `cwd` (weighted below identity; long paths scatter-match)
- **Content tier**: prompt index (newer prompts score higher), then pane, then transcript
- **Cross-source bonus**: matching multiple sources adds a small boost
- **Status and recency become tiebreakers** within equal scores, not the primary sort key

Grouping interaction (groups are never split):

- Group order = max of member scores, replacing the alphabetical sort during search; one strong match surfaces the whole group
- Members sort by their own score within the group, so the best match sits directly under the header; equal scores fall back to the existing status/recency order
- Pinned groups: relevance overrides pinning during search (same precedent as search force-expanding collapsed groups)

Secondary, same theme:

- Render a match-source indicator on rows (also fixes the dead `paneMatch` flag)
- Daemon snippet selection prefers user prompts and recent turns over first-found assistant text

## Constraints

- Async sources (pane, transcript) arrive ~250ms after keystroke and will re-rank the list. Selection must stay pinned by session ID and scores must only change on query/cache change (see the j/k sort-churn fix in #45).
- Matching semantics are out of scope. Substring matching for content sources is deliberate (fuzzy scatter-matches long haystacks, see PR #3 notes); this issue changes ordering only.

epilande:
## Summary

Search was a filter, not a ranker: matching rows kept the picker's waiting-first/recency sort, so searching for a project by name could bury the real match under an unrelated session that happened to contain the query in its transcript. While a query is active, rows now sort by a composite relevance score and the old status/recency order only breaks ties. Empty-query ordering is completely unchanged.

- **Identity tier** (base 3000 + 1000x fuzzysort score): `project`, `gitBranch`, and the group key, added as a new search field, so tmux session/window names are now searchable
- **Location tier** (2000 + 500x): `cwd`
- **Content tier**: prompt substring matches scored by recency (1000-1500), pane 600, transcript 400
- **Cross-source bonus**: +50 per extra matched source; the max total bonus (200) stays below the smallest tier gap so bonuses can never jump a tier (comment pins the invariant)
- Groups order by max member score during search and pinned groups yield to relevance (same precedent as search force-expanding collapsed groups); members sort by their own score, insertion order keeps them under the header
- Rows whose match is otherwise invisible render a dim `[pane]`/`[transcript]`/`[cwd]` tag, which also puts the previously dead `paneMatch` flag to work
- Daemon transcript snippets now rank user turns before assistant turns, newest first, instead of returning the first 5 in file order; the candidate pool is every match in the 2MB tail (a count cap would bias the pool toward the oldest matches)
- Fixes a latent bug where a bare JSON primitive line in a transcript threw into the outer catch and dropped the whole session's matches

Scores are a pure function of query + caches, and selection stays pinned by session id, so the ~250ms async pane/transcript re-ranks don't reintroduce the #45 sort-churn problem.

## Related issue

Closes #50

## Testing

- [x] `bun test` passes (3004 tests, 19 new: tier separation vs bonus, stable tiebreak, empty-query untouched, group-key searchability, selection pinned across late transcript re-rank, group max-score order, pin yield, >25-match snippet regression, indicator show/suppress)

Live checks against real sessions: searching `cameo` ranks the Cameo group (identity) above a 33s-old `working` session with a pane hit, which the old sort put on top; searching `flash` ranks FlashJump above an alphabetically-earlier transcript-only match, tag rendered:

```
 ▼ FlashJump (1)
 1 ● idle    FlashJump:feat/scroll-enter-to-labels  /clear
 ▼ .dotfiles (1)
 2 ● idle    ~/.dotfiles:master  [transcript] …flash.lua, harpoon.lua…
```

Snippet ranking probed directly against real transcripts: for `merge`, a July 9 user match ranks above a July 20 assistant match.

## Notes for reviewers

- Two independent review passes (Claude + Codex) before this went up. Codex caught the one real bug: the original 25-candidate cap filled oldest-first and starved recent user prompts; fixed with a regression test proven to fail against the old logic.
- Deliberate leftovers, agreed non-blocking: `matchSources` is consumed only by tests (documents intent for future UI), and the tag-reduced prompt budget is not applied to the highlight paths (unreachable while the tag-suppression rule holds).
- Pin-yield and waiting-demotion are unit-tested but weren't observable live (no pinned groups or waiting sessions existed during verification).
- Fuzzysort v3.1.0 scores are already normalized 0..1; the 0.3 threshold is unchanged.

### DIFF
--- a/src/daemon/transcript-search.test.ts
+++ b/src/daemon/transcript-search.test.ts
@@ -287,13 +287,14 @@ describe("transcript-search", () => {
         { id: "s1", agentType: "claude", logPath },
         "match",
       );
-      // The two well-formed user turns match; the malformed lines are skipped
-      // without dropping the whole session (which the outer catch would do).
+      // The two well-formed user turns match (ranked newest-first); the
+      // malformed lines are skipped without dropping the whole session
+      // (which the outer catch would do).
       expect(result).not.toBeNull();
       expect(result!.matches.length).toBe(2);
       expect(result!.matches.map((m) => m.snippet)).toEqual([
-        "first match here",
         "second match here",
+        "first match here",
       ]);
     });
 
@@ -333,13 +334,13 @@ describe("transcript-search", () => {
       );
       // The bare `null` line is a JSON primitive, not an object; it must be
       // skipped without dropping the whole session (which the outer catch
-      // in `searchTranscript` would otherwise do).
+      // in `searchTranscript` would otherwise do). Matches rank newest-first.
       expect(result).not.toBeNull();
       expect(result!.matches.length).toBe(3);
       expect(result!.matches.map((m) => m.snippet)).toEqual([
-        "first match here",
-        "second match here",
         "third match here",
+        "second match here",
+        "first match here",
       ]);
     });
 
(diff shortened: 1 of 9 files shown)


epilande:
ccmux-verify-pr surface demo (search ranking on ccmux picker, worktree build). Shows: composite relevance ranking and match-source tags from src/tui/store.ts, src/tui/utils/grouping.ts, src/tui/components/SessionItem.tsx, src/tui/components/SessionList.tsx. Surface demo driven by typing queries, not an agent lifecycle. The flash query frame shows all three tiers ranked: identity (FlashJump) > [pane] > [transcript].

https://github.com/user-attachments/assets/076c72e7-8a19-4ea3-b9e6-4f90e4fd39de

PR #56, opened 2026-07-25
fix(daemon): Classify shells by exact basename, not comm substring

### LINKED ISSUE #52: fix(daemon): Shell classification matches path substrings, hiding real Bash permission waits
## Problem

`findShellDescendants` classifies a process as a shell with a substring match:

```ts
// src/daemon/process-tree.ts:114
SHELL_NAMES.some((s) => proc.comm.includes(s));
```

On macOS, `ps -axo comm` returns the full argv[0] string (a path), not the executable basename. So the substring match fires on directory components: any path containing `.local/share/...` matches `"sh"` via "**sh**are", `crashpad_handler` matches via "cra**sh**pad", `ApplePushService` via "Pu**sh**", and so on.

The only production consumer is `detectToolExecution` (`src/daemon/state-reconciler.ts:440-480`). It selects Claude sessions that are `waiting` with `attentionType: "permission"` and `pendingTool: "Bash"`, and flips them to `working` (clearing the attention indicator) when any shell descendant exists. The intent is "the user already approved the Bash and it is running, stop showing attention."

With the false positives, that flip fires when the "shell" is actually a language server. A Claude session genuinely blocked on a Bash permission prompt gets force-flipped to `working` and the attention indicator is silently hidden for as long as the false-positive process lives. Language servers live as long as the editor, so this is effectively permanent for any session in a repo with an LSP running under a path containing "share".

## Evidence (measured live)

- 9 processes on a dev machine (~1200 processes) classified as shells purely by path substring, all non-shells. Examples: `~/.local/share/mise/installs/node/26.3.0/bin/node` (TypeScript language servers), `~/.local/share/cursor-agent/versions/.../node`.
- 3 of those sat directly under a live `claude` process in a tmux pane and 1 under a live `cursor-agent`, so the misclassification is active in real sessions today, not theoretical.
- Direction of the divergence: "substring says shell, exact-basename says not" = 9, all false positives. "exact-basename says shell, substring says not" = 0. Zero real shells are lost by tightening the rule.
- macOS `comm` and the first whitespace token of `command` are byte-identical across all ~1200 processes (verified), including login shells (both report the dashed `-zsh` form) and rewritten-argv0 processes (`sshd-session: user [priv]`).

## Fix

Two-part change to the classification rule in `src/daemon/process-tree.ts`:

1. Derive the comparison key as `basename(stripOneLeadingDash(comm))`: strip one leading `-` (login shells), take the text after the last `/`. Never match against the joined command string (that adds 29 more false positives on a typical machine, e.g. `vim .zshrc`, browser "Helper" processes, and `/usr/bin/login ... /bin/bash -c exec -l /bin/zsh`, which is the parent of a shell, not a shell).
2. Replace `includes` with exact set membership, and expand `SHELL_NAMES` to compensate for shells the substring rule caught accidentally: `bash`, `sh`, `zsh`, `fish`, `dash`, `ksh`, `csh`, `tcsh`, `ash`. Exact matching also kills the residual basename-level false positives (`sshd-session:`, `ssh-agent`, `sharingd`, `ReportCrash`, `shazamd`), of which `ssh-agent`/`sshd-session` can plausibly be descendants of an agent pane.

Linux note: `/proc/<pid>/comm` is already a 15-char-truncated basename with no path and no leading dash, so the basename derivation is a no-op there and exact matching remains correct. The current substring rule has no path false positives on Linux; this change keeps it that way.

## Behavior change

Sessions that were being silently flipped to `working` will now correctly hold at `waiting` with the attention indicator lit. This is more visible attention, not less, and it is the correct behavior. The inverse failure (a real shell missed, session stuck at `waiting` while the command runs) does not occur with the expanded name set (verified zero false negatives on a live machine).

## Test requirements

Existing coverage cannot catch this: `src/daemon/process-tree.test.ts:75-89` builds the tree from the live machine's `ps` and asserts the implementation's own predicate character-for-character, so it can never fail. Every reconciler-level test stubs `findShellDescendants` out entirely.

Add fixture-driven tests (canned ps output, no live spawn):

- Positive: `-zsh`, `-/bin/zsh`, `/bin/zsh`, `/bin/bash --noprofile --norc`, `/opt/homebrew/bin/fish`, `sh -c ...`
- Negative: `~/.local/share/mise/installs/node/26.3.0/bin/node` (the live bug), `/usr/bin/login -flp user /bin/bash -c exec -l /bin/zsh`, `sshd-session: user [priv]`, `/usr/libexec/sharingd`, `/System/Library/CoreServices/ReportCrash`
- An end-to-end reconciler case: Claude session `waiting`/`permission`/`Bash` with a language-server descendant stays `waiting`; with a real `zsh` descendant flips to `working`

## Context

Found during the 2026-07-25 performance audit while evaluating whether `ProcessTree.build()`'s separate `ps -axo pid,ppid,comm` spawn could be eliminated (it can, but saves only ~4.5ms since the two ps calls already run concurrently; that part is deliberately out of scope here). `SHELL_NAMES`/`findShellDescendants` date to the initial import with no recorded rationale, so no prior decision is being reversed.

epilande:
## Overview

Fixes shell-process classification in `findShellDescendants` to match the exact process basename instead of a substring against the raw `comm` path.

Resolves #52

## Motivation

`detectToolExecution` flips a Claude session stuck on a Bash permission prompt to `working` (clearing the attention indicator) once it sees a shell descendant of the pane's Claude process, on the assumption that the user already approved the command and it's running. The old check matched `SHELL_NAMES` with `String.includes` against `comm`, which on macOS is a full path, not a basename. Any process whose path contains a shell name as a substring, like a language server living under `~/.local/share/...` (matches "sh" via "share"), got classified as a shell. That silently hid real Bash permission waits for as long as the false-positive process lived, which for an LSP is effectively the lifetime of the editor.

### Classification

- **src/daemon/process-tree.ts** - Adds `shellCommKey()`, which derives an exact-match key from `comm` (minus one leading `-` for login shells, minus path prefix via `path.basename`), and switches `findShellDescendants` from substring matching to exact set membership against an expanded `SHELL_NAMES` (adds `dash`, `ksh`, `csh`, `tcsh`, `ash`, which the old substring rule caught by accident). Also extracts `ProcessTree.fromPsOutput()` out of `build()` so the parsing half is testable without spawning a real `ps`. This is a deliberate narrowing versus the old substring rule, which also happened to match `mksh`, `ksh93`, `yash`, `xonsh`, `elvish`, and `osh`; the new exact set doesn't cover those, but a miss just fails conservatively (the row stays `waiting` with the attention indicator lit, never silently swallowed), and `detectToolExecution` only fires for Claude's Bash tool in the first place.

### Tests

- **src/daemon/process-tree.test.ts** - Replaces the old live-ps test (which asserted the implementation's own predicate against itself and could never fail) with fixture-driven cases: positive fixtures (`-zsh`, `-/bin/zsh`, `/bin/zsh`, `/bin/bash`, `/opt/homebrew/bin/fish`, and a real space-containing macOS path) and negative fixtures (the `.local/share` language server, `sshd-session:`, `sharingd`, `ReportCrash`, and a space-containing app-bundle Helpers path). A separate defensive-tolerance block documents that command-style (argv-shaped) input is not real `comm` output and is not specially parsed, including one known false positive on a synthetic argv string.
- **src/daemon/state-reconciler.test.ts** - Adds an end-to-end reconciler case: a Claude session waiting on a Bash permission prompt stays `waiting` when its only descendant is a language server, and flips to `working` when the descendant is a real `zsh`.
- **src/daemon/process-tree-test-helpers.ts** - Shared `PS_HEADER`/`psLine()` fixture builders for canned ps output, used by both test files.

## Breaking Changes

Sessions that were previously (incorrectly) silently flipped to `working` by the false-positive match now correctly hold `waiting` with the attention indicator lit. This is more visible attention, not less, and is the intended fix, but it is a user-visible behavior change for anyone whose false-positive process (e.g. an LSP under a path containing "sh") was masking a real Bash wait.

## Testing

- [x] `bun test` - 3029 pass, 0 fail
- [x] Fixture-driven unit tests for `shellCommKey()` and `findShellDescendants()` covering all positive/negative cases from the issue
- [x] End-to-end reconciler test covering the language-server-stays-waiting and real-shell-flips-to-working cases
- [ ] Live picker e2e against the shared daemon (handled separately per instructions; daemon not restarted from this branch)

### DIFF
--- a/src/daemon/process-tree-test-helpers.ts
+++ b/src/daemon/process-tree-test-helpers.ts
@@ -1,0 +1,14 @@
+/**
+ * Shared fixtures for building canned `ps -axo pid,ppid,comm` output to feed
+ * `ProcessTree.fromPsOutput()` in tests, without spawning a real `ps`.
+ *
+ * Used by process-tree.test.ts and state-reconciler.test.ts.
+ */
+
+/** Header row `fromPsOutput` discards (mirrors real `ps -axo pid,ppid,comm`). */
+export const PS_HEADER = "  PID  PPID COMM";
+
+/** One `ps` output line for the given pid/ppid/comm. */
+export function psLine(pid: number, ppid: number, comm: string): string {
+  return `${pid} ${ppid} ${comm}`;
+}
--- a/src/daemon/process-tree.test.ts
+++ b/src/daemon/process-tree.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from "bun:test";
-import { ProcessTree } from "./process-tree";
+import { ProcessTree, shellCommKey } from "./process-tree";
 
 describe("ProcessTree", () => {
   test("build() creates a tree from ps output", async () => {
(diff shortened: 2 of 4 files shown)


epilande:
### End-to-end verification

Screen recording of `ccmux picker` (sped up), filtered to a throwaway Claude session. The session is deliberately set up in the environment this PR is about: a long-lived `~/.local/share/mise/installs/node/26.3.0/bin/node` process running as a direct child of the Claude process, i.e. a descendant the old substring rule read as a shell because `.local/share` contains "sh".

What to watch:

1. The row holds `waiting` with the attention indicator lit and `Bash` as the pending tool while a Bash permission prompt is open. This is exactly the tuple `detectToolExecution` selects on, so on `main` the false-positive descendant would flip this row to `working` and hide the attention indicator.
2. After approval the row moves to `working` while the approved command's real `/bin/zsh` child runs, confirming the tightened rule still detects genuine shells.

Measured live at the instant the row was `waiting` / `permission` / `Bash`, against the real Claude PID: `findShellDescendants` returned 1 "shell" under the old substring rule and 0 under the new exact-basename rule, and the row held `waiting` across 60+ seconds of reconciler scans without flipping. Across the whole machine (~1200 processes) the old rule produced 30 false positives and the new rule zero, with no real shells lost (76 detected by both). 3 of 8 live Claude processes had at least one such false-positive descendant at the time of the run.

One thing worth recording for later, found while reaching the state above. The `pendingTool === "Bash"` half of `detectToolExecution`'s filter is only reachable through a carry-over: the ccmux Claude hook cannot populate `pending_tool` anymore (Claude Code >= 2.1.209 sends a generic permission message with no tool name, per the comment in `ccmux-state-notify.sh`), so `pendingTool` at prompt time comes solely from `marker.pending_tool ?? session.pendingTool` carrying a value the log adapter set for an earlier or in-flight Bash. That is a race, and it lost 3 of 5 attempts here, which narrows the real-world blast radius of the bug this PR fixes but does not change the fix itself. Not a blocker for this PR.

https://github.com/user-attachments/assets/e9076a49-20e8-471d-9dc6-bc79aa3e6311

PR #57, opened 2026-07-26
perf(daemon): Source process tty from lsof fds instead of the ps tty column

### LINKED ISSUE #54: perf(daemon): Source process tty from lsof fds instead of the ps tty column
## Problem

The daemon's per-tick process discovery (`src/daemon/processes.ts:148`) runs:

```
ps -eo pid,ppid,tty,etime,command
```

The `tty` column is expensive on macOS: ~90-100ms with it, ~20-29ms without (measured repeatedly; also reproducible with `/usr/bin/time`). The cost is a fixed device-name table build inside BSD `ps`, not per-process work; restricting to specific pids does not help (2+ pids still builds the full table; only the single-pid form has a fast path).

Separately, the `lsof` call that discovery already makes for cwd resolution (`batchGetProcessCwds`, `processes.ts:266`) runs `lsof -p <pids> -Ffn` with no fd filter, dumping every open fd of processes that hold hundreds. That costs ~50ms per tick on its own.

Combined status quo: ~147ms of subprocess time per 5s scan tick, which is most of the daemon's steady-state cost.

## Fix

Drop `tty` from the ps invocation and harvest it from the lsof call, tightened to:

```
lsof -a -p <pids> -d cwd,0,1,2 -Ffn
```

The pane tty appears as the `n` value of fds 0/1/2. Measured: this lsof shape costs ~18ms (816 bytes of output vs 34KB today). New total: ~29ms (ps) + ~18ms (lsof) = ~47ms per tick, a ~100ms/tick saving, with tty coming along free.

### Implementation traps (all verified, encode them)

1. **`-a` is mandatory.** Without it, lsof ORs the `-p` and `-d` selectors and enumerates every process on the machine (~110ms, output includes pids never requested). This fails silently and looks like it works.
2. **Read fds 0/1/2 only, and read all three.** Codex runs with fd2 on `/dev/null` while fd0/fd1 hold the pane tty, so any single fd is insufficient. Do not scan higher fds: ccmux's own OSC notification backend opens other panes' ttys on high fds (`src/lib/notify-osc.ts:129-130`, observed on fds 4/5/11 of a live claude process), so a blanket "any /dev/ttys\* fd" rule would mis-attribute ttys.
3. **Parser robustness.** Tolerate an `f<n>` record with an empty `n` payload (observed live), and interleaved `ftxt`/other records between `fcwd` and the fd block.
4. **Filter reordering.** The `!normalizedTty` skip (`processes.ts:191`) currently runs before lsof; it must move after. New order: ps parse, agent match, lsof (cwd + tty), tty filter, plugin-host filter, `dropWrapperParents`. This preserves the documented constraint that the plugin-host filter precedes `dropWrapperParents` (`processes.ts:211-216`). The reorder adds ~zero lsof targets in practice (command-matched processes overwhelmingly have ttys).

## Why this is safe (verified empirically)

- **Zero tty disagreements.** Across ~1200 processes, `ps` and fd-based lsof never disagree about which tty a process has. The only differences are presence/absence.
- **All 7 agent types probed live** (claude, codex, opencode, gemini incl. its npm wrapper pair, cursor-agent, copilot, pi): 15/15 processes agree between methods. The gemini wrapper parent and child resolve to the same tty under both, so `dropWrapperParents`' `${agentType} ${tty}` group key is preserved.
- **Processes that ps sees with a tty but fds do not** are pipe-stdio children (MCP servers, `typescript-language-server --stdio`, `npm exec ...`): none match any agent def (agent matching keys off basename of argv[0], `src/lib/agents.ts:1282-1294`).
- **This fixes a latent bug.** Subprocess-mode invokes (`codex exec`, `cursor-agent --print`, etc., spawned with piped stdio by `subprocess-invoker.ts:191-192`) are excluded today only because the daemonized daemon has no controlling terminal. Run the daemon in the foreground from a pane (the normal dev loop) and they inherit that pane's tty per `ps`, so discovery would create phantom pane-tracked rows fighting the real agent in that pane. Under the fd-based harvest they are correctly dropped because their fds are pipes.
- **Hook/marker matching is unaffected.** Marker-to-pane tty matching (codex/copilot/antigravity adapters) compares the marker's own tty (captured by the hook script) against the tmux pane tty; it never reads the ps-derived field.
- **Known exposure, accepted:** a setsid'd process still holding terminal fds (observed: `nvim --embed` servers) reports `??` under ps but a tty under fds. None match built-in agent defs; a custom `AgentConfig` matching a daemonized helper could newly surface a row.

Rejected alternative for the record: a pid-to-tty cache (fetch once per new pid). It has a pid-reuse hazard with no reliable guard (`startTime` is derived from second-granularity `etime` and jitters across ticks, so `(pid, startTime)` keys do not hold), leaves the 50ms lsof untouched, and adds invalidation state to a long-running daemon.

## Consumers of the tty field (for reviewer orientation)

The field feeds: the discovery filter itself (which processes become sessions, `processes.ts:191`), wrapper dedup (`dropWrapperParents`, `processes.ts:99-113`), the binder's tty-provenance pane matching (`binder/scan.ts:167-190`, has an ancestry fallback), boot migration (`binder/migrate.ts:48-53`), stale-marker cleanup (`index.ts:618-628` + `session-markers.ts:259-266`, fail-open on missing tty), the Bash `waiting -> working` escalation (`state-reconciler.ts:460-470`), and pane-tracked session creation (`index.ts:733-747`). None of these change; only the source of the value does.

## Pre-merge requirements

- [ ] Unit tests for the `batchGetProcessCwds` parser. It currently has zero coverage and a history of silent failure (see `1e75623`, where a missing `f` selector made every cwd lookup fail silently on lsof 4.99). Fixtures: codex shape (fd2 `/dev/null`, fd0/1 tty), all-pipe fds, empty `n` payload, interleaved `ftxt` records.
- [ ] Migrate `processes.test.ts:132-280`: those fixtures mock ps output with a tty column and return empty lsof output, so under the new scheme every fixture process would be tty-less and dropped; the wrapper-parent and plugin-host tests will fail until the mocks feed real `-Ffn` fixtures.
- [ ] Linux decision: the device-table cost is a BSD `ps` behavior and Linux `ps` reads tty cheaply from procfs, while lsof builds differ in silently-incompatible ways. Either gate the harvest on `process.platform === "darwin"` and keep the ps column on Linux, or verify the exact lsof invocation on Linux (including lsof 4.99) before shipping it cross-platform.
- [ ] Verify tick cost before/after with `CCMUX_PERF=1` subprocess timings.

## Context

From the 2026-07-25 performance audit. The tty column dates to the initial import ("Discover supported agent processes using ps with TTY information") with no recorded rationale for ps as the source; the `ppid` column was added later for gemini wrapper dedup. No prior decision is being reversed by re-sourcing the value.

epilande:
## Overview

Drops the `tty` column from the daemon's per-tick `ps` call and harvests the tty from the `lsof` call discovery already makes for cwd, tightened to only the fds that can hold it. Full discovery pass on a 1200-process machine: **159ms to 52ms per 5s tick**.

Resolves #54

## Motivation

Process discovery ran `ps -eo pid,ppid,tty,etime,command` every 5 seconds. On macOS, asking BSD `ps` for the `tty` column makes it build a full device-name table: ~136ms with the column vs ~55ms without, a fixed cost that restricting to specific pids does not avoid. Separately, the cwd-resolution `lsof` dumped every open fd of processes that hold hundreds, for another ~78ms.

The pane tty is the name of fds 0/1/2, so lsof can report it alongside cwd for free. That collapses most of the daemon's steady-state subprocess cost into one cheap call.

### Discovery

- **src/daemon/processes.ts** - A `DiscoveryPlatform` record (`psColumns` + `lsofArgv`) holds the whole platform choice. macOS uses `pid,ppid,etime,command` plus `lsof -a -p <pids> -d cwd,0,1,2 -Ffn` (2.6KB of output vs 59KB); everywhere else keeps the ps column and the untightened lsof. `psColumns` is the single source of truth, with every column index derived from it, so the layout is stated exactly once
- **src/daemon/processes.ts** - `parseLsofFdOutput()` replaces the inline cwd scanner: it pairs each `n` with the `f` record immediately before it, so interleaved `ftxt`/`fmem` records and empty `n` payloads (both observed live) cannot shift a name onto the wrong fd
- **src/daemon/processes.ts** - The `!tty` filter moves after lsof, since lsof is now what supplies the value. `parsePsOutput()` and `resolveDiscoveredProcesses()` are extracted as pure functions, so the load-bearing order (drop tty-less, drop codex plugin hosts, then `dropWrapperParents`) is directly testable. tty is settled before the fold runs, which makes the fold platform-blind and the "never mix the two sources" rule hold by construction
- **src/daemon/processes.ts** - `normalizePsTty()` also recognizes Linux's `?` for "no controlling terminal", which the old inline check missed

### Tests

- **src/daemon/processes.test.ts** - `parseLsofFdOutput` fixtures for the codex fd layout (fd2 on `/dev/null`, fd0/1 on the tty), all-pipe stdio, an empty `n` payload, interleaved `ftxt`/`fmem`, a tty on a high fd, non-terminal device names, the Linux `pts` form, and per-pid isolation. The parser had zero coverage before this, and a history of silent failure
- **src/daemon/processes.test.ts** - `resolveDiscoveredProcesses` tests pin the filter order in one scenario
- **src/daemon/processes.test.ts** - `parsePsOutput` and the two platform records are tested directly against both layouts, so ubuntu-only CI covers the macOS ps columns and the `-a` / `-d cwd,0,1,2` flags. Those flags fail silently when dropped, so leaving them host-covered only was the wrong place to stop
- **src/daemon/processes.test.ts** - Migrated discovery fixtures render their ps columns from the host's platform record rather than re-deriving the platform check. The wrapper-parent fixture switches to the gemini pair, since its old claude wrapper line matched no agent def and never exercised the case it claimed to cover

### Docs

- **docs/architecture.md** - New "Process discovery" section: the platform split and why, the three lsof constraints, and the filter order

## Three traps this encodes

1. **`-a` is mandatory.** Without it lsof ORs the `-p` and `-d` selectors and enumerates every process on the machine. Measured: 860 pids for a 26-pid request, 142ms. It fails silently, because the pids you asked for are all still there in the flood.
2. **Read fds 0, 1, and 2, all of them, and nothing above them.** Codex runs with fd2 on `/dev/null` while fd0/1 hold the pane tty, so no single fd suffices. ccmux's own OSC notification backend holds _other_ panes' ttys on high fds (`src/lib/notify-osc.ts`, observed on fds 4/5/11 of a live claude process), so a blanket "any tty-looking fd" rule would bind a session to the wrong pane.
3. **`-F` needs the `f` selector.** Pre-existing, now documented in place: lsof 4.99+ emits no fd-type lines for a bare `-Fn`, so `fcwd` never appears and every cwd lookup fails silently (see `1e75623`).

## Platform gate

The harvest is gated on `process.platform === "darwin"`. Linux `ps` reads tty straight from procfs, so there is nothing to win there, and Linux lsof builds vary in how they honor `-d` in ways that could not be verified here. A silent "no fds reported" would drop every session, so the conservative path keeps both the ps column and the original lsof invocation. The gate resolves to one `DiscoveryPlatform` value at module load and never spreads past it; the parse and fold below are shared.

Linux verification of the tightened lsof shape is deferred, not attempted.

## Breaking Changes

None. Discovery output is unchanged: on this machine, main and this branch return identical `(pid, agentType, tty, cwd)` rows.

One behavior change worth naming, and it is a fix: subprocess-mode invokes (`codex exec`, `cursor-agent --print`, spawned with piped stdio) are excluded today only because the daemonized daemon has no controlling terminal. Run the daemon in the foreground from a pane, as the dev loop does, and they inherit that pane's tty per `ps` and create phantom rows fighting the real agent. Under the fd harvest they are correctly dropped, because their fds are pipes.

## Testing

- [x] `bun test` - 3030 pass, 0 fail (processes.test.ts: 23 tests to 42)
- [x] Non-darwin path verified by rerunning `processes.test.ts` with `process.platform` stubbed to `"linux"` - 42 pass
- [x] Live ps-vs-lsof comparison across 1203 processes: 94 agree, **0 conflicts**. Differences are presence/absence only. ps-with-tty-but-no-fd-tty is all pipe-stdio helpers (`npm exec`, `typescript-language-server --stdio`, MCP servers) matching no agent def; fd-tty-but-no-ps-tty is `nvim --embed` (setsid'd), also matching no agent def
- [x] Live discovery on this branch: 8 agent processes, every `tty` matches `ps -o tty= -p <pid>`, 0 mismatches. The ccmux daemon itself (no controlling terminal, fds not on a tty) is excluded
- [x] Live A/B: main's `discoverAgentProcesses` and this branch's return identical rows
- [x] Per-tick timing, interleaved runs: main 159ms median, branch 52ms median
- [ ] Linux verification of the tightened lsof invocation (deferred; Linux keeps the old code path)


epilande:
E2E verification recording: a pane-tracked gemini session created and bound purely through the lsof-sourced tty, driven idle -> working -> waiting -> idle under the PR build. Discovery A/B against main returned identical (pid, agentType, tty, cwd) rows for the live process set, including a codex with fd2 on /dev/null, and the phantom pipe-stdio session reproduced under main is absent under this branch. Measured discovery pass: 168ms (main) -> 55ms (branch).

https://github.com/user-attachments/assets/e4963b28-339e-49a9-89ae-76f43621f4d2

epilande:
Review pass turned up a real hazard in moving the tty to lsof: `batchGetProcessFdInfo` swallowed every failure and returned an empty map, so on macOS a dead lsof gave every row `tty: null`, the tty filter dropped all of them, and discovery returned `[]` without throwing. That is the one answer the scan loop acts on destructively, so a single bad tick would wipe every session and unlink every hook marker. It now separates "lsof ran" (non-zero exit with records, the routine pid-died-mid-call case, still degrades gracefully) from "lsof did not" (spawn threw, or non-zero with nothing parseable), and throws `ProcessDiscoveryError` on the latter where lsof is the tty source. Marker cleanup also picked up the same two-scan hysteresis `binder/cleanup.ts` already had for the PID-liveness check, so a one-tick lsof omission of a live pid no longer deletes that session's marker. Plus fixtures for all three lsof failure shapes, exercised against both platform records explicitly rather than whichever one the host happens to be.

PR #58, opened 2026-07-26
perf(daemon): Coalesce and merge git enrichment, take git out of the PR sweep

### LINKED ISSUE #53: perf(daemon): Coalesce and merge git enrichment on the SSE init path, fix branch TTL vs sweep mismatch
## Problem

The SSE `init` event, which a freshly launched picker/sidebar paints its first frame from, awaits full git enrichment for every visible session:

```ts
// src/daemon/server.ts:1445
sessions: await this.enrichSessions(matched),
```

`enrichSession` (`src/daemon/server.ts:397-429`) awaits `getGitBranch` then `isGitWorktree`, each a separate `git rev-parse` spawn behind a 30s-TTL cache. Three compounding issues:

1. **No in-flight coalescing.** The caches store the resolved value after the await, so `Promise.all` over N sessions sharing one cwd fires N duplicate concurrent git spawns (two waves of N, one per field).
2. **Two spawns where one suffices.** Branch and git-dir are available from a single `git rev-parse --abbrev-ref HEAD --git-dir` call.
3. **The 2-minute PR sweep re-derives git every cycle.** `sweepBranchPRs` (`src/daemon/server.ts:317-322`) loops `await this.enrichSession(session)` serially over all visible sessions. `BRANCH_CACHE_TTL_MS` (30s, `server.ts:92`) is always expired by sweep time (`PR_SWEEP_INTERVAL_MS` = 2min), so every sweep pays 2 serial git spawns per visible session, forever, even on a fully idle machine. The comment claiming "Sweeps are cheap (cached reads; only expired keys spawn gh)" is true of `gh` and false of `git`.

Measured (41 sessions, one repo, warm git): init enrichment costs ~123ms today. With the fixes below: ~8ms same-repo, ~61ms worst case (all-distinct cwds). The sweep burst is ~740ms of serial spawning every 2 minutes.

The same fan-out sits on `GET /sessions` (`server.ts:770`), whose consumers (`ccmux show`, `ccmux screen`, `ccmux debug`) never read `gitBranch`/`isWorktree`; the fixes below make that path cheap without an API change.

## Fix

Three changes, all in `src/daemon/server.ts`:

### 1. Coalesce in-flight lookups (promise cache)

Copy the existing in-house precedent: `VersionResolver` keeps an `inflight: Map<string, Promise<...>>` beside its value cache (`src/daemon/version-resolver.ts:385,410-423`) and deletes the entry in a `finally` (`:448`). Two correctness notes:

- Keep the try/catch inside the memoized function (as `getGitBranch`/`isGitWorktree` already do at `server.ts:353-364,376-388`) so a rejected promise can never be cached.
- Stamp `expiresAt` at resolution time, not from a `Date.now()` captured before the spawn (current code at `:348` starts the TTL clock early).

### 2. Merge the two git spawns

`git -C <cwd> rev-parse --abbrev-ref HEAD --git-dir` returns both values, one per line, in argument order. Verified output shapes:

| State                              | stdout                                    | Result                                           |
| ---------------------------------- | ----------------------------------------- | ------------------------------------------------ |
| main checkout                      | `main` / `.git`                           | branch, not worktree                             |
| worktree on branch                 | `<branch>` / `.../.git/worktrees/<name>`  | branch, worktree                                 |
| worktree detached                  | `HEAD` / `.../.git/worktrees/<name>`      | literal `HEAD` (pre-existing behavior), worktree |
| non-repo / deleted cwd             | exit 128, no stdout                       | null, false                                      |
| **fresh `git init` (unborn HEAD)** | **exit 128, stdout = single line `HEAD`** | must be null, false                              |

The unborn-HEAD row is the trap: parse stdout without gating and every freshly initialized repo shows a phantom `HEAD` branch. **Gate on exit code 0 AND exactly two stdout lines.** The `gitDir.includes("/worktrees/")` test (`server.ts:384`) keeps working unchanged; the merged call's git-dir output is byte-identical to the standalone call's.

### 3. Fix the TTL/sweep mismatch

Raise `BRANCH_CACHE_TTL_MS` above `PR_SWEEP_INTERVAL_MS` (or have the sweep skip the git half and only refresh PR data). Fix 1 does not help the sweep (it is a serial loop, nothing concurrent to coalesce); fix 2 only halves it. Only this eliminates the guaranteed periodic burst. Branch changes surface within one TTL; a longer TTL trades a few minutes of staleness on a value that changes rarely and self-heals on any `session_updated` re-enrichment.

## Explicitly out of scope

Deferred/stale-on-init enrichment (init with core fields, enrich in background). Investigated and rejected for now: it requires a branch-change re-broadcast path that does not exist (`onBranchPRsChanged` at `server.ts:324-345` is keyed by branch and only covers `branchPRs`), idle sessions would never heal a stale init (broadcasts are change-gated, `sessions.ts:623`), and a late-arriving branch re-budgets `fitProjectCell` (`session-columns.ts:591-644`), visibly shifting text on every row. The fixes above get within ~1ms of it with zero render change. If revisited later: the SSE buffering added for init-first ordering already guarantees follow-up events land after init.

## Safety notes for the implementer

- `gitBranch` is not a grouping or sort key anywhere in the TUI (`getGroupKey` keys only on project/cwd/tmuxTarget), and every enriched field is null-safe at every read site. Not directly relevant to these fixes (the first frame does not change), but useful background.
- The `init` event must remain the first event a client sees; these fixes only shrink the await window and do not reorder anything.

## Test requirements

- Unit: N concurrent `enrichSession` calls for one cwd spawn exactly one git process (spy on spawn).
- Unit: unborn-HEAD repo yields `branch: null, isWorktree: false` (fixture: `git init` in a temp dir).
- Unit: rejected/failed spawn is not cached; next call retries.
- Behavior: sweep no longer spawns git when the branch cache is fresh.
- Benchmark before/after with `CCMUX_PERF=1` or a spawn counter, at 20+ same-cwd sessions.

## Context

From the 2026-07-25 performance audit. Pre-release history confirms this is accretion, not design: an earlier internal perf audit had already flagged the awaited init enrichment as a first-paint defect (it was tiered low and never done), the PR resolver got in-flight dedup and a concurrency cap while the git caches next to it never did, and the 30s TTL predates the 2-minute sweep with no commit ever reconciling the two.

epilande:
## Overview

Coalesces and merges the git lookups behind session enrichment, and takes git out of the PR sweep so an idle daemon stops re-spawning it every two minutes.

Resolves #53

## Motivation

A freshly launched picker paints its first frame from the SSE `init` event, which awaits full git enrichment for every visible session. That enrichment fired two `git rev-parse` spawns per session behind caches that only stored resolved values, so `Promise.all` over 41 sessions sharing one repo spawned 82 git processes in two waves instead of the one the answer actually needs.

The 2-min PR sweep had the same problem from the other end. It re-enriched every visible session purely for the `prResolver.get()` side effect, and the 30s git TTL was always expired by sweep time, so every sweep re-derived git for every distinct visible cwd, forever, even on a fully idle machine. The comment claiming "Sweeps are cheap (cached reads; only expired keys spawn gh)" was true of `gh` and false of `git`.

Measured on 41 same-repo sessions with warm git: **123ms -> 8.8ms** on the init path (82 spawns -> 1), and periodic sweep spawns drop to **zero**.

### Git enrichment

- **src/daemon/server.ts** - Replaces `getGitBranch` / `isGitWorktree` (and their two caches) with a single `getGitInfo`, resolved by one `git rev-parse --abbrev-ref HEAD --git-dir` call that returns both values, one line each
- **src/daemon/server.ts** - Adds a `gitInfoInflight` promise map beside the value cache, mirroring `VersionResolver`, so concurrent callers for one cwd share a single spawn. The memoized resolver never rejects (a poisoned promise would otherwise be shared by every coalesced caller) and stamps `expiresAt` after the spawn rather than before it
- **src/daemon/server.ts** - Gates parsing on exit 0 AND exactly two stdout lines. A freshly `git init`ed repo has an unborn HEAD: git exits 128 but still prints a lone `HEAD` line, which an ungated parse would render as a phantom branch on every new repo

### PR sweep

- **src/daemon/server.ts** - `sweepBranchPRs` no longer routes through `enrichSession`. It reads the branch straight from the git cache (stale entries included, since a stale key still names the right branch far more often than not), falls back to the log-derived branch, and calls `prResolver.get` directly, so it never spawns git and is now synchronous
- **src/daemon/server.ts** - Extracts `effectiveCwd`, which the sweep needed and which two other call sites were already re-deriving by hand
- **src/daemon/server.ts** - `sweepBranchPRs` rotates its starting position through the visible-session list by one session per sweep (`sweepOffset`). Review caught that the original fixed iteration order let `PRResolver`'s concurrency cap (4 in flight) always hand its slots to the same leading sessions, starving every key past them indefinitely; rotating means every key gets a guaranteed refresh attempt within `len` sweeps
- **docs/architecture.md** - Documents that the sweep touches PR keys off the cache instead of re-enriching, and the rotation's fairness bound

The issue offered two options here, raising `BRANCH_CACHE_TTL_MS` above `PR_SWEEP_INTERVAL_MS` or dropping the git half from the sweep. I started with the TTL bump and backed it out: raising the TTL is a global freshness knob turned to fix a sweep-local problem. It would have left the TUI showing a stale branch (and therefore stale PR data) for up to 5 min after a `git checkout`, and the justification I wrote for it ("re-resolves on the next `session_updated`") was simply false, since nothing invalidates the cache on re-enrichment. It also only reduced the burst rather than removing it: the entries are primed in lockstep by the same sweep, so they expire in lockstep too. Fixing the sweep instead sends periodic git spawns to zero and leaves the TTL at 30s, so branch freshness is untouched.

Accepted trade: because the sweep reads branch off `gitInfoCache` instead of re-deriving it, an idle session's branch label no longer self-heals off a `git checkout` run in its pane on its own; that only happens on the pane's own next organic event or a fresh SSE connect, not on the periodic sweep.

### Tests

- **src/daemon/server.test.ts** - Adds `stubGitSpawn` (records git argv, answers non-git spawns) and covers: 20 concurrent enrichments spawning git once, one merged spawn resolving both fields, an unborn-HEAD repo (real `git init` fixture) yielding `null` / `false`, a thrown spawn staying uncached while a non-zero exit is cached, the sweep spawning no git even against a cold cache, and the sweep keying off the cached branch when one is warm
- **src/daemon/server.test.ts** - Adds a fairness test for the sweep rotation: 10 distinct branch keys against a real `PRResolver` (its actual 4-in-flight cap, not a stand-in), asserting every key gets a refresh attempt within `len` sweeps

## Breaking Changes

None

## Testing

- [x] `bun test` - 3019 pass, 0 fail
- [x] Coalescing test verified to fail without the in-flight map (stashed the fix, test went red)
- [x] Benchmarked the before/after spawn pattern at 41 same-cwd sessions with warm git: 126.6ms / 82 spawns -> 8.8ms / 1 spawn
- [ ] Live picker e2e against the shared daemon (handled separately; the daemon was not restarted from this branch)


epilande:
## e2e verification

Ran this against a live daemon swapped to the branch build, with the main-repo build as an A/B baseline. Git spawns were counted exactly, via a PATH shim that logs every `git` argv the daemon forks (it also catches the `gh` resolver's internal git calls, so the numbers below count only the enrichment `rev-parse`).

**Spawn counts** (12 sessions across 10 distinct cwds):

| build       | cold SSE init                                                   | per 2-min sweep, idle machine |
| ----------- | --------------------------------------------------------------- | ----------------------------- |
| main        | 22 (two calls per session, no per-cwd dedup)                    | 20                            |
| this branch | 9 (one merged call per distinct cwd; the 10th was already warm) | 0                             |

The sweep really did fire inside the measured window, so the zero is meaningful: sweep #1 landed at boot+120s and its `gh pr list` shows up in the shim log 3s later, with no `rev-parse` next to it. Cold-cache init latency went from 59.5ms to 31.9ms; warm reconnects are ~6ms.

**Correctness.** Every enriched field is identical to the baseline:

- Branches correct across 6 distinct repos, and both worktree rows keep `isWorktree: true`, so the picker still renders the `+` suffix.
- A freshly `git init`ed cwd (unborn HEAD) shows no phantom `HEAD`. The trap is real, confirmed directly: `rev-parse --abbrev-ref HEAD --git-dir` there exits 128 but still prints a lone `HEAD` on stdout, so the exit-0-plus-two-lines gate is doing the work.
- A non-repo cwd renders with no branch from git and no errors.
- PR badges still appear and refresh (#61, #63 and #58 all resolved, including one that filled in after the background lookup landed on a cold daemon).
- A `git checkout -b` surfaces to a fresh connect after 30.1s, i.e. exactly the 30s TTL this PR deliberately kept.

**One behavior worth knowing, not a regression.** An already-attached picker does not repaint an idle session's branch change, even across a sweep boundary. The daemon serves the new branch to any fresh connect, but a branch-only change broadcasts no `session_updated`. I A/B'd it on the main build and got identical behavior (still stale 203s after the flip), so this is pre-existing and untouched by this PR.

Static checks: typecheck, knip, build, and `bun test` (3018 pass, 0 fail).

A screencast of the picker under the branch daemon (worktree `+`, unborn-HEAD repo, non-repo cwd, live branch flip) was recorded and frame-verified, but there was no logged-in browser session available on this run to attach it inline, so it is saved locally at `~/Downloads/ccmux-pr58-verification.mp4`.

epilande:
E2E verification recording for the fixture scenarios: worktree branch with the + suffix, a freshly git-init-ed repo (unborn HEAD) rendering with no phantom HEAD branch, a non-repo cwd rendering cleanly, and a live git checkout -b surfacing within the 30s TTL on a fresh connect. Full measured results (git spawn counts, sweep behavior, latency) in the comment above.

https://github.com/user-attachments/assets/90da398e-22d6-465c-ac8a-292d2eb974b8

PR #59, opened 2026-07-26
perf: Small latency wins (parallel link passes, redundant lsof skip, SSE stringify-once, capture debounce and cache)

### LINKED ISSUE #55: perf: Small latency wins (parallel link passes, redundant lsof skip, SSE stringify-once, capture debounce and cache)
## Overview

Batch of five small, independent latency wins from the 2026-07-25 performance audit. Each is a contained change with low regression risk; they can land as one PR or be cherry-picked individually. None changes behavior, only when/how often work runs.

### 1. Parallelize the six agent link passes

`src/daemon/index.ts:644-649` awaits `linkCodexSessions`, `linkOpenCodeSessions`, `linkCursorSessions`, `linkPiSessions`, `linkAntigravitySessions`, `linkCopilotSessions` one after another inside the scan tick. The passes cover disjoint agent types with no shared mutable state, and each has its own I/O fan-out (`adapters/link.ts:54` awaits a `getPaneHostingPid` per marker pid). `Promise.all` is a direct substitution. Win: removes serialized I/O latency from the tick's critical path whenever multiple agents have markers.

### 2. Skip redundant per-tick lsof for already-resolved sessions

`src/daemon/index.ts:754-758`: `createOrUpdatePaneTrackedSessions` calls `resolveNativeSessionId` for every pane-tracked process whose agent defines `sessionFilePattern` (pi and copilot in practice), spawning `lsof -p <pid> -Fn` every tick even when the existing session already carries a `nativeSessionId`. Look up the existing session by pane first and skip the spawn when the id is already resolved. Win: one fewer subprocess per such session per tick, forever.

### 3. Serialize SSE events once per broadcast

`src/daemon/server.ts:1971`: `broadcastEvent` calls `sendToClient` per client, and each call runs `JSON.stringify(event)` on the same object. Stringify once outside the loop and pass the encoded frame. Win: O(clients) redundant serialization becomes O(1) per event; matters most during event bursts with several attached picker/sidebar clients.

### 4. Debounce the Preview selection capture

`src/tui/components/Preview.tsx:285`: the preview pane captures the selected session's pane on every selection change with no debounce. Holding <kbd>j</kbd> through 40 rows spawns 40 `tmux capture-pane` calls whose results the existing stale-resolve guard then discards. An ~80ms debounce collapses that to one or two captures with no perceptible preview lag. Win: eliminates a subprocess storm during rapid navigation, which is exactly when the TUI should feel snappiest.

### 5. Cache pane content across search keystrokes

`src/tui/store.ts:361-395`: the whole-session search effect fires 250ms after each query change and runs `Promise.all` of `tmux capture-pane` over all sessions, with no reuse between runs. Typing a query with natural pauses re-captures every pane per pause. A pane-to-content cache with a 2-3s TTL makes all but the first pause free; pane content staler than a couple of seconds is irrelevant for search ranking. Win: at N sessions, typing costs N captures once instead of N per pause.

## Non-goals

- The scan-tick ps/lsof restructuring and the SSE init enrichment fixes are tracked separately (bigger changes with their own risk profiles).
- The marker-file double-read per tick (`src/daemon/index.ts:579` + `:634` both walk `~/.config/ccmux/session-pids/`) is a candidate sixth item, but it is ~0.3ms at typical marker counts; include only if touching `session-markers.ts` anyway (have `cleanupStaleMarkers` consume the map `refreshMarkerCache` just built; cleanup still needs its own readdir for the `.tmp` sweep).

## Testing

- [ ] Item 1: existing link-pass tests still pass; add one test asserting all passes run when one rejects (`Promise.all` vs `allSettled` decision: a single pass throwing currently aborts the remaining passes mid-tick, so `allSettled` with per-pass error logging is the safer substitution)
- [ ] Item 2: test that a session with `nativeSessionId` set spawns no lsof (spy on the spawn)
- [ ] Item 3: test that N clients receive byte-identical frames from one stringify (spy on `JSON.stringify` or refactor to make the frame observable)
- [ ] Item 4: `testRender` + `createMockKeys` rapid j/k sequence, assert capture call count with mocked `capturePane`
- [ ] Item 5: store test asserting a second query within the TTL reuses captured content (mock `capturePane`, count calls)
- [ ] TUI items (4, 5): verify rendering in a detached tmux session per AGENTS.md before declaring done

epilande:
## Overview

Batches five small, independent latency wins from the 2026-07-25 performance audit: concurrent agent link passes, a skipped redundant lsof spawn, single-stringify SSE broadcasts, a debounced Preview capture, and a TTL cache for search pane-content captures.

Resolves #55

## Motivation

None of these change behavior, only when and how often work runs. Each was a contained fix with low regression risk, called out individually in the audit: serialized link passes add up when several agents have markers, a per-tick lsof spawn for already-resolved sessions is pure waste, SSE broadcasts re-serialize the same event once per client, holding <kbd>j</kbd>/<kbd>k</kbd> spawns a `tmux capture-pane` per row the preview never gets to render, and typing a search query re-captures every visible pane on every keystroke pause.

### Daemon

- **src/daemon/index.ts** - `runLinkPasses` runs the six per-agent marker link passes concurrently via `Promise.allSettled` instead of six sequential `await`s, so one pass throwing no longer aborts the rest mid-tick (each pass now also logs at most once per distinct error message instead of every tick). `createOrUpdatePaneTrackedSessions` skips the lsof spawn behind `resolveNativeSessionId` when the pane's existing session already has a `nativeSessionId` for the same pid; that skip is time-bounded to 60s (`SessionManager.NATIVE_SESSION_ID_CACHE_TTL_MS`) rather than permanent, so a pane-tracked session running without hooks that opens a new session file under the same pid (e.g. `/new`) still gets re-resolved instead of pointing at the dead session forever.
- **src/daemon/sessions.ts** - Adds `SessionManager.getResolvedNativeSessionId(paneId, agentType, pid)`, which owns the pane-tracked id derivation and the pid/nativeSessionId check so the daemon doesn't need to know the internal id scheme.
- **src/daemon/server.ts** - `broadcastEvent` now `JSON.stringify`s the event once and hands every client the same encoded frame (`sendFrame`), instead of re-stringifying per client inside `sendToClient`.

### TUI

- **src/tui/components/Preview.tsx** - Debounces the selection-change pane capture by ~80ms (`SELECTION_CAPTURE_DEBOUNCE_MS`), so rapid `j`/`k` navigation settles on one capture instead of firing one per row.
- **src/tui/store.ts** - Adds a per-pane TTL cache (`paneContentCache`, default 2.5s, injectable via `searchPaneCacheTtlMs`) for the whole-session search effect, reused across debounced query changes within the TTL and pruned to the current session list's panes on every run.

### Tests

- **src/daemon/index.link-passes.test.ts** - `runLinkPasses` isolation: all six passes run even when one rejects, rejections are logged with the pass name and reason, and a clean run logs nothing.
- **src/daemon/index.pane-tracked-lsof-skip.test.ts** - Spies on the lsof spawn to confirm it's skipped when a session's `nativeSessionId` is already resolved for the same pid, and still fires on pane hand-off or first resolution.
- **src/daemon/server.test.ts** - Confirms `broadcastEvent` calls `JSON.stringify` exactly once regardless of client count and that every client receives a byte-identical frame.
- **src/tui/components/Preview.test.tsx** - Drives a rapid sequence of selection changes and asserts the capture count stays low instead of one per row.
- **src/tui/store.test.ts** - Confirms a second query within the TTL reuses cached pane content (no new `capturePane` call) and that content re-captures once the TTL elapses.

## Breaking Changes

None

## Testing

- [x] `bun test` passes (3029 tests, 0 failures)
- [x] Unit tests added per item covering allSettled isolation, no-lsof-when-resolved, single-stringify broadcast, capture debounce call count, and search TTL reuse/expiry
- [x] Rendering verified in a detached tmux session (`ccmux picker` against the shared daemon): live session list rendered correctly, rapid `j`/`k` settled the preview on the correct row with real captured content, and search filtering (`/ccmux`, adding/removing a character) updated results correctly through the debounce/cache path


epilande:
## E2E verification

Ran the full verify-pr pass against this branch: static tiers, then the PR daemon swapped in live on 2269 with `CCMUX_PERF=1`, then the PR build's picker driven in a detached tmux session. Each of the five items was measured A/B against the `main` build on the same machine with the same session roster.

**Tier 0**: `typecheck`, `knip`, `build`, `bun test` all green (3022 pass, 0 fail). The five targeted test files pass on their own too (336 pass, 0 fail).

### Item 1, concurrent link passes

Swapped the daemon to this build with live hook-tracked sessions attached. Every session kept its `nativeSessionId` and `logPath` across the swap, byte-identical to the pre-swap roster. A fresh Claude session started after the swap linked to native tracking with both fields inside one tick. The `allSettled` isolation itself (one pass throwing without aborting the rest) is covered by the new unit tests only, not reproduced live.

### Item 2, redundant lsof skip

Measured with the daemon's own `subprocesses/scan` counter, with one pane-tracked Pi session in the roster (Pi is one of the few agents that actually reaches the lsof path, since Cursor has no `sessionFilePattern` and Codex returns early).

| build   | `lsof-session` per scan                                                                    |
| ------- | ------------------------------------------------------------------------------------------ |
| main    | 1.0, sustained across every window                                                         |
| this PR | 0.2 in the first window (the one real resolution), then absent (0.0) in every window after |

So 5 spawns per 5-scan window becomes 0 once resolved, which is the intended shape.

Worth noting for review: on the skip path a pane-tracked session's existing `nativeSessionId` now gets re-asserted every tick where it previously passed `undefined`. That is a genuine no-op, `setNativeSessionId` early-returns on an identical id and `createPaneTrackedSession` guards each field on inequality. Confirmed empirically: over roughly 35 scans the pane-tracked row emitted exactly 1 `session_updated`.

### Item 3, SSE stringify-once

Two simultaneous `curl -N` clients on `/events` for 70 seconds, covering 29 events (1 init, 1 session_created, 22 session_updated, 5 heartbeat). Both captures were byte-identical, same sha256.

### Item 4, Preview capture debounce

Simulated a held `j` through 12 rows (presses 40ms apart), then let the preview settle.

| build   | `tmux capture-pane` spawns        |
| ------- | --------------------------------- |
| main    | 8 (one per session row landed on) |
| this PR | 1                                 |

Both builds settled on the same correct row with real captured pane content, so the debounce costs nothing in correctness.

### Item 5, search pane-content TTL cache

Typed `ccmux` one or two characters at a time with roughly 1.2s pauses, 9 sessions in the list.

| pause | main       | this PR                      |
| ----- | ---------- | ---------------------------- |
| 1     | 9 captures | 9 captures                   |
| 2     | 9          | 0 (cache hit)                |
| 3     | 9          | 0 (cache hit)                |
| 4     | 9          | 9 (TTL elapsed, re-captured) |
| total | 36         | 18                           |

The re-capture landing exactly on the fourth pause matches the 2.5s TTL. Filtering stayed correct at every keystroke in both builds (typing `fixture` narrowed 9, 9, 9, 8, 4, 4, 4 with identical final results).

### Recording

Recorded a 64s screencast of this build's picker doing the rapid `j`/`k` navigation and the paused search typing. Frames confirm the preview settling on the correct row with real content, and search filtering to 5/9 mid-query. The inline video upload needs a logged-in browser session, which was not available here, so the clip is saved locally rather than attached. The daemon items above were verified separately against this same daemon and are not what the clip shows.

Daemon was restored to the `main` build afterwards and the roster verified identical to the pre-swap state.

epilande:
E2E verification recording under the PR build: a 12-press j burst settles the preview on the correct row with one debounced capture instead of eight, and typing a search query with natural pauses filters correctly at every keystroke while the 2.5s TTL cache absorbs the intermediate re-captures (A/B: 36 -> 18 pane captures for the same query). Measured details in the comment above.

https://github.com/user-attachments/assets/9e8fe82e-ffa1-44c4-a8c7-e11902d0b8d2

PR #61, opened 2026-07-26
perf(sidebar): Cut the cost of background sidebars

epilande:
## Overview

Cuts the cost of background sidebars. Sidebars in off-screen windows stop animating, ticking, and probing tmux; the `window-resized` hook no longer boots a ccmux process per window; and the width persister checks its cheap gates before spawning. At 25 windows with one working agent, total sidebar CPU drops from 72.8% of a core to 10.5%.

## Motivation

`ccmux sidebar` runs one full TUI process per tmux window, but at most one window per attached session is on screen. The other N-1 processes did full work anyway:

- Every spinner frame is a full-buffer redraw, so each background sidebar burned ~3% of a core whenever any session was `working`, and all of that escape output funnels through the single-threaded tmux server. This is the "sidebar feels laggy" report.
- A client attach or terminal resize fired the `window-resized` hook once per window, each firing booting a full ccmux CLI that resized every sidebar on the server: N boots x N resize-panes (measured: 635 subprocess spawns and a 3.3s stall at 25 windows).
- Every j/k in one sidebar broadcast the selection over SSE to all sidebars, and each spawned `tmux list-panes` to decide whether to flash a pane that is not on screen.

### Visibility gating (sidebar only, picker untouched)

- **src/tui/utils/window-visibility.ts** - New primitive: visible iff the sidebar's own window is active and its session attached, from one `tmux display-message`. Refreshes on `active_pane` SSE events (250ms debounce), terminal dimension changes, and a 30s safety poll. Fails open: any error means visible
- **src/tui/utils/tmux-window-state.ts** - The `display-message` fetch extracted from `sidebar-width.ts`, shared by the width persister and the visibility gate
- **src/tui/utils/useStatusIcon.ts** - `setSpinnerPaused()`: pauses the shared spinner interval while preserving refcounts; resume bumps the frame so the UI repaints immediately
- **src/tui/utils/pane-flash.ts** - Flash scheduler extracted from App; an invisible sidebar drops the request before arming its debounce, so no `list-panes` spawn
- **src/tui/utils/idle-gc.ts** - Collects the heap 5s after a sidebar goes hidden. A gated background sidebar allocates too little for JSC's allocation-driven GC to ever run, so it can strand its boot-time high-water mark (observed: 9 of 25 processes at 126-155MB vs the ~95MB settle)
- **src/tui/App.tsx** - Wires visibility into the spinner pause, idle GC, the tick (forced to the 10s cadence while hidden, immediate catch-up on return), and the flash path

### Resize hook

- **src/commands/sidebar.ts** - `window-resized` hook body is now pure shell scoped to `#{hook_window}`: two short-lived tmux clients per firing, zero ccmux boots, touches only its own window. The `--resize` CLI path stays global so pre-upgrade hooks persisted in running tmux servers keep working, and `parseResizeHook` recognizes both bodies so the next `--apply-width` replaces the old one

### Width persister

- **src/tui/utils/sidebar-width.ts** - Gate evaluation split into local gates (prefs mtime, configured-width no-op) and window gates (ceiling, focus). Local gates run first, so the common rejected path no longer pays a `tmux display-message` spawn. Gate semantics unchanged from #44

## Measurements

Isolated harness (own tmux server, daemon, and config home): 25 windows across 5 sessions, one attached 200x50 client, one fake `working` agent, a PATH shim counting every tmux/ccmux spawn. Same scripts, both builds.

| Metric                              | main (v1.2.1)                                | this branch                          |
| ----------------------------------- | -------------------------------------------- | ------------------------------------ |
| Steady-state CPU, 1 working session | 43.8 CPU-s/60s (72.8% core)                  | 6.3 (10.5%)                          |
| Per-sidebar                         | all 25 at ~1.75s                             | visible 2.27s, background 0.15-0.18s |
| Steady-state CPU, all idle          | 12.1% core                                   | 8.2%                                 |
| tmux server CPU (same window)       | 0.97s                                        | 0.27s                                |
| Resize storm, 25 windows            | 635 spawns, 25 ccmux boots, 3.25s to quiesce | 75 spawns, 0 boots, 0.77s            |
| Spawns per navigation keypress      | ~22.5                                        | ~1                                   |

CPU numbers reproduced across two interleaved trials per build to within 1%. An interleaved RSS A/B found no mean memory regression (the idle-GC commit addresses the one dispersion effect it did find). New steady-state cost: 2 `display-message` spawns per sidebar per minute (the safety poll).

Real-workload A/B (25 windows across 7 sessions on a live server, one continuously working agent, back-to-back 30s samples per build after a 60s settle):

| Metric                 | main (v1.2.1)                      | this branch        |
| ---------------------- | ---------------------------------- | ------------------ |
| Fleet CPU, 25 sidebars | 31.8 CPU-s/30s (106% of a core)    | 4.3 (14%)          |
| Per background sidebar | ~1.27s (4.2% each)                 | ~0.13s (0.4% each) |
| Visible sidebar        | ~1.3s                              | ~1.1s              |
| tmux server CPU        | 1.35s (4.5%)                       | 0.32s (1.1%)       |
| Fleet RSS              | 3.03GB (121MB mean, 90s post-boot) | 2.33GB (93MB mean) |

Single samples with uncontrolled load, so indicative rather than rigorous; the isolated harness numbers above are the controlled comparison.

Functional checks on the live builds: the visible sidebar's spinner animates normally; background spinners freeze; switching windows swaps the roles both directions in ~1.5s; the resize storm re-pins all sidebars to their configured width.

## Breaking Changes

None. tmux servers still holding the old `window-resized` hook body keep working and self-heal to the new one on the next width change.

## Testing

- [x] `bun run typecheck` clean, 3132 tests pass (22 new: visibility transitions, spinner pause/resume, flash gating, hook string, gate ordering, idle GC)
- [x] `#{hook_window}` population and format escaping verified live on tmux 3.6a in an isolated server
- [x] Before/after measurement passes on the isolated 25-window harness (numbers above)
- [x] Live render checks: detached (hidden) sidebar renders correctly with frozen spinner; attaching a client resumes animation; forced GC survival verified

### Fixes

- **src/tui/utils/sidebar-width.ts** - A window resize re-pins the sidebar to the configured width, and that settle was rejected without refreshing `lastWindowWidth`, so the next genuine drag read as coinciding with a resize and was silently dropped (the first drag persisted on main). A rejection with quiet prefs now still pays for the tmux query to keep the baseline honest; fresh-prefs propagation echoes keep the zero-fetch fast path. Regression test included.
- **src/tui/App.tsx** - The daemon suppresses `active_pane` for ccmux-titled panes, so a sidebar that is its own window's active pane never received the event that unhides it and stayed gated for up to the 30s poll while being used. A keypress now refreshes visibility, debounced to one spawn per burst.
- **src/tui/App.test.tsx** - Sidebar tests constructed the real visibility gate against ambient `TMUX_PANE`, so results depended on the runner's tmux state (a detached-session pane made them fail). `fetchWindowState` is now pinned to a visible window, with the module spread so pure exports stay real.

### Cleanup

- **refactor(tui)** - Collapses the tick visibility effect's duplicated body into a shared `runTick`, extracts the `isWithinPrefsQuietPeriod` predicate used by both gate sites, pairs `stopSpinnerInterval` with the existing `startSpinnerInterval`, enumerates all four visibility re-check signals at the `createWindowVisibility` call site, and moves `parseWindowState` coverage into its own `tmux-window-state.test.ts`.

Suite grows to 3134 tests; typecheck clean; render checks re-verified live.

### DIFF
--- a/src/commands/sidebar.test.ts
+++ b/src/commands/sidebar.test.ts
@@ -4,6 +4,7 @@ import {
   parseAutoOpenHook,
   parseSidebarPaneIds,
   parseResizeHook,
+  resizeHookCommand,
   spawnDelaySeconds,
   ccmuxPortEnvPrefix,
   sidebarSpawnCmd,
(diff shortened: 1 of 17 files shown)


epilande:
E2E screen recording (sped up): the picker built from this branch tracking a live Claude Code session through idle, working, waiting (permission prompt), and back to idle. This exercises the shared state-rendering paths the PR touched (status-icon spinner module, tick scheduling) in their visible-window behavior. Representative run; the gating change affects all agents' rows. The hidden-window half (frozen spinner, slow tick, idle GC) cannot appear on camera by definition and is verified by the before/after measurements in the PR description.

https://github.com/user-attachments/assets/5461d7d6-ffae-45bb-aa18-e59b987d9b0b

PR #62, opened 2026-07-27
feat(agents): Add built-in support for oh-my-pi (omp)

### LINKED ISSUE #60: [Feature]: Support oh-my-pi (omp)
> *This was generated by AI during triage.*

## Problem / motivation

[oh-my-pi](https://github.com/can1357/oh-my-pi) (`omp`) is an actively developed hard fork of Pi (badlogic's pi-mono). Running multiple `omp` sessions in tmux, ccmux doesn't see them at all: the process doesn't match any built-in agent, so there's no state tracking, no waiting notifications, and no session jumping.

## Proposed solution

Add a built-in `omp` adapter, structurally mirroring the Pi adapter (`adapters/pi` + `src/plugins/pi/ccmux.js`). Research against the omp source and live verification (omp 17.0.8 / 17.1.3, macOS arm64) makes this mostly mechanical, with a few omp-specific quirks:

- **Process detection**: omp sets `process.title = "omp"` in its CLI entrypoint, and the distributed binary's argv[0] basename is already `omp`, so `processMatch: /^omp$/i` anchors cleanly. Its npm package is the *scoped* `@oh-my-pi/pi-coding-agent`, which pi's own `commandPatterns` (`/pi-coding-agent[/\\]dist[/\\]cli\.js/i`) also matches — the omp def must sit **before** pi in `BUILTIN_AGENTS` and require an `oh-my-pi` path component in its patterns.
- **Hook mechanism**: omp keeps Pi's extension API (`session_start`, `before_agent_start`, `agent_start`/`agent_end`, `session_shutdown`) and auto-discovers `*.js` extensions from `<agent dir>/extensions/`. One session per process, same 1:1 marker correlation as Pi.
- **Approval waits (new vs Pi)**: unlike Pi, omp HAS a tool-approval pause (`tools.approvalMode`, per-tool `prompt`, safety overrides) and emits observational `tool_approval_requested` / `tool_approval_resolved` events — but only when an extension subscribes to them (`hasHandlers` check in `wrapper.ts`), so the ccmux extension both enables and observes them. This gives omp a real `waiting_permission` state with the gated tool's name; tracking pending approvals as a set of toolCallIds keeps overlapping approvals correct.
- **Notification keys (verified live)**: the approval prompt is the shared `SelectList` with `["Approve", "Deny"]`, always opening with index 0 ("Approve") highlighted, no digit/letter shortcuts, so `approve: ["Enter"]` can't mis-approve; `deny: ["Escape"]` uses the constant `esc cancel` affordance and fails the tool call closed ("Tool call denied by user", command verified not to run).
- **Agent dir resolution quirk**: omp resolves its agent dir as `join(homedir(), PI_CONFIG_DIR || ".omp", "agent")` with the env var joined *verbatim* — an absolute `PI_CONFIG_DIR=/u/x/.config/omp` yields `~/u/x/.config/omp/agent`, and that mangled dir is what the binary actually scans (confirmed via `omp config path` and live extension discovery). The install path constant must replicate this join bug-for-bug or the extension lands in a dir omp never reads.
- **Terminal fallback**: default working message is `Working…` with a U+2026 ellipsis (not Pi's three ASCII dots); approval prompt body starts with `Allow tool: <name>` — the waiting rule must precede the working rule since the loader stays on screen.
- **Reply-on-finished**: same composer semantics as Pi (`/^\s*!/` unsafe pattern — omp trims before `!` bash-mode detection).
- **Invoke**: `omp -p "<prompt>"` print mode returns clean stdout (verified via `ccmux invoke`).

## Alternatives considered

A custom agent definition via ccmux config covers process matching + terminal rules, but can't install the extension, so there's no authoritative working/idle signal, no `waiting_permission` (approval waits are the main thing I'd want notifications for), no Approve/Deny actions, and no `invoke` support.

## Relevant agent(s)

Pi (adjacent upstream), new agent: oh-my-pi

## Additional context

I have a complete, tested implementation of this adapter locally (hook adapter + authored extension + agent def + tests, mirroring the Pi structure), verified end-to-end in tmux: process detection, extension install, marker lifecycle (`idle → working → idle`, prompt capture), `waiting_permission` with `pendingTool: bash`, live Approve (Enter) / Deny (Escape) keystrokes, and `invoke`. Happy to share it as a reference branch if that helps the design discussion — not opening a PR at this point per request.

epilande:
Adds built-in support for [oh-my-pi](https://github.com/can1357/oh-my-pi) (`omp`), an actively developed hard fork of Pi that kept Pi's extension API. Closes #60.

Credit to @yorisilo for the issue #60 research, which correctly identified nearly every quirk this PR handles (the def-ordering collision with pi, the `hasHandlers`-gated approval events, the `PI_CONFIG_DIR` verbatim join, the U+2026 working string, and the Approve/Deny select semantics). All of it was re-verified against the omp 17.1.3 source and a live install before implementation.

## What's included

Structurally this mirrors the existing Pi support (adapter + bundled extension + agent def), with one meaningful upgrade: omp HAS a tool-approval pause, so unlike Pi it gets a real permission wait with actionable notifications.

- **Agent def** ordered immediately before `pi` in `BUILTIN_AGENTS` (`findAgentForProcess` is first-match-wins and pi's `commandPatterns` regex matches omp's resolved launcher path; omp's own patterns require the `oh-my-pi` component so the reverse collision is impossible).
- **Process detection covers both launch shapes.** omp's npm bin shim runs under Bun, where `process.title` assignment is a no-op for `ps` (unlike pi, whose shebang is node), so the live process is `bun /.../bin/omp ...`. A `/[/\\]bin[/\\]omp(?:\s|$)/i` command pattern catches the shim shape; `processMatch /^omp$/i` covers standalone/node shapes.
- **Bundled extension** (`~/.omp/agent/extensions/ccmux.js`, installed via `ccmux setup --agent omp`, sentinel-gated like Pi's): marker at `session_start`, working/idle on `agent_start`/`agent_end`, prompt capture, and `waiting_permission` + `pending_tool` from `tool_approval_requested`/`tool_approval_resolved`. Overlapping approvals stay waiting until the last one resolves; aborted turns and uncorrelatable resolves are handled defensively. Subscribing to the approval events does not change behavior for default (`yolo`) users; the events are `hasHandlers`-gated upstream and only fire when an approval mode is configured.
- **Notification actions**: Approve on Enter (the select opens on index 0 = Approve, no digit shortcuts), Deny on Escape (fail-closed upstream: the tool call is denied and does not run), reply-on-finished guarded by `/^\s*[/!]/`: omp trims the submitted text before BOTH its `!` bash-mode check and its slash-command dispatch, so neither prefix is space-defusable (live-verified with the daemon's exact delivery sequence: a defused ` /new` still dispatched and destroyed the session, and a spaceless ` /<token>` left the fuzzy command selector open so the submitting Enter invoked an arbitrary match).
- **Install path is bug-compatible with omp's config-dir resolution**: omp joins `PI_CONFIG_DIR` verbatim under `$HOME` (live-verified; an absolute value gets mangled), so `OMP_AGENT_DIR` replicates the join rather than "fixing" it. Remaining relocations (`PI_CODING_AGENT_DIR`, profiles, XDG) are documented as a TODO with `omp config path` named as the eventual resolution.
- **All the easy-to-miss registration points**: `runLinkPasses` + `linkOmpSessions()` (the pass that does the real linking, since the marker beats the first scan), a version-resolver alias (`@oh-my-pi/pi-coding-agent` does not contain "omp"), `agentColorFor` (teal, deliberately shared with Pi to signal the fork kinship; rows stay tellable apart via the `om` short code and label), `createBuiltinHookAdapters()`, and the hardcoded test lists.
- **Terminal fallback** for no-hooks sessions: `allow tool:` (waiting, ordered first) and `working…` (U+2026). The working rule only covers the pre-intent window because omp swaps the loader label for model-supplied text mid-turn; with the extension installed the marker is authoritative, and the docs say so.
- **Invoke**: `ccmux invoke omp` shells `omp -p "<prompt>"` (stdout). Resume in print mode is unverified, so `resumeArgs` is not exposed.
- Tests (29 new across plugin/adapter/renderer plus agent-def matching incl. bun-shim and negative guards) and a full docs sweep (agent-adapters.md caveats section, architecture.md, invoke.md, README, AGENTS.md).

## Verification

- `bun run typecheck`, full `bun test` (3183 pass / 0 fail), knip, build, prettier all clean.
- Live e2e on omp 17.1.3 (macOS, tmux, real Gemini turns): extension install, plain-`omp` detection (the bun-shim shape), marker lifecycle with `nativeSessionId` binding and 17.1.3 in the version column, working/idle flips with prompt capture, approval wait surfacing `waiting`/`permission`/`pendingTool: bash`, Deny via Escape (gated `touch` verifiably did not run), Approve via Enter (it did), the no-hooks terminal-rule path (both rules, ordering not masked), `ccmux invoke omp` returning the exact reply, and picker rendering at 200x50 and 60x30 (agent accent, `om` shortcode, responsive collapse intact; the accent was verified as yellow and later switched to Pi's teal).

## Known gaps / follow-ups

- Desktop notification delivery was not live-fired for omp; the waiting tuple is identical to Antigravity's, so it rides the same unit-tested notifier path.
- No-hooks sessions under-report `working` for most of a turn (dynamic loader label, inherent; the extension marker is the fix and the default).
- `omp --resume <id>` exists (printed on session exit) and could later back `resumeArgs` for invoke once print-mode resume is verified.
- A setup-time `omp config path` check would resolve every config-dir relocation at once; deferred to keep setup fast and offline-safe.


epilande:
End-to-end run of omp support on this branch, against omp 17.1.3 with `tools.approvalMode = always-ask`. Playback is sped up.

The row moves `idle` to `working` to `waiting` to `idle`, then a `/new` session switch at the end:

- **waiting** is a real tool-approval pause, not a simulated one. The marker on disk carries `state: waiting_permission` with `pending_tool: bash`, and the picker renders the row as `waiting` with `bash` in the pending-tool column.
- Approving with **Enter** runs the gated command (the `touch` really executed) and clears the row back to `idle` with `pendingTool` null.
- The `/new` at the end exercises the `session_switch` handler: the old marker is reaped rather than leaked, and the row re-binds to the new session id.

Verifies `src/plugins/omp/ccmux.js` (oldest-wins approval publishing, `session_switch` handling), `src/daemon/adapters/omp/hook-adapter.ts` (marker projection now reusing the cascade's shared fold), and `src/daemon/cascade-evaluator.ts`.

https://github.com/user-attachments/assets/a72b1d08-66b8-4087-8605-f955c19c4cfa

Issue #69, opened 2026-07-29
[Feature]: Spawn a session into a new worktree


epilande:
## Overview

Adds a worktree destination to the spawn pipeline: create a worktree and launch an agent in it, from the CLI (`ccmux spawn --worktree [name]`) and from the picker's new session dialog.

Depends on #64 (spawn API groundwork), #65 (new session dialog), and #67 (worktree detection fixes).

## Motivation

Spawning today targets an existing directory (`--cwd`). Running parallel work on one repo means creating and wiring up a worktree by hand before ccmux can spawn into it. The picker already knows the repo context of every group and session, so "new agent on this repo in an isolated checkout, with this prompt" should be a single action.

This is a dispatch feature, not an isolation feature: agents that isolate their own session in a worktree already handle that themselves. The gap is starting additional parallel sessions from outside.

### Creation engine (daemon)

- Placement: `<main>/.claude/worktrees/<name>`, the same location agent-created worktrees use, so the whole worktree population lives under one convention and existing tooling sees one layout.
- Branch: created from the slugified name. `--base` selects the starting point; default is the main checkout's current branch.
- Naming: explicit `<name>` is both directory and branch. With a prompt and no name, derive a deterministic slug from the first words of the prompt (`"fix sidebar flicker on resize"` -> `fix-sidebar-flicker`). No name and no prompt is an error. No model calls, ever; naming is purely mechanical.
- File setup honors the conventions Claude Code already applies to its own worktrees: `worktree.symlinkDirectories` from `.claude/settings.json` and `.worktreeinclude`. ccmux-created worktrees behave identically to agent-created ones with zero new config.
- Create-or-open: if a worktree with the requested name already exists, spawn into it instead of failing.
- Concurrency: serialize worktree/git-config mutation per repo so parallel spawns are safe.
- Orphan handling: a leftover directory at the target path with no `.git` inside is removed; one with a `.git` is refused with a clear error.

### API and CLI

- `POST /spawn` accepts `worktree: { name?, base? }` alongside the existing fields. The handler creates the worktree, then continues through the normal spawn path with the new path as `cwd`, so placement, split direction, and per-agent prompt handling from #64 apply unchanged.
- `ccmux spawn <agent> --worktree [name] [--base <ref>]` maps onto the same request.

### Picker

- The new session dialog (#65) gains a destination field: current checkout (default) or new worktree, with a name input prefilled from the prompt slug. The dialog is already structured for this field to slot in.

## Out of scope

[...]

Issue #68, opened 2026-07-29
[Feature]: Prune merged and stale worktrees


epilande:
## Overview

Adds worktree cleanup: a picker surface and a `ccmux worktree prune` CLI that find finished worktrees, show why each one is removable (PR merged, PR closed, upstream gone, merged locally), and clean up the worktree directory, local branch, leftover pane, and per-directory agent state in one action.

Depends on #67 (worktree detection fixes).

## Motivation

The worktree lifecycle ends by hand today. After a branch is squash-merged on GitHub (with branch auto-delete), three leftovers remain on the machine: the worktree directory, the local branch, and often a tmux pane where the finished agent session sits idle. Agents create worktrees faster than they get cleaned up, and worktrees where work actually happened are exactly the ones that stick around.

Per-directory agent state accumulates too: as a data point from one machine, 127 of 199 project entries in `~/.claude.json` point at paths that no longer exist, most of them old worktrees.

ccmux is in the right position to close this loop. The daemon already caches PR state per branch and knows whether an agent is still alive in each worktree, which are the two facts that make cleanup safe.

### Candidate detection (daemon)

- For each repo that has worktrees (`git worktree list --porcelain` from the main checkout), classify every worktree with a reason: `PR merged`, `PR closed`, `upstream gone`, or `merged locally`. The `branchPRs` cache (`src/daemon/pr-resolver.ts`) already holds PR state; `upstream gone` needs a `git fetch --prune` per repo, run once when a prune surface opens.
- Dirty worktrees (uncommitted or untracked changes) are flagged prominently and never pre-selected.
- Session gating: a worktree whose session is `working` is excluded from the list entirely. Idle and finished sessions are listed; removing their worktree kills the agent process and closes the pane.
- The main checkout is never a candidate.

### Removal

- Rename the worktree directory to a trash sibling first, then delete. This frees the path immediately even when a shell still has it as cwd, and gives an undo window while the run is in progress.
- Delete the local branch only for `merged`/`upstream gone` reasons, with force only where the merge is proven.
- Close the pane after confirming the agent process exited.
- Run `git worktree prune` afterward, removing stale `locked` markers left by interrupted `git worktree add` runs so metadata is actually reclaimed.
- State cleanup: remove the deleted path's entry from `~/.claude.json`, writing a backup first.

### Surfaces

- Picker: a keybinding and a group-header context menu item open the prune list, showing reason, dirty flag, and session state per row, with multi-select and a confirmation step.
[...]

PR #72, opened 2026-07-29
fix(daemon): Worktree detection and display fixes

### LINKED ISSUE #67: [Feature]: Worktree detection and display fixes
## Overview

Fixes how worktrees are detected and closes the display gaps around worktree sessions: proper git-dir comparison, invoke-row grouping, the missing worktree marker on the branch column, repo identity on worktree rows, and a worktree list in the group preview. Groundwork for worktree cleanup and worktree spawning (follow-up issues).

## Motivation

- `isWorktree` is a substring test: `readGitInfo` checks `--git-dir` output for the literal `/worktrees/` (`src/daemon/server.ts:464`). It false-positives on any repo whose path contains that segment, and it cannot resolve which main checkout a worktree belongs to, which the cleanup and spawn features need.
- `project` and `gitBranch`/`isWorktree` are keyed off different cwds. Project derivation uses the log-derived `session.cwd` (`src/daemon/sessions.ts:533`), git info uses `paneCwd ?? cwd` (`src/daemon/server.ts:473`). A pane that `cd`s into a worktree updates one and not the other.
- Fabricated invoke rows use `cwd.split("/").pop()` and hardcode `isWorktree: false` (`src/tui/store.ts:183,212`). An invoke launched from a worktree groups under the worktree's directory name forever, while real sessions in the same worktree group under the repo name. The comment at `store.ts:173` documents the bug.
- The standalone `branch` column omits the `+` worktree marker (`src/tui/components/SessionItem.tsx:571`) that the default project cell includes (`src/tui/components/session-columns.ts:575`). Moving branch to its own column silently loses the indicator.
- Worktree rows never show repo identity. `formatProjectPath` keeps two path segments (`SessionItem.tsx:105`), so a worktree renders as `worktrees/<name>` at best, and the prefix is the first thing dropped under width pressure (`session-columns.ts:613`). Two same-named worktrees of different repos render identically.

### Detection

- `readGitInfo` adds `--git-common-dir` to the existing single `rev-parse` spawn (`src/daemon/server.ts:459`). A cwd is a worktree iff the resolved git-dir differs from the common dir. Derive the main checkout root from the common dir and expose it on `EnrichedSession` (e.g. `mainRepoRoot`), since the follow-up features key off it.

### Consistent cwd keying

- Derive `project` and git info from the same effective cwd so they cannot disagree. The pane-preferred cwd is the better source of truth for both.

### Invoke rows

- Fabricated invoke rows go through the same project derivation as real sessions (`src/daemon/project-derivation.ts`), either by importing `deriveProject` in the store or by having the daemon enrich invocations before broadcast. `isWorktree` follows.

### Display

- **SessionItem.tsx** - `branch` column gains the `+` marker, matching the project cell.
- **session-columns.ts** - worktree rows render repo identity as `<repo>/<worktree-name>` (e.g. `ccmux/parking`) instead of `worktrees/<name>`, and the repo segment survives width pressure ahead of the worktree name.
- **GroupPreview.tsx** - group preview lists the repo's worktree sessions (name, branch, status). Full on-disk worktree enumeration is not needed here; that arrives with the cleanup feature.

## Out of scope

- Worktree cleanup and pruning (follow-up issue).
- Worktree-destination spawning (follow-up issue).
- `projectCache` invalidation on worktree removal (`src/daemon/project-derivation.ts:50`); lands with cleanup, which is what deletes worktrees.

## Verification

- Unit tests: git-dir vs common-dir detection (main checkout, worktree, submodule `.git` file), invoke-row grouping through `deriveProject`, branch label with the marker.
- Picker and sidebar capture per the AGENTS.md workflow with at least one session running in a worktree: column alignment, `+` marker on both cells, repo identity under width pressure, group preview list.

epilande:
Closes #67

## Summary

Fixes worktree detection and closes the display gaps around worktree sessions, as groundwork for worktree pruning (#68) and worktree spawning (#69).

- **Detection**: `readGitInfo` now compares `--git-dir` against `--git-common-dir` (added to the existing single `rev-parse` spawn) instead of substring-matching `/worktrees/` in the path. Both dirs are resolved against the cwd first, which matters because git prints a cwd-relative `--git-common-dir` from subdirectories of a plain checkout. The main checkout root is exposed on `EnrichedSession` as `mainRepoRoot` for the follow-up features.
- **Consistent cwd keying**: `project` and git info are now derived from the same pane-preferred effective cwd at enrich time, so a pane that `cd`s into a worktree can no longer split the two.
- **Invoke rows**: fabricated invoke rows are enriched daemon-side (`deriveProjectInfo`) before broadcast, so they group under the repo like real sessions instead of under the worktree directory name. Fields are optional on the event with a basename fallback so an older daemon can still serve a newer board. Enrichment stays synchronous to preserve strict SSE ordering.
- **Display**: the standalone `branch` column gains the `+` worktree marker to match the project cell; worktree rows render `<repo>/<worktree>` (e.g. `ccmux/parking`) instead of `worktrees/<name>`; the group preview lists the repo's worktree sessions (name, branch, status).

## Deviation from the issue

The issue asked for the repo segment to survive width pressure ahead of the worktree name. Implemented literally, that produced a meaningless `c…/` stub while starving the worktree name. Final rule: the worktree name floors at 3 chars while sharing the cell, the repo segment drops out whole below 5 chars instead of degrading to a stub, and the worktree name reclaims the freed space. Repo identity survives to roughly 100 columns and degrades cleanly below.

## Verification

- `bun run typecheck` clean; `bun test` 3208 pass / 0 fail (new tests: git-dir vs common-dir detection incl. submodule `.git` file, relative common-dir, and a repo living under a directory literally named `worktrees`, plus a real-git worktree fixture; invoke-row grouping through `deriveProjectInfo`; branch-column marker; `repo/worktree` labeling and shrink order; group preview list).
- Live against an isolated daemon (separate port, `CCMUX_HOME`, tmux socket) with sessions in two real worktrees plus the main checkout: `GET /sessions` returned correct `isWorktree`/`mainRepoRoot` for all three; picker captured at 120/100/70/50 columns and sidebar at 40 columns showing the `+` marker and `repo/worktree` labels; group preview showed the Worktrees block only for the repo with worktree sessions.

E2e recording to follow.


epilande:
## Verification

Ran the tier sweep against a daemon built from this branch, on an isolated stack (its own `CCMUX_HOME`, port and tmux server) so the live one was never touched.

**Static**: `typecheck`, `knip`, `build` and `bun test` green (3208 pass / 0 fail).

**Detection, live against real worktrees**: three Claude sessions in this repo - the main checkout, `.claude/worktrees/parking` and `.claude/worktrees/hunk-review`. The daemon reports `isWorktree: false` for the main checkout and `true` for both worktrees, `mainRepoRoot` resolves to the same main checkout for all three, and all three group under `ccmux`.

**The false positive is gone**: a plain repo living under a directory named `worktrees/` is no longer called a worktree. Checked against real git from both the repo root (`--git-dir` and `--git-common-dir` both print `.git`) and from a subdirectory (absolute `--git-dir`, cwd-relative `../.git` common dir) - the shape only the resolve-then-compare gets right.

**Display**: worktree rows render `ccmux/parking:feat/sidebar-parking+`, the main checkout renders the plain path `epilande/ccmux:main`. Moving `branch` to its own column keeps the marker (`feat/sidebar-parking+` vs a bare `main`). The group preview grows a `Worktrees` block listing each worktree session with its branch, and drops the block when the group has none.

**Under width pressure**: between 130 and 170 columns the repo segment survives while the worktree name truncates (`ccmux/hunk-rev…:hunk-review+`). Narrower than that the worktree name reaches its 5-char floor and the repo prefix goes, which is the documented last-resort order.

**Invoke**: a live `ccmux invoke codex --cwd .../worktrees/focusdance` produced a row labeled `ccmux/focusdance`, grouped under `ccmux` alongside the real sessions and listed in the group preview's `Worktrees` block, instead of stranding itself under `focusdance` for its lifetime.

One path not exercised live: the re-derivation of `project` from the pane cwd when a pane `cd`s out of the directory the log recorded. That one is covered by the new unit test only.

A screen recording of the picker through all of the above will be attached separately.

epilande:
> [!NOTE]
> **Superseded** by the [updated recording](https://github.com/epilande/ccmux/pull/72#issuecomment-5116929927). This clip predates the review fixes: it shows the group preview before the worktree dedupe, and before worktree rows were labeled by worktree root instead of the pane cwd.

### E2E recording

Picker against a daemon built from this branch, with real Claude sessions in the main checkout and in two worktrees. Shows the `<repo>/<worktree>` labels and `+` markers, the group preview’s Worktrees block, a worktree agent going live-working, and a `ccmux invoke` row from a worktree appearing as `ccmux/focusdance` grouped under `ccmux` rather than stranded under its own directory name.

Note: the per-row preview reads “Pane is on a different tmux server” because the picker ran in its own pty outside the isolated tmux server used for verification. That is an isolation artifact, not a behavior of this branch.


https://github.com/user-attachments/assets/4b11d779-1e8c-4b3e-af9a-b029646b867a

epilande:
### E2E recording (updated for the review fixes)

Supersedes the earlier recording, which predates the fixes and shows the stale group preview.

Filmed against 2898b1f on an isolated stack. A sandbox repo (`claude-toolkit`) with one worktree hosting two Claude agents: one at the worktree root, one deliberately in `worktree-detection/src/tui`.

1. Both picker rows render `claude-toolkit/worktree-detection:fix/worktree-detection+` — identical labels despite one pane sitting in a subdirectory. That row previously read `claude-toolkit/tui`, inventing a worktree that does not exist.
2. The group preview Worktrees block shows one line, `worktree-detection fix/worktree-detection ×2`, where it previously showed two identical lines. The session lines below carry their real cwds with no `(worktree)` suffix.
3. The `src/tui` agent runs a turn: its row flips to working, the group summary follows, and the Worktrees line adopts the most attention-worthy status of the sessions it represents.
4. Back to idle, labels unchanged throughout.


https://github.com/user-attachments/assets/2c61e9bd-a734-496b-b318-85dcbe150196

Issue #66, opened 2026-07-29
[Feature]: Picker: fork a session into a parallel pane


epilande:
## Overview

Adds a "Fork" action to the picker's session context menu: spawn a new pane with the same agent and cwd whose conversation continues from the source session's history, leaving the original untouched.

Depends on #64 (target pane + split direction). Claude Code only in v1.

## Motivation

The strongest version of "duplicate this session" is branching the conversation, not starting fresh: the agent is heading down path A, fork it and try path B in a parallel pane. Claude Code supports this natively (`claude --resume <session-id> --fork-session` creates a new session id and does not touch the original), and ccmux already tracks `nativeSessionId` for hook-installed sessions, so the pieces exist.

## Design

- New `AgentDef` field `forkCommand?: string`, a command template with `{id}` (same pattern as `resumeCommand`, `src/lib/agents.ts:71`). Claude Code: `claude --resume {id} --fork-session`.
- Context menu on session rows (`sessionMenuItems()`, `src/tui/App.tsx:437`) gains **Fork**, shown only when the agent defines `forkCommand` AND the session has a `nativeSessionId`. Pane-tracked-only rows (no hooks, no session id) do not show it.
- Placement: split next to the source session's pane by default (uses the spawn API's `target` + direction). The forked session lands visually beside the original.
- Daemon side: either a `fork: true` + `sessionId` variant on `POST /spawn`, or a dedicated endpoint; the handler resolves the session, builds the command from `forkCommand`, and reuses the spawn pane-creation path.
- CLI parity: `ccmux spawn <agent> --fork <session-id>` falls out of the same handler and is worth exposing.
- Keyboard: menu-only is acceptable for v1. If a key is added, note that plain `f` is taken (hide-idle toggle, `App.tsx:1166`).

## Per-agent notes for the implementer

- Claude Code is the only verified fork: `--fork-session` writes a new transcript and session id. Fork while the source is mid-turn should be tested; worst case, gate Fork on the session not being in `working` state.
- Codex and others: resuming a session that is still live may contend over rollout/session state. Do not enable fork for an agent without verifying its resume semantics against a live original. Adding an agent later is just defining `forkCommand` once verified.
- The forked pane gets tracked like any spawn: hooks fire on the new process and the marker pipeline binds it normally. No special reconciler work expected.

## Relation to planned worktree work

[...]

Issue #65, opened 2026-07-29
[Feature]: Picker: new session dialog with context-aware defaults


epilande:
## Overview

Adds a "new session" dialog to the picker, opened with <kbd>n</kbd> or from the existing right-click context menus. Every field is optional with context-aware defaults, so the fast path is a couple of keystrokes: `n`, Enter.

Depends on #64 (split direction, target pane, per-agent prompt handling).

## Motivation

The picker can attach, kill, restart, and review sessions, but creating one still means dropping to a shell for `ccmux spawn` or launching the agent manually. The picker already knows the context that makes spawning fast (selected session's agent and cwd, group's repo), so a dialog prefilled from that context turns "another Claude on this repo" into two keystrokes.

### Dialog fields (all optional, Enter accepts defaults)

1. **Agent**: list of spawnable agents (built-ins detected on PATH plus custom agents from config), j/k navigation, number keys for instant pick. Default: the selected row's agent, else the last agent spawned from the dialog.
2. **Placement**: new window / horizontal split / vertical split. Default: new window.
3. **Prompt**: free-text input, empty by default. Passed through the per-agent prompt handling from #64.

cwd is derived from context rather than asked for: selected session row -> that session's cwd; selected group header (grouped by repo) -> the group's cwd; no selection -> the picker's own cwd. Show the derived cwd in the dialog so the user can see what they are getting.

### Entry points

- <kbd>n</kbd> from the session list. Plain `n` is unbound today; only <kbd>Ctrl</kbd>+<kbd>N</kbd> is taken (moves selection, `src/tui/App.tsx:1076`).
- Session row context menu: new item "New session here" (prefills agent + cwd from the row).
- Group header context menu: new item "New session in repo" (prefills cwd from the group).
- No empty-space right-click; entry points are the key and the existing menus only.

### Post-spawn behavior

- Popup picker: dismiss and jump to the new pane (spawn-then-jump matches the picker's purpose).
- Sidebar: spawn without stealing focus; the target pane should be in the window's main area, not the sidebar strip.
- Placement must target the pane the picker was launched over, not whatever tmux considers current. The spawn API's `target` parameter carries this; the picker resolves the client's active pane at launch.

## Implementation pointers

- Keymap switch: `src/tui/App.tsx` (~line 1035).
[...]

PR #73, opened 2026-07-29
feat(spawn): Split direction, target pane, and per-agent prompt handling

### LINKED ISSUE #64: [Feature]: Spawn API: split direction, target pane, per-agent prompt handling
## Overview

Extends the spawn pipeline (daemon `POST /spawn` + `ccmux spawn` CLI) with split direction, an explicit target pane, and per-agent initial-prompt handling. This is groundwork for spawning sessions from the picker (see the follow-up issues), but each improvement stands on its own for CLI users.

## Motivation

- `ccmux spawn --split` always uses tmux's default stacked split. There is no way to request a side-by-side split from the CLI or the API.
- The daemon runs `split-window`/`new-window` with no `-t` target (`src/daemon/server.ts:1780`), so tmux resolves "current" from the daemon's server context, which is arbitrary when the daemon is not attached to the client that made the request. Callers need to say where the pane goes.
- The initial prompt is passed as `${cmd} --prompt '<text>'` for every agent (`src/daemon/server.ts:1771`). This is wrong for pi, which takes the prompt as a positional argument and has no `--prompt` flag (verified against `pi --help`), and for Copilot, whose `--prompt` means one-shot print mode rather than an interactive session with an initial prompt. Other agents (Claude, Codex) also take positional prompts and should be verified individually.

## Current behavior

- Route: `src/daemon/server.ts:732` -> `handleSpawn` at `src/daemon/server.ts:1679`. Builds the agent command (resume via `resumeCommand` template, prompt via hardcoded `--prompt`, else bare executable), then runs `tmux split-window|new-window -c <cwd> -P -F '#{pane_id}'` and `send-keys` the command into the new pane.
- CLI: `src/commands/spawn.ts` exposes `--cwd`, `--resume`, `--prompt`, `--split` (boolean), `--detach`.
- Agent definitions: `src/lib/agents.ts` (`AgentDef` at line 61). `resumeCommand` (line 71) is the existing per-agent command template precedent, using `{id}` as the placeholder. Custom agents merge overrides in `mergeAgentConfig` (~line 314) with types in `src/lib/preferences.ts` (`AgentConfig`).

### API (`POST /spawn`)

- `split`: accept `"h" | "v"` in addition to the existing boolean (boolean stays supported, mapping to the current default direction).
- `target`: optional tmux pane id. When present, pass `-t <target>` to `split-window`/`new-window` so placement is deterministic. When absent, keep current behavior.

### CLI (`ccmux spawn`)

- `--split [direction]`: optional value `h` or `v`; bare `--split` keeps today's behavior.
- Pass the caller's own pane (`$TMUX_PANE`) as `target` when run inside tmux, so "split the pane I ran this from" works regardless of which window is active elsewhere.

### Per-agent prompt handling

- Add `promptCommand?: string` to `AgentDef`, a template with a `{prompt}` placeholder (same pattern as `resumeCommand`'s `{id}`). Example: pi would use `pi '{prompt}'` semantics (positional), agents with a real flag declare it explicitly.
- `handleSpawn` uses `promptCommand` when defined; agents without it either fall back to positional (if verified safe) or reject prompt spawns with a clear error rather than emitting a broken flag.
- Escaping stays centralized (the existing single-quote escaping at `server.ts:1770`); the template substitutes the already-escaped value.
- Make `promptCommand` overridable in custom agent config (`AgentConfig` in `src/lib/preferences.ts`), same as `resumeCommand`.
- Verify the correct invocation per built-in agent as part of this work (interactive session with prompt submitted, not print/one-shot mode). Record findings per agent in the PR.

## Out of scope

- Picker UI for spawning (follow-up issue).
- Fork/resume UX (follow-up issue).
- Worktree-destination spawning (planned separately as part of the worktree roadmap).

## Verification

- Unit tests for the spawn command builder (direction mapping, target passing, prompt templates per agent).
- Live check per changed agent in a detached tmux session (see the AGENTS.md verification workflow): spawn with a prompt, confirm the agent starts interactively and the prompt is submitted, confirm split direction and target pane land where requested.

epilande:
Resolves #64

## Overview

Extends the spawn pipeline (daemon `POST /spawn` + `ccmux spawn`) with split direction, an explicit target pane, and per-agent initial-prompt handling. Groundwork for picker spawning (#65), fork (#66), and worktree spawning (#69), but each improvement stands alone for CLI users.

- **Split direction**: `split` accepts `"h" | "v"` alongside the existing boolean (boolean keeps the current stacked default). CLI: `--split [direction]`.
- **Placement**: `target` (an explicit pane) and `callerPane` (where the request came from) are separate fields, because they mean different things. The CLI fills `callerPane` from `$TMUX_PANE` so a spawn lands in your session instead of wherever the unattached daemon resolves "current"; `--target <pane-id>` requests a specific placement and `--target none` opts out.
- **Per-agent prompts**: new `AgentDef.promptCommand` template (`{bin}`/`{prompt}` placeholders, same pattern as `resumeCommand`), overridable in custom agent config. All nine built-ins declare their verified interactive invocation; an agent without one gets a clear 400 naming the config key instead of a broken flag.
- Command/argv construction extracted to a pure `src/daemon/spawn-command.ts` so direction mapping, placement shapes, templates, and escaping are unit-testable without live tmux.

## Why per-agent templates: the old `--prompt` was wrong for 8 of 9 agents

Verified against each CLI's `--help` on a live machine. The old `<binary> --prompt '<text>'` was correct only for OpenCode. For Antigravity, Copilot, and Gemini it silently selected one-shot print mode; for the other five, `--prompt` is not a flag at all.

| Agent       | Interactive + initial prompt           | Basis                                                                                                     |
| ----------- | -------------------------------------- | --------------------------------------------------------------------------------------------------------- |
| Claude Code | `claude '<prompt>'` (positional)       | `Arguments: prompt`; interactive by default, `-p/--print` is non-interactive                              |
| Codex       | `codex '<prompt>'` (positional)        | `codex [OPTIONS] [PROMPT]`; `codex exec` is the one-shot subcommand                                       |
| Cursor      | `cursor-agent '<prompt>'` (positional) | `prompt  Initial prompt for the agent`; `-p/--print` non-interactive                                      |
| OpenCode    | `opencode --prompt '<prompt>'`         | default TUI command's `--prompt  prompt to use`; positional is a project path                             |
| Pi          | `pi '<prompt>'` (positional)           | own help examples: interactive vs `-p` non-interactive                                                    |
| oh-my-pi    | `omp '<prompt>'` (positional)          | same examples as pi                                                                                       |
| Antigravity | `agy -i '<prompt>'`                    | `--prompt` is an alias for `--print`; `-i/--prompt-interactive` runs the prompt and continues the session |
| Copilot     | `copilot -i '<prompt>'`                | `-i/--interactive <prompt>`; `-p/--prompt` exits after completion                                         |
| Gemini      | `gemini -i '<prompt>'`                 | `-p/--prompt` is headless; `-i/--prompt-interactive` continues interactively                              |

Live-verified by spawning: Claude, OpenCode, Gemini, Pi, oh-my-pi. Codex, Cursor, Copilot, and Antigravity rest on the help evidence above plus a table test.

## Prompt handling is hardened against injection

Prompt text is attacker-shaped input that ends up in a shell command typed into a pane. Two defects found in review, both with working exploits, are closed:

- **`$` replacement patterns.** `String.replace` with a string replacement expands `` $` ``, `$&`, `$'`, and `$$` in the _replacement_, so a prompt of ``$`; touch ./PWNED; #`` spliced template text back in, reopened the quoted word, and executed. Substitution is now a single pass with a function replacer. Single pass matters independently: sequential passes let a binary containing the literal text `{prompt}` relocate the prompt outside the quotes the guard just verified.
- **Quote-context bypass.** The guard inspected only the characters adjacent to the placeholder, so `sh -c "{bin} '{prompt}'"` passed while the enclosing double quotes made the escaping inert. Templates are now parsed the way `sh` reads them: every `{prompt}` must land in a genuine single-quoted context with quotes balanced, and double quotes, backticks, and `$(` are refused outright.

Also: a NUL in a prompt survived escaping but made `Bun.spawn` reject the argv _after_ the pane existed, orphaning it behind an opaque 500. Prompts are validated before anything is created, and every post-creation failure now kills the pane it created.

## Implementation notes

- **`new-window` cannot take a pane id** (tmux errors "can't specify pane here", and `-t @id` errors "index in use"). An explicit `--target` resolves pane to window and uses `-a -t @id`; the implicit caller pane resolves to its **session** and appends with `-t $id:`, which renumbers nothing. `-a -t @id` was verified to shift every later window's index, so using it for a plain `ccmux spawn` would have silently reindexed the caller's windows.
- **Single-server invariant.** Pane ids collide across tmux servers, so the CLI compares its socket against `/server-info` and drops the implicit caller pane on a mismatch, the same guard `ccmux switch` already applies.
- **`{bin}` placeholder** keeps launcher overrides (e.g. Claude's `command` preference) working on the prompt path. The launcher must be quote-neutral, which is what allows the quoting scan to treat `{bin}` as inert.
- **Uniform errors.** A dead target pane returns the same 400 on both the split and new-window paths; the split path used to surface tmux's raw stderr as a 500.

## Breaking Changes

- Because `--split` takes an optional value, Commander consumes a following bare token: `ccmux spawn --split codex` now errors where it used to parse. The error names the fix (`ccmux spawn codex --split`), and `--split=h` also works.

## Behavior changes to note

- `ccmux spawn` run inside tmux now creates its window at the end of **your** session rather than wherever the unattached daemon considered "current". Existing window indices are left alone.
- `ccmux spawn` without `--cwd` now starts the agent in the directory you ran it from. It previously used the ccmux install directory, since `bin/ccmux` cds to the package root; `review` and `sidebar` already restored `CCMUX_CALLER_PWD` and spawn did not. A relative `--cwd` resolves against your directory too.
- `--prompt ""` is now a 400 instead of silently spawning a bare agent.
- Targeting a pane in a different tmux session creates the pane but does not move you to it, so pair that with `--detach`.

## Testing

- [x] `bun test` 3270 pass / 0 fail across 149 files on the final branch state. New coverage: a nine-row built-in prompt-template table plus a guard that fails if a new built-in lacks `promptCommand`, the injection payloads round-tripped through a real `/bin/sh`, every quote-context bypass, both directions of `--detach`, a new `spawn.test.ts` for the CLI (caller pwd, caller-pane vs explicit target, cross-server drop, `--target none`), a new `pane-io.test.ts` for `resolvePaneLocation`, and route-level coverage in `server.test.ts`
- [x] Injection payloads (`` $` ``, `$&`, `$'`, `$$`) round-trip through a real `/bin/sh` and arrive as literal text; the canary file is asserted absent
- [x] Live matrix against an isolated daemon (scratchpad `CCMUX_HOME`, private port): h/v/legacy splits, explicit target, implicit caller pane, every 400 path, hostile prompts, and the caller-directory fix
- [x] Window indices confirmed unchanged by an implicit spawn and shifted by an explicit `--target`, matching intent
- [x] E2e verification passed; recording posted below

Known gaps: the "new window, no target" default is unit-test-only, since running it live risks landing a window in the attached session.


epilande:
## Verification

Verified end to end against an isolated daemon running this branch (separate config home and port, detached tmux sessions only). All tiers pass.

### Static

- `bun run typecheck`, `bun run check:unused`, `bun run build` all clean
- `bun test`: 3231 pass, 0 fail (33 of those are the new `spawn-command.test.ts`)

### Placement, live against a real tmux server

Every case ran through `POST /spawn` and was confirmed by reading back `pane_left` / `pane_top` / `window_index`, not just the response body.

| Case                             | Result                                                                                                          |
| :------------------------------- | :-------------------------------------------------------------------------------------------------------------- |
| `split: "h"`                     | New pane beside the target (`left=0` and `left=101`, same row)                                                  |
| `split: "v"`                     | New pane stacked below the target                                                                               |
| `split: true`                    | Stacked, identical to `"v"`, so existing callers are unchanged                                                  |
| `target` on a split              | Splits the named pane, not the caller's                                                                         |
| `target` on a new window         | Pane id resolved to its window, new window inserted directly after it (existing windows shifted 2 to 3, 3 to 4) |
| `ccmux spawn` with no `--target` | Picks up `$TMUX_PANE` and lands next to the caller                                                              |

Rejections behave as designed: `split: "x"` and `split: "H"` return 400, a window id or session name as `target` returns 400, a dead pane for a new window returns 400 `Unknown target pane`, and `--split diagonal` is rejected at the CLI before the request goes out.

### Prompt handling

`--prompt` for an agent with no `promptCommand` returns 400 naming the exact config key, and a custom agent that declares one is honored (`{bin}` resolves through an `executable` override).

Escaping holds under a hostile prompt. Spawning with `it's $(whoami) && rm -rf; danger` types a single-quoted word into the pane and the agent receives the text literally, with no command substitution.

Three agents were spawned for real and each started interactively with the prompt already submitted, then answered:

- `claude 'Reply with exactly: OK'`
- `pi 'Reply with exactly: OK'`
- `omp 'Reply with exactly: OK'`

That covers the positional shape for Claude plus two of the agents that were previously help-verified only. The remaining flag-shaped agents (`agy`, `copilot`, `gemini`) and `codex` / `cursor` are still covered by the table test rather than a live spawn.

### One thing worth a follow-up (pre-existing, not from this PR)

`ccmux spawn` without `--cwd` starts the agent in the ccmux install directory rather than the directory it was run from. `bin/ccmux` deliberately `cd`s to the repo root and exports `CCMUX_CALLER_PWD`, and `spawn.ts` uses `process.cwd()`. `review.ts` and `sidebar.ts` already read `CCMUX_CALLER_PWD` instead.

Reproduced from a pane in `/private/tmp/ccmux73-demo`: the spawned panes all came up in the ccmux checkout. That line is untouched by this PR, but it sits right next to the "lands next to the caller" behavior this PR adds, so it is worth resolving the same way the other two commands do.

epilande:
### E2E recording

Real tmux session showing, in order: horizontal split, vertical split, `--target` splitting a pane other than the caller’s, and `ccmux spawn claude --prompt` starting Claude interactively with the prompt auto-submitted.


https://github.com/user-attachments/assets/363f9413-c664-4860-a60f-34ee80da3f94

epilande:
### E2E recording, part 2: window placement and caller directory

Supplements the [first recording](https://github.com/epilande/ccmux/pull/73#issuecomment-5116662592), which covers the splits, `--target`, and prompt escaping. This one covers the two user-visible behavior changes that clip did not show.

One continuous take:

1. `tmux list-windows` — before state: `1 shell, 2 api, 3 web, 4 docs`.
2. `ccmux spawn codex` with no placement flags.
3. `tmux list-windows` — windows 1-4 byte-identical, 5 appended. Nothing reindexed, where the pre-fix code inserted adjacently and shifted every later window.
4. `tmux list-panes -s -F "#{window_index} #{pane_current_path}"` — the spawned window sits in the callers directory, not the ccmux checkout.

The closing frame holds all four steps on screen at once.

Filmed with codex rather than claude, and the cwd is proven via tmux's own `pane_current_path` rather than an agent banner — the pane's actual cwd is stronger evidence than what an agent chooses to print. Both filmed behaviors are agent-independent: the agent name is a positional argument on the same no-placement-flag path.


https://github.com/user-attachments/assets/eb31db3f-59ec-4f7b-8ab0-c8aa019a5855

PR #76, opened 2026-07-29
feat(worktree): Prune merged and stale worktrees

### LINKED ISSUE #68: [Feature]: Prune merged and stale worktrees

epilande:
Closes #68

## Summary

Adds worktree cleanup: a picker surface and a `ccmux worktree prune` CLI that find finished worktrees, show why each one is removable, and clean up the worktree directory, local branch, leftover pane, and per-directory agent state in one action.

Builds directly on #67's `mainRepoRoot`, which is what makes "which main checkout does this worktree belong to" answerable at all.

- **Classification** (`src/daemon/worktree-prune.ts`, `worktree-git.ts`): `git worktree list --porcelain` per repo, each worktree classified `pr-merged` > `merged-locally` > `upstream-gone` > `pr-closed`. Dirty worktrees are flagged and never pre-selected. A worktree whose session is `working` is excluded entirely; the main checkout is never a candidate.
- **Removal**: stop agent → close pane → rename to a trash sibling → clear stale locks + `git worktree prune` → delete branches → state cleanup → delete trash.
- **Surfaces**: `W` in the picker (plus a "Prune Worktrees" group-header context menu item scoped to that group's repo), and `ccmux worktree prune` with `--dry-run` and `--state`.
- **State cleanup** (`src/daemon/agent-state.ts`): removes the deleted path's `~/.claude.json` `projects` entry, writing a backup first. `--state` additionally clears the accumulated backlog of entries pointing at paths deleted outside ccmux.

## Safety properties

These are the load-bearing bits, called out so they survive future refactors:

- **Removal re-scans and refuses stale input.** `POST /worktrees/prune` takes paths, re-scans in-process, and 409s any path the fresh scan no longer classifies as removable. A stale client list, a replay, or a hand-written POST cannot delete a directory that has since become active.
- **Branch deletion follows proof, not category.** `-D` only for `pr-merged` (a squash merge leaves the tip unmerged by git's definition). `merged-locally` and `upstream-gone` use the safe `-d`; `pr-closed` keeps the branch. **Deviation from the issue:** it asks for branch deletion on `upstream gone`, and this does delete — but with `-d`, because a deleted remote branch is a strong hint, not proof. An unmerged branch survives with a reported failure instead of being force-deleted on a guess.
- **The dirty gate lives in the destructive core** (`allowDirtyPaths` in `runPrune`), not in the surfaces, so both the picker and the CLI inherit it. In the picker, `a` (select all) covers only clean rows, so no bulk key can be the thing that deletes uncommitted work.
- **Phase order is load-bearing.** Branch deletion must come *after* `git worktree prune`: until the admin entry is gone, git considers the branch checked out in a worktree and refuses to delete it with or without `-D`.
- **No `--yes` and no automatic mode.** `--dry-run` covers scripted use; a non-TTY run refuses. Every removal is logged with what was deleted and where the trash went.

## Verification

- `bun run typecheck` clean; `bun test` 3269 pass / 0 fail across 149 files.
- Live end-to-end against throwaway sandbox repos under `/private/tmp` with an isolated daemon (own port, `CCMUX_HOME`, and tmux server): classification for each reason including a real `fetch --prune`, dirty flagging, the working-session gate moving a worktree in and out of candidates, a CLI interactive run, a picker run confirming the dirty gate holds until explicitly opted in, `--state` against a fixture HOME (nested entries and orphan backlog removed, live entries preserved, backup written), and the sidebar at 44 columns.
- Two bugs the live runs caught and fixed: a pane that closed together with its agent was reported as a failure, and git's `hint:` advice was being pasted into step details.

E2e recording to follow.


epilande:
## Verification: PASS

Ran the full diff-driven verification. Everything was driven against a throwaway sandbox repo under `/private/tmp/v76`, with an isolated daemon (its own `CCMUX_HOME`, port 2307, its own tmux server, and `HOME` pointed at a scratch dir). No real repo or worktree was ever in scope, and the daemon's session inventory only ever contained the sandbox's own sessions.

| Tier               | Verdict                                                        |
| ------------------ | -------------------------------------------------------------- |
| 0 static           | PASS, typecheck + knip + build clean, 3269 tests pass / 0 fail |
| 2 TUI render       | PASS, dialog renders at 200 and 46 columns                     |
| 4 CLI / HTTP       | PASS, both endpoints plus `worktree prune --dry-run --state`   |
| new daemon modules | PASS, exercised live through the endpoints                     |

### Sandbox

One repo with five linked worktrees, each built to land on a different arm of the classifier, plus three real Claude sessions living in three of them:

- `fix-login`, merged into `main` locally
- `feat-export`, pushed then deleted on the remote, so its upstream reads `[gone]`
- `spike-cache`, merged locally and dirty (1 modified, 1 untracked)
- `hold-me`, merged locally but `git worktree lock`ed
- `still-going`, unmerged with no upstream

The scan classified all five correctly: three candidates with the right reason and `branchDeletion`, `hold-me` reported as skipped ("locked"), and `still-going` silently absent. `pr-merged` and `pr-closed` are the two reasons the sandbox can't reach (no GitHub remote), so those stay covered by `worktree-prune.test.ts` only.

### The dirty gate holds on both sides

In the dialog, `a` selects only the clean rows. Selecting the dirty row with `space` checks it but leaves the footer count at `prune 2`; `D` flips it to `will be deleted (D)` and the count to 3, and pressing `D` again withdraws it. The confirm line then reads `Delete 2 worktree(s), 2 branch(es), skipping 1 dirty?`.

The daemon enforces it independently. A real (`dryRun: false`) POST naming the dirty worktree with an empty `allowDirty` was refused, and the directory, the uncommitted work, the branch and the session all survived.

### The write half only accepts what its own scan classified

Every path outside the current candidate list came back `409`: the unmerged worktree, the locked one, and a path that isn't a worktree at all. An empty selection is a `400`. There is no way in through this endpoint to name an arbitrary directory.

### The run itself

Pruning the two clean worktrees stopped both agents (confirmed dead by pid), closed both panes, moved each directory to its `.ccmux-trash-*` sibling and then deleted it, reclaimed the git admin entries, and dropped the matching entries from the agent state file. The picker went from 3 sessions to 1 live on screen.

Two behaviours worth naming because they look like failures and aren't:

- `fix-login` was deleted with `git branch -d`. `feat-export` was **kept**, reported as `not fully merged`. That is the documented `upstream-gone` policy working: a deleted remote branch is a hint, not proof, so it stays on the safe `-d` and an unmerged branch survives with the refusal reported rather than being force-deleted on a guess. The worktree is still gone, which is what was asked for.
- `close pane` treats an already-gone pane as success, which is the common case when stopping the agent closes its own pane.

### One thing worth a look

At sidebar widths the compact footer does its job (`space pick · D dirty · enter 0 · q` fits, and the count stays visible), but the row's second line truncates mid-word: `merged into  [       DIRTY 1m/1u —`. The `press D to include` hint is cut off entirely, which is the one piece of text a user needs in order to know the row is being held back. Not a blocker, and `partitionSelection` is well covered, but the compact branch itself has no test.

### Gaps

- `pr-merged` / `pr-closed` classification and `ghPRStateLookup`: unit tests only, no live GitHub round trip.
- The "an agent is working here" skip: verified through the `locked` skip, which shares the same code path, but not with an agent actually held in `working`.
- Interactive `ccmux worktree prune` (the readline selection and confirm): `parseSelection` is unit tested and the non-TTY refusal was verified, but the prompts themselves were not driven.

A screencast of the picker walkthrough (classify, select, the dirty gate, confirm, and the run completing) is attached separately.

epilande:
### E2E recording

A real destructive run against throwaway sandbox repos under `/private/tmp`, driven with real keystrokes (not a surface demo), on an isolated daemon whose only sessions were the sandbox ones.

1. Picker with 3 sandbox sessions; `W` opens the dialog showing three classified rows with reason, `[claude idle]`, and `DIRTY 1m/1u`, plus "1 not offered" for the locked worktree.
2. `a` selects only the clean rows (count 2). Pressing space on the dirty row leaves the count at 2 — the gate holding.
3. `D` opts the dirty row in (count 3, marked "will be deleted"), and `D" again withdraws it.
4. Confirm reads "Delete 2 worktree(s), 2 branch(es), skipping 1 dirty?"; `y` runs it.
5. Result log shows agents stopped, panes closed, directories moved aside, branches handled. Closing the dialog shows the picker drop from 3 sessions to 1.

Verified on disk afterwards: both directories gone with no trash left, the merged branch deleted, both agent pids dead, both tmux windows gone, git admin entries reclaimed, and the sandbox state file entries dropped with a backup written. The dirty worktree survived with its uncommitted and untracked work intact and its agent still alive.

Also proven separately: a real `dryRun:false` POST naming the dirty worktree with an empty allowDirty was refused with nothing deleted, and every non-candidate path (unmerged, locked, not-a-worktree) returned 409.


https://github.com/user-attachments/assets/ea35d40f-6229-45a8-b695-049f192b6bad

PR #78, opened 2026-07-29
feat(tui): New session dialog with context-aware defaults

### LINKED ISSUE #65: [Feature]: Picker: new session dialog with context-aware defaults

epilande:
Closes #65

## Summary

Adds a new-session dialog to the picker, opened with <kbd>n</kbd> or from the existing context menus. Every field is optional with context-aware defaults, so the fast path is two keystrokes: `n`, Enter.

- **Agent**: spawnable agents (built-ins found on PATH plus custom config agents), `j`/`k` or number keys. Defaults to the selected row's agent, else the last agent spawned from the dialog.
- **Placement**: new window / split right / split down. Defaults to new window.
- **Prompt**: free text, empty by default, delivered through #64's per-agent `promptCommand` handling.
- **Directory** is derived rather than asked: session row → that session's cwd, group header → the group's cwd, no selection → the picker's own cwd. Shown in the dialog so it is never a surprise.
- New `GET /agents` for spawnable-agent discovery.
- Entry points: <kbd>n</kbd>, "New session here" on session rows, "New session in repo" on group headers. Background (paneless) rows work too, using their cwd.

## Design notes

**Placement travels as `callerPane`, not `target`.** #64 gave the two different meanings: an explicit `target` for a new window resolves to `{kind:"window"}` and runs `new-window -a -t @win`, renumbering every later window; `callerPane` resolves to `{kind:"session"}` and appends. For a split, `callerPane` still splits exactly that pane. One field, correct for all three placements. Verified live: spawning a new window with later windows present left their indices untouched.

**Structured for #69.** `NEW_SESSION_FIELDS` in `store.ts` is the single list driving focus order, option-key routing, and rendered rows, and a simplify pass folded the three per-field switches into one `focusedOptionField()` lookup. Adding the worktree destination field is one array entry, one draft key, one case, and one render branch.

## Three defects the live run flushed out

Unit tests would not have caught any of these:

1. **`/agents` and `/spawn` disagreed.** The route read the config file fresh while `POST /spawn` used the daemon's boot-time agent list, so an agent added since the daemon started appeared in the menu and answered "Unknown agent" on Enter. The route now resolves every name through the daemon's own lookup, so the menu can only offer what `/spawn` accepts.
2. **`resolveLaunchPane` enumerated the wrong window.** A bare `tmux list-panes` resolves the *session's* current window, not the caller's, so a picker in a non-current window split an unrelated pane. Now passes `-t $TMUX_PANE`; a popup, which has no pane, correctly keeps the client's current window.
3. **`lastSpawnAgent` never reached disk.** The one-shot picker calls `process.exit(0)` the moment its spawn lands, and the store's 300ms persist debounce never survived it — so persistence failed in exactly the case it exists for. Now written straight through and awaited.

Also fixed: the App tests were writing to the developer's real `~/.config/ccmux/state.json`. `setUIState` is now stubbed, which also closes a pre-existing leak from the `f`/`p`/`b` key tests.

## Verification

- `bun run typecheck` clean; `bun test` **3471 pass / 0 fail** across 154 files.
- Rendered at 100x30, 50x30 and 34x40 headless, plus live in tmux at 200x50 and in a 34-column sidebar rail. At 34 columns the placement options stack so nothing runs past the border.
- Live, against an isolated daemon: split targeted the launch pane rather than the picker's; new window appended without moving later windows; the prompt reached the pane single-quote escaped through the per-agent template; context defaults correct for all three cases; the remembered agent survived process exit and was the default in the next picker; in sidebar mode the split hit the main-area pane, the rail was untouched, and focus did not move. One real Claude spawn answered `OK`.


epilande:
## Verification

Ran the full diff-driven verification against an isolated stack (scratch `CCMUX_HOME`, own daemon on port 2323, own tmux server) so none of this touched the live fleet.

**Tier 0 (static):** PASS. `typecheck`, `knip`, `build`, and `bun test` all green (3471 pass / 0 fail). The six test files this PR touches account for 507 of those.

**Tier 2 (TUI render):** PASS. The dialog was captured at three widths and each responsive branch behaves as written:

- 200 cols renders the full placement row (`1 [New window] 2 Split right 3 Split down`)
- 62 cols drops to the compact labels (`Window` / `Right` / `Down`)
- 36 cols stacks the options vertically so all three stay visible

The picker's footer carries the key hints and the dialog suppresses its own, so the hints appear exactly once. The sidebar, which has no footer, draws its own row.

**Tier 4 (daemon HTTP):** PASS. `GET /agents` returns the nine PATH-gated built-ins with `displayName`, `shortCode`, and `supportsPrompt`. The same route 404s against a daemon running `main`, which confirms the picker degrades to a clear failure rather than a silent one on a stale daemon, exactly as the handler comment describes.

**Tier 3 (spawn wiring):** PASS, verified live with real agents rather than mocks.

- **Split lands beside the launch pane.** Spawning with `Split right` from a picker in window 1 put the new pane in window 1 next to the launcher. Window list stayed `0,1,2`.
- **New window appends without renumbering.** This is the behavior from #64 that matters most here. Windows were `0:agent 1:launch 2:notes` and the picker was launched from window 1, so an insert-at-caller would have taken index 2 and pushed `notes` to 3. Instead the new window appended at 3 and `notes` kept 2. That is `callerPane` resolving to `{kind:"session"}` rather than `{kind:"window"}`.
- **Sidebar spawn does not steal focus and does not split the rail.** With a 30-column sidebar rail on the left and a work pane on the right, spawning from the sidebar split the work pane, left the rail untouched at 30 columns, and left the active pane exactly where it was. This exercises `parseLaunchPane`'s fallback branch, since the active pane at sidebar launch was the rail itself.

## One finding, non-blocking

Inside a real 30-column sidebar rail, the placement labels truncate to the point of being ambiguous:

```
Placement 1 [New     ]
          2  Split
          3  Split
```

Options 2 and 3 are indistinguishable. The stacked branch deliberately prefers the full label over `compactLabel`, and the comment explains why (a stacked row has the width to itself), but at the rail's actual width the box is only 26 columns, which leaves 8 for the label. `New window`, `Split right`, and `Split down` all wrap and render only their first word.

This is cosmetic and the number keys still select correctly. Worth noting that `compactLabel` already exists and would fit; it is just gated on `compact() && !stacked()`. At 36 columns and up the stacked layout renders the full labels correctly, so this only bites the narrowest surface.

## Not covered

Every built-in on this machine has a `promptCommand`, so `supportsPrompt` came back `true` for all nine and the "can't start with a prompt" placeholder and the submit-time guard were never reachable live. Both have unit coverage in `NewSessionDialog.test.tsx` and `App.test.tsx`.

epilande:
### E2E recording

Driven end to end with real keystrokes and three real agent spawns, against an isolated daemon. 92s at 180x46.

Arc: picker → `n` → context defaults (agent, placement, prompt, directory) → j/k navigation → number keys → tab across fields → spawn as Split right → relaunch → spawn as New window → sidebar rail → spawn from the sidebar.

The three headline behaviors are proven against tmux state rather than pixels:

1. **Split landed beside the launch pane** — window 1 gained a second pane, window list unchanged at 0,1,2.
2. **New window appended without renumbering.** Launched deliberately from window 1 with `notes` sitting at window 2, so an insert-at-caller would have taken index 2 and pushed notes to 3. The status bar in frame reads `0:agent 1:launch- 2:notes 3:2.1.220*`.
3. **Sidebar spawn did not steal focus and did not land in the rail** — 30-column rail untouched, the new pane split the work pane, active pane and current window both unchanged. This exercises `parseLaunchPane`s fallback branch, since the active pane at sidebar launch was the rail itself.


https://github.com/user-attachments/assets/5b0811ad-950b-406f-ad9c-0cb11140d072

Issue #70, opened 2026-07-29
[Feature]: Fork a session into a new worktree


epilande:
## Overview

Extends Fork (#66) with a worktree destination: branch the conversation into a new agent running in its own checkout, instead of a split pane sharing the source session's working tree.

Depends on #66 (fork) and #69 (worktree spawning).

## Motivation

An in-place fork shares the working tree, so the original and the fork edit the same files; that fits quick divergence but not trying two real implementations of the same change. With a worktree destination, "the agent is heading down path A, fork it and try path B" gives each path its own branch and checkout, and both can run to completion in parallel.

## Design

- The Fork action gains a destination choice: split beside the source (default, the #66 behavior) or new worktree.
- Fork command construction is unchanged from #66 (`forkCommand` template, `nativeSessionId` gating, Claude Code only in v1). Only pane placement and cwd differ, which is why #66 keeps command construction separate from placement.
- The worktree comes from the spawn creation engine: branch based on the parent session's branch, default name `<parent-branch>-fork`, numeric suffix on collision. Deterministic, no prompts needed.
- The parent's uncommitted changes are not carried over; the fork starts from the branch's committed state. The dialog states this so it never surprises.
- The forked session is tracked like any spawn: hooks fire, the marker pipeline binds it, and the row groups under the repo with the worktree marker.

## Out of scope

- Carrying uncommitted changes into the fork.
- Fork for non-Claude agents; same per-agent verification gate as #66.

## Verification

- Live e2e per the AGENTS.md workflow: start a hook-installed Claude session in a repo, build some conversation, fork into a worktree, verify the new session resumes the history under a new session id, edits in the fork do not appear in the source checkout, and the original session continues unaffected.
- Component test for the destination choice in the fork flow.

epilande:
## Design settled: the stated premise is false, but a proven mechanism exists

**The premise this issue inherited from #66 does not hold.** #66's docs described fork-into-a-worktree as "the same fork command with a different `cwd`". That was disproven live on Claude Code 2.1.220 during #79's review: Claude resolves `--resume <id>` against the **current directory's** project dir (`~/.claude/projects/<encoded-cwd>/`), so a fork into any other directory returns HTTP 200 while the pane prints `No conversation found with session ID: …` and drops to a bare shell. ccmux cannot see the failure — the pane opened, the keys sent, the route succeeded. #69 has since added a guard refusing `fork` + `worktree` so that dead pane is no longer reachable.

[...]

PR #79, opened 2026-07-29
feat(picker): Fork a session into a parallel pane

### LINKED ISSUE #66: [Feature]: Picker: fork a session into a parallel pane

epilande:
Closes #66

## Summary

Adds **Fork** to the picker's session context menu (and <kbd>Shift</kbd>+<kbd>F</kbd>): spawn a new pane running the same agent in the same cwd, whose conversation continues from the source session's history, leaving the original untouched.

- New `AgentDef.forkCommand`, a `{bin}`/`{id}` template. Claude Code: `{bin} --resume {id} --fork-session`. `{bin}` means a `command` preference (wrapper binary) survives.
- **Fork** appears only when the agent defines `forkCommand` **and** the session has a `nativeSessionId`, so pane-tracked-only rows never show it. Gating data reaches the TUI as a `forkableAgents` prop, so a *custom* agent configured with a `forkCommand` gets the item too.
- Placement: splits next to the source session's own pane by default.
- Wire: `fork: "<session-id>"` on `POST /spawn`, resolved by tracked id falling back to native id. The daemon takes agent and cwd from the resolved session and ignores `body.agent`; an explicit `cwd` still wins. `fork` + `resume` or `fork` + `prompt` is a 400.
- CLI parity: `ccmux spawn --fork <session-id>`.

## Safety and the #70 seam

`buildAgentForkCommand` is a separate exported function that knows nothing about pane placement, and the fork path adds no targeting of its own — `split`/`target` mean exactly what they mean for any other spawn. So fork-into-a-worktree (#70) is the same command with a different cwd and destination. Stated in the code comments and in `docs/agent-adapters.md#forking-a-session`.

Four guards, each load-bearing: refuse an agent with no `forkCommand`; refuse a non-string template; **refuse a template missing `{id}`**, because it would silently start a *fresh* session that looks like success; and validate the id against `NATIVE_SESSION_ID_PATTERN` inside the builder rather than only at the route, so the guard travels with the function. Substitution goes through the existing single-pass `substitutePlaceholders` — never string `.replace`, which was a proven injection vector in #64 — with a test that a `{bin}` containing `{id}` is not revisited.

## No `working`-state gate, and why

The issue offered one as a fallback. It isn't needed, verified live: forking a source **10 seconds into a long answer** left the source to complete its turn normally and return to idle, while the fork came up with history as of fork time (the in-flight user message present, the not-yet-written answer absent) sitting at an empty composer rather than auto-running it.

## Verification

- `bun run typecheck` clean; `bun test` 3434 pass / 0 fail across 152 files.
- Live against real Claude Code 2.1.220 on an isolated stack: source session answered ALPHA, forked, new session got its own id, its own marker and its own transcript. **Divergence proven at the transcript level** — the source jsonl contains ALPHA + GAMMA and no BETA; the fork's contains ALPHA + BETA and no GAMMA. Both bound as separate native rows with no reconciler work.
- Picker path end to end: <kbd>Shift</kbd>+<kbd>F</kbd> on a row produced a left/right split of that row's own pane (source at x=101, fork at x=151) with the correct history.

## Claude-only in v1

`docs/agent-adapters.md#forking-a-session` records why each other built-in was left out: a bare `--resume` reopens the *same* id for codex/cursor/copilot, and opencode/pi/omp resume "the most recent" with no id. Adding one later is defining `forkCommand` once its resume semantics are verified against a live original.

## Pre-existing behavior worth knowing

A freshly forked Claude pane has a marker but no daemon row until it takes its first turn, because native tracking needs a transcript. Same for any fresh `ccmux spawn claude`. It self-heals on the first prompt.

E2e recording to follow.


epilande:
End-to-end run of fork on this branch against real Claude Code sessions (v2.1.220), on a throwaway daemon and tmux server so none of it touched the live fleet.

**Continuity and divergence, on screen.** The source session was given two codewords (ALPHA, then DELTA) and left idle. Forking it from the picker opened a pane beside the original, and that pane came up already holding the shared history. Each side then got a different follow-up, and each answered from the history they share plus its own new fact:

- original: "Also remember BETA. Now list every codeword you know." gives `ALPHA, DELTA, BETA`
- fork: "Also remember GAMMA. Now list every codeword you know." gives `ALPHA, DELTA, GAMMA`

The original stayed live and usable the whole time. The two rows carry distinct native session ids, and the transcripts agree with the screen: the source jsonl has BETA and no GAMMA, the fork jsonl has GAMMA and no BETA, and both share the pre-fork turns.

**Menu gating.** Right-clicking a Gemini row (pane-tracked, no `nativeSessionId`) shows Attach / Kill / Restart / Review diff and no Fork. Right-clicking the Claude row shows **Fork** with hint `F`. Bare `f` still toggles hide-idle (list empties, "Hide Idle ON"), Shift+F forks the selected row, and the help overlay lists `F  Fork session`.

One thing worth knowing: Shift+F is a no-op when the selection is sitting on a group header rather than a session row, since it keys off `store.selectedSession()`. That is the picker's normal selection model, not obviously wrong, but it did cost a confused minute.

**CLI and route.** `ccmux spawn --fork <claude-id>` printed `Forked ... into pane %7: claude --resume ... --fork-session` and opened in the source's cwd with no `--cwd` passed. `ccmux spawn --fork <gemini-id>` refused with the actionable "no native session id ... install hooks with 'ccmux setup'" message. `POST /spawn` returned 400 as documented for an unknown id, shell metacharacters in `fork`, an empty string, a non-string, and `fork` combined with either `resume` or `prompt`.

**Static.** typecheck, knip, build, and 3434 tests all green.

Not re-verified live here, both covered by unit tests and by the author's own run: forking mid-turn, and a custom agent's `forkCommand` coming from config.

Verifies `src/daemon/spawn-command.ts` (`buildAgentForkCommand`), `src/daemon/server.ts` (`resolveForkSource` plus the fork branch of `POST /spawn`), `src/lib/agents.ts` (`forkCommand`, `forkableAgentNames`), `src/commands/spawn.ts`, and the picker wiring in `src/tui/App.tsx`.

Screencast of the run is attached separately.

epilande:
> [!NOTE]
> Filmed before the final review fixes. Two things changed after this recording and are not reflected in it: **Fork now sits below Restart** in the context menu (it appears and disappears reactively, so any position above Kill could slide Kill under a hovering cursor), and pressing `F` on a non-forkable row now **toasts the reason** instead of doing nothing. The fork behaviour the video demonstrates — continuity, divergence, and the original being unaffected — is unchanged.
### E2E recording

One continuous take against real Claude Code 2.1.220 on an isolated stack.

The source session is given two codewords on camera (ALPHA, then DELTA). Then in the picker:

1. Right-click a **Gemini** row (pane-tracked, no native session id) — the menu shows Attach / Kill / Restart / Review diff and **no Fork**.
2. Right-click the **Claude** row — Fork is offered, hinted `F`.
3. Bare `f` toggles hide-idle (list empties, then restores on a second press), proving the key split.
4. `Shift+F` forks: a pane opens **beside the original in the same window**, the picker switches to it, and the fork comes up already showing the shared ALPHA/DELTA transcript.

Then the divergence, with both panes visible in one frame:

- **left (original)**: "Also remember BETA. Now list every codeword you know." gives ALPHA, DELTA, BETA
- **right (fork)**: "Also remember GAMMA. Now list every codeword you know." gives ALPHA, DELTA, GAMMA

Shared prefix, divergent tails, on screen simultaneously. The original stayed live and answered normally after being forked.

Independently reproduced at the transcript level: the source jsonl contains BETA twice and GAMMA never; the fork jsonl contains GAMMA twice and BETA never; both share ALPHA; and the two rows carry distinct native session ids.

Not in the video but verified: `ccmux spawn --fork <claude-id>` resumed the history in the source cwd; the same command on a Gemini row refused with the actionable install-hooks message; and all six documented `POST /spawn` 400s (unknown id, shell metacharacters, empty, non-string, fork+resume, fork+prompt).


https://github.com/user-attachments/assets/8029484e-82b5-4432-87b4-be68c186469a

PR #84, opened 2026-07-29
feat(worktree): Spawn a session into a new worktree

### LINKED ISSUE #69: [Feature]: Spawn a session into a new worktree

epilande:
Closes #69

## Summary

Adds a worktree destination to the spawn pipeline: create a worktree and launch an agent in it, from the CLI (`ccmux spawn --worktree [name] [--base <ref>]`) and from the picker's new-session dialog.

- **Creation engine** (`src/daemon/worktree-create.ts`): worktrees land at `<main>/.claude/worktrees/<name>`, the same location agent-created worktrees use. Branch from the slugified name; `--base` selects the start point, default is the main checkout's current branch. Explicit name is both directory and branch; with a prompt and no name the slug derives deterministically from the prompt's first words. No name and no prompt is an error. No model calls, ever.
- **File setup** honors the same conventions Claude Code applies to its own worktrees, so a ccmux-created worktree behaves identically with zero new config.
- **Create-or-open, for explicit names**: an existing worktree of that name is spawned into rather than refused. A prompt-derived name that collides takes the first free numeric suffix (`-2`, `-3`, ...) instead, so parallel spawns on one prompt each get their own checkout. Concurrency is serialized per repo. A leftover _empty_ directory at the target is cleared; anything else (a `.git`, including a broken symlink, or any content at all) is refused.
- **API/CLI**: `POST /spawn` accepts `worktree: { name?, base? }` and then continues through the normal spawn path with the new path as `cwd`, so #64's placement, split direction and per-agent prompt handling all apply unchanged.
- **Dialog**: a `Where` field offering this checkout (default) or a new worktree, with the derived name previewed live as you type.

## The file-setup contract was verified, not assumed

The `worktree.symlinkDirectories` and `.worktreeinclude` behaviors are only partly documented upstream, so they were established by observing what Claude Code actually does against throwaway fixtures. Placement, symlink direction, missing-source handling, and glob copying all match. The `.worktreeinclude` **dual filter** — a file must match a pattern _and_ be gitignored — was confirmed by modifying a tracked file first and verifying the worktree received the _committed_ content, proving the include pass does not copy over the top. That pass is delegated to git rather than reimplemented, because gitignore semantics (negation, anchoring, directory-only patterns, precedence) are too subtle to hand-roll.

Two deliberate divergences from Claude Code are recorded in the code with reasoning: branch names are the bare slug (not `worktree-<name>`), and no session lock is taken.

## A composition hole this branch closes

Merging #66's fork with this feature made a case reachable that neither PR could have found alone. `resolveForkSource` refuses a fork whose explicit `cwd` differs from the source's — but a fork into a worktree sends **no** `cwd`, so that guard passes and the relocation only materializes later when `worktree` overwrites the spawn cwd. Measured with the refusal disabled: `{fork, worktree:{name}}` answered 200 and created the worktree, producing the dead pane that a cross-directory fork always produces (Claude resolves `--resume` against the current directory's project dir).

Now refused, **before** `createWorktree`, so a rejected request leaves no directory and no branch behind. The test uses a _named_ worktree deliberately: the bare `{fork, worktree:{}}` form is rejected only by accident, since a fork carries no prompt and therefore no name can derive — a test written against that form would pass against no guard at all.

## Review fixes

The request-changes round landed as four commits on top of the reviewed head:

- **M1, and it changes already-merged prune behavior, not just this feature.** A branch created from base sits exactly on the base tip, and a commit is an ancestor of itself, so prune classified a just-created worktree as `merged-locally` with a safe branch delete, and only a `working` session gated it. `isMergedInto` now excludes a branch whose tip equals the base's own tip (this deliberately also suppresses fast-forward-merged branches, which `pr-merged` and `upstream-gone` still catch), and the candidate gate skips a worktree with a session bound in **any** state (working, idle, or waiting). Anyone bisecting prune behavior: the change is in this PR, commit `4ae8baf`.
- **M2**: derived-name collisions no longer create-or-open; they suffix (see Summary above). Three concurrent spawns of one prompt yield three worktrees, verified against real git.
- **S1/S2**: the leftover-directory guard no longer recursively clears. Only an empty directory is removed; a broken `.git` symlink or a nested repo's files survive with a refusal instead. This deliberately departs from #69's "no `.git` inside, so remove it" design: `git worktree add` refuses a non-empty target anyway, so clearing bought nothing and risked files.
- **S3/O1**: the response's `worktree` object gains `branchCreated: boolean` and `base?: string` (absent when nothing was cut), and the CLI reports "new branch Y from Z" / "existing branch Y" / "reusing" honestly. Response-only additions; request parsing (`worktree: { name?, base? }`) is unchanged.
- **S4**: `--base` without `--worktree` is rejected before `ensureDaemon()`, so argument validation can no longer boot a daemon, and the CLI test is hermetic.
- The escape-refusal test's final assertion normalized back to a directory the test itself created; it now asserts the filesystem on both guard halves and goes red against a weakened guard.
- Docs (O2): README spawn flags and prune safety rules, AGENTS.md and architecture.md module entries. Prettier drift (O3) reverted.

Every added or changed test was proven able to fail by reverting the specific fix and watching it go red, including a proof that the `rev-list --count == 0` alternative to M1 would delete the `merged-locally` reason for genuinely merged branches.

An adversarial gpt-5.6-sol pass over the fix commits then caught a real hole in the M1 fix itself, closed in three follow-up commits:

- `db511d5`: the tip-equality exclusion was a per-base `continue`, so with local `main` behind `origin/main` (the ordinary not-yet-pulled state, which prune's own fetch produces) a fresh worktree still classified merged via the origin/main ancestry check. Matching any eligible base tip now answers false for all bases; regression test uses a real remote fixture with local main reset behind origin.
- `51e1496`: the empty-leftover clear swapped its recursive `rm` for an atomic non-recursive `rmdir` (content appearing between check and delete now fails instead of being taken), and a derived name always passes `-b` so a branch minted mid-create by an external actor fails loudly instead of being silently reused.
- `0d581f9`: the README safety rule scoped to what the code guarantees (a branch still sitting on a base's tip is never classified as merged), since a worktree cut from an older ref can classify immediately, with the session gate as its protection.

Known accepted limitation, stated here so it is a decision rather than a surprise: a zero-commit worktree whose base has since advanced past it, or one cut from an older ref, classifies `merged-locally` once no session is bound. The wide session gate is what protects it while an agent is attached; without one it is indistinguishable from debris, which is what prune exists to clear.

## Verification

- `bun run typecheck` clean; `bun test` **3630 pass / 0 fail** across 155 files after the review fixes (3611 at the reviewed head).
- The reviewer's M1 and M2 fixtures were re-run independently against the fixed modules: a zero-commit worktree is not a candidate with no session, an idle session, or a waiting session; a genuinely merged branch still classifies `merged-locally`; colliding prompts yield `fix-the-flaky` and `fix-the-flaky-2`; an explicit name still opens.
- Live against an isolated stack (own config home, port, and tmux server): explicit name creates the worktree and branch and opens the pane inside it; both file-setup conventions apply (a symlinked directory and a genuinely copied file); prompt-derived naming produces the issue's own example; `--base` lands on the right tip; #64's per-agent prompt handling applies unchanged through the worktree path; orphan handling works both ways; and the daemon detects the new session as a worktree, grouping it under the repo with the marker.
- Dialog verified live end to end: prompt → Tab → select → Enter created the worktree and opened the pane in it.
- A rendering bug caught only by live capture: a long derived name pushed the dialog's right border off screen. The name is now budgeted and truncates, with a test asserting every box row still closes with its border — content assertions structurally cannot see a broken frame.

## Second review round

A second full pass over the reviewed head `0d581f9` (four parallel review lanes, then adversarial verification of every finding against fixtures) landed as five commits, tip `dc3dc41`:

- **`8dcc2c7`**: two silent wrong-answer bugs in file setup, both reproduced before fixing. The copy half's `existsSync` guard read a dangling checked-out symlink as absent and wrote THROUGH it to a path outside the worktree while reporting the entry as included; the guard now matches the symlink half. And `ls-files` C-quoting silently dropped non-ASCII `.worktreeinclude` entries; both calls now run `-z` with a NUL split.
- **`5f09d16`**: worktree creation was the handler's first side effect, so a bad agent name or stale pane orphaned a checkout and branch that the tip-equality rule keeps prune from ever offering. Creation now runs after the agent, command, and placement validations; only tmux-level failures can still leave a worktree behind, and the retry note now splits by name mode (a derived name suffixes on re-run rather than reusing).
- **`1388a68`**: the dialog's New worktree destination always failed on an empty prompt (its default state) with server advice the dialog has no field to act on. It now refuses locally with actionable wording and the selected row shows `New worktree (add a prompt)`. Verified in a live capture.
- **`c533c66`**: `--base` on a reused worktree did nothing without a word; the reuse line now says so. The CLI's worktree wire shape is now pinned by tests (the daemon rejects `worktree: true`, so a refactor away from the object shape previously would have passed the suite).
- **`dc3dc41`**: dialog docs for the picker's Where field and base semantics; every cheatsheet line copy-paste runnable again.

Verified findings deliberately not taken, so they are decisions rather than oversights: a resume+worktree guard (the hazard predates this PR via `--cwd`, and a correct guard must key on "resumeCommand defined and lacks `{id}`" or it would break the agents whose resume carries the id); releasing the repo lock before file setup (the cheap version hands a concurrent create-or-open a half-set-up checkout); and surfacing the resolved worktree name in the sidebar toast.

After the fixes: `bun run typecheck` clean, `bun test` **3644 pass / 0 fail** across 155 files, every new test proven able to fail by reverting its fix.

E2e recording of the dialog path is in the comments below (recorded at the reviewed head; the fix commits change error paths and collision naming, not the recorded flow).


epilande:
## Review: request changes. 7/10 as-is, 9/10 with the two Must-fixes.

The engineering here is strong, and the areas I expected to find problems in came back clean under execution: path escape, option injection, the `isInside` guard, and the `.worktreeinclude` dual filter all hold up against deliberately hostile input. What this PR does have is two composition problems, both of the same shape as the fork+worktree hole the PR body already describes and closes: two individually correct features that misbehave when combined.

Method note: findings labelled **VERIFIED** were reproduced by executing the real modules (`createWorktree`, `scanRepos`, `resolveWorktreeIncludes`, the CLI) against throwaway git fixtures under `/private/tmp`. Findings labelled **READ** come from tracing code without running it. Nothing ran against this repo or any live worktree; all fixtures are removed.

---

## Must-fix

### M1. Prune offers a worktree this feature just created, and stopping it kills the agent. VERIFIED

`src/daemon/worktree-create.ts:26-32` composed with `src/daemon/worktree-prune.ts:496`, `:504`, `:515`, `:534`.

A worktree created here branches from base, so its tip _is_ base, so `isMergedInto` returns true and `reasonFor(null, true, false)` yields `merged-locally`. It carries no lock (divergence 2, deliberate), and it is clean, because file setup writes only gitignored copies plus symlinks that `readDirtyState` already exempts. So `dirty` is false, no `D` opt-in is required, and `branchDeletion` is `"safe"`.

Running `scanRepos` on a fixture immediately after `createWorktree` returned the new worktree as a removal candidate, reason `merged-locally`, detail `merged into main`, `dirty=false`, for sessions `[]`, `idle`, **and `waiting`**. Only `working` produced a skip, because the gate at `:504` is `status === "working"` and nothing else.

Failure scenario: a user spawns an agent into a new worktree. The agent finishes its turn and sits idle, or is waiting on a permission prompt, having so far only read files. The user later presses `W`, sees the worktree they created a minute ago described as finished work, and confirms. `stopSessions` (`worktree-prune.ts:756`) sends SIGTERM to the agent, closes its pane, renames the directory to trash and deletes it, then deletes the branch. The point-of-no-return re-check at `:903` only re-reads dirty state; it never re-checks the session.

The file header's justification does not hold for this population. It says a ccmux-created worktree "becomes prunable the moment its agent exits". It is prunable the moment it is created. Relatedly, `README.md:336` advertises "a worktree with a **working** agent is never offered" as the safety rule, and this PR is what makes that rule insufficient.

Suggested fix, smallest first: take the lock Claude Code takes and release it when the agent exits, which restores exactly the invariant prune already depends on. Alternatives are widening the prune session gate from `working` to any live session, or suppressing `merged-locally` when the branch has no commits beyond its base and a session is bound.

### M2. Prompt-derived names collide constantly, silently placing two agents in one worktree. VERIFIED

`src/daemon/worktree-create.ts:82-90` (three words, no stopword handling) with create-or-open at `:426-448`.

Stopwords consume two of the three word slots, so realistic prompts collide far more often than the docstring's `fix-sidebar-flicker` example suggests:

- `"fix the flaky test in the sidebar renderer"` and `"fix the flaky test in the binder"` both derive `fix-the-flaky`
- `"add a test for the prune trash rename path"` and `"add a test for the worktree include filter"` both derive `add-a-test`

Spawning all four in sequence produced two worktrees. The second and fourth calls returned `created=false` and dropped their agent into the first agent's directory, on the first agent's branch.

This breaks the feature's own headline use case, the one the module header cites ("start three agents on this"). The picker is worse than the CLI here: the CLI at least prints `Reusing worktree`, while the dialog surfaces nothing.

Suggested fix: create-or-open is right for an **explicit** name, since that is documented intent, and wrong for a **derived** one, where no user intent stands behind the collision. When the name was derived and the target already exists, suffix it (`-2`, `-3`) rather than opening. Dropping stopwords or raising `SLUG_WORDS` helps but does not close it.

---

## Should-fix

### S1. A broken `.git` symlink defeats the refuse-if-`.git` guard, and the directory is recursively deleted. VERIFIED

`src/daemon/worktree-create.ts:455-470`. `existsSync` follows symlinks, so a `.git` symlink whose target is missing reads as absent and the whole tree is cleared. A fixture containing `.git -> /nonexistent` plus `precious.txt` returned ok, and `precious.txt` was gone.

This file already contains the right helper: `isSymlink()` at `:363`, written for precisely this hazard in the file-setup pass. It just is not used in the guard. Reachability is admittedly low, since a real linked worktree's `.git` is a file rather than a symlink, but the fix is one line and the blast radius is an unbounded recursive delete.

```ts
const dotGit = join(path, ".git");
if (existsSync(dotGit) || isSymlink(dotGit)) {
  /* refuse */
}
```

### S2. The same guard inspects only the top level, so a nested repository is destroyed. VERIFIED

Same lines. A fixture with `<path>/sub/.git` and `<path>/sub/someones-work.txt` was recursively removed and the work was lost. Since `git worktree add` refuses a non-empty target anyway, refusing any non-empty directory this run did not create is strictly safer than clearing it, and costs the user one explanatory message rather than their files.

### S3. The base defaults to the main checkout's branch, not the caller's. VERIFIED

`src/daemon/worktree-create.ts:139-166`, called from `src/daemon/server.ts:2356` with `gitInfo.mainRepoRoot`.

With main on `main` and a worktree `alpha` on branch `alpha` carrying its own commit, creating `beta` produced it at `main`'s tip, without alpha's commit. So an agent already working inside a worktree that spawns a sub-worktree silently gets it cut from `main`.

The doc comment's claim, that "a user working on a release branch gets worktrees off that branch", holds only when the caller is in the main checkout. With nine worktrees in this repo that is the minority case. Nothing reports it either: `WorktreeCreation` (`:169-179`) carries `branch` but not the base it was cut from, so neither the CLI line nor the dialog shows where the branch came from.

Suggested fix: add `base` to `WorktreeCreation` and print it, and consider defaulting to the caller's HEAD rather than the main checkout's.

### S4. The new CLI test starts a real daemon and leaves it running. VERIFIED

`src/commands/spawn.test.ts:228-240` shells out to `bun run src/index.ts spawn claude --base main`. In `src/commands/spawn.ts`, `await ensureDaemon()` sits at `:129` and the `--base requires --worktree` check at `:147`, so the daemon boots before the argument is rejected.

Reproduced on an isolated port with a scratch config home: the command printed `Starting daemon...`, exited 1 with the correct message, and left a daemon listening. On a developer machine that binds the shared port 2269 with the branch build, which is the known hazard of a worktree-build daemon taking over the machine's real sessions and hooks.

The fix is also a product improvement: move the validation above `ensureDaemon()`. Pure argument validation should never need a daemon, and it makes the test hermetic.

---

## Optional

### O1. `created: true` when an existing branch was reused. VERIFIED

`src/daemon/worktree-create.ts:503`. The branch reuse at `:478-487` is correct and intentional, but the result reports `created: true` regardless, so the CLI prints `Created worktree fix-thing on branch fix-thing` for a branch that may already carry twenty commits. Worth distinguishing branch-created from worktree-created.

### O2. Documentation gaps. VERIFIED

`README.md:263-271` lists every other spawn flag and omits `--worktree` and `--base`. `AGENTS.md:96` lists "Worktree Prune" as a daemon module with no "Worktree Create" counterpart, and the concern-to-path table at `docs/architecture.md:303` has no row for the new module.

### O3. Prettier drift, introduced by this PR. VERIFIED

`src/tui/store.ts:70` (the `NewSessionField` union fits on one line) and `src/tui/App.tsx:2115` (`onSelectDestination` is misindented). `origin/main` is prettier-clean for both files and this branch is not. There is no CI format gate, so this is cosmetic.

---

## Verified safe

Stating these explicitly, because they were the obvious places to suspect and they all held:

- **Path escape and option injection are fully blocked** by `slugify` (`worktree-create.ts:62-69`). Tested `..`, `../..`, `../../../etc`, `/etc/passwd`, `a/../../b`, `..\..\win`, `-rf`, `--force`, `-b`, `HEAD`, `refs/heads/main`, NUL and control characters, newlines, unicode lookalikes, `@{upstream}`, `foo.lock`, and a 300-character name. Every one either resolves to a safe single component under `<main>/.claude/worktrees/` or is refused. No git call goes through a shell.
- **`--base` cannot inject a flag.** `--force` and `-b` are rejected by `rev-parse --verify` as unknown revs. A tag base works and branches from the tag. A detached main checkout yields the literal `HEAD`, which `worktree add` accepts, exactly as the comment claims.
- **A symlinked target path is unlinked rather than followed.** This was the catastrophic candidate. A fixture where `.claude/worktrees/linked` was a symlink to a directory of user files left that directory and its contents completely intact; `fs.rm` unlinked the symlink.
- **`isInside` (`:252-259`) holds against untrusted repo content.** Against a hostile `.claude/settings.json` carrying `../../../../../../private/tmp/ESCAPED`, an absolute path, `.git`, a nested entry, and a missing entry: both escapes were refused with nothing created outside, the missing source was skipped, the nested entry was linked harmlessly, and the worktree's own `.git` was never touched.
- **The `.worktreeinclude` dual filter is correct and genuinely delegated to git.** On a purpose-built fixture: a tracked file matching an include pattern, modified in the main working copy, arrived with the **committed** content; an untracked file matching a pattern but _not_ gitignored was correctly dropped; an ignored file matching no pattern was dropped; and an unanchored pattern matched at depth, per real gitignore semantics.
- **Concurrency is correct.** Five simultaneous same-name creates yielded exactly one `created=true` and four opens. Five different names in one repo all succeeded with consistent git state.
- **Lifecycle composition with prune's trash rename is clean.** Registered-but-directory-missing produces a clear, actionable refusal; after `git worktree prune` the existing branch is correctly reused; after a full prune the name recreates cleanly. No resurrection and no corruption.
- **The fork+worktree refusal fires before `createWorktree`, and its test is load-bearing.** It uses a named worktree against a real repo and asserts 400, plus no `.claude/worktrees` directory, plus zero tmux calls. With the guard removed the request would answer 200 and create the directory, failing three assertions.
- **No `worktree-` branch-prefix coupling** anywhere in `src/`, and `WORKTREE_DIR` is not duplicated, because prune discovers worktrees through `git worktree list` rather than by path convention. Nothing to drift.
- **The TUI importing a daemon module is established precedent**, not a new layering break: `PruneDialog.tsx` already imports from `worktree-prune`.
- **`bun run typecheck` clean, `bun test` 3611 pass / 0 fail across 155 files.** Reproduced independently; matches the PR body exactly.

On the AGENTS.md rule about wiring new daemon modules into `src/daemon/index.ts`: `worktree-create.ts` does not need it. That rule targets lifecycle-owning modules with start/stop, and its sibling `worktree-prune.ts` is likewise absent from `index.ts`. Both are called on demand from `server.ts`. No `as any` anywhere in the diff.

---

## Coverage boundary

So the next reader knows what was and was not looked at:

- **Not run:** the full daemon plus tmux end-to-end stack. Everything above was exercised at the module and CLI level against real git. The PR body reports live e2e verification, and I did not attempt to reproduce it.
- **Not reviewed in depth:** the TUI rendering changes beyond reading `NewSessionDialog.tsx`, its tests, and the store and App wiring. I did not launch the picker or sidebar to capture frames, so the dialog's live rendering at real breakpoints rests on the PR author's own capture plus the border-assertion test, which does look genuinely load-bearing.
- **Not settled:** whether `withRepoLock`'s key (the raw `mainRepoRoot` string) can ever see two spellings of one repo. `getGitInfo` uses `rev-parse --path-format=absolute`, which git resolves against its own realpath'd cwd, so this appears safe, but I did not prove it against a symlinked repo path.
- **Not covered:** interactions with `ccmux invoke`, transcript search, and the notification path. I did check prune, spawn placement, fork, worktree detection, and the dialog.
- Four parallel static reviewers were also running against this diff and had not reported when this was written, so everything here is from a single reviewer.

epilande:
## E2E verification: the feature's own claims hold; two composition findings reproduce

Ran the diff-driven verification against `f774848`, including a real Claude session driven through the picker dialog. Everything ran in a throwaway repo under `/private/tmp`, against a private daemon on port 2337, on its own tmux server with a scratch `CCMUX_HOME`. No real repo, worktree, or session was ever a target.

**Verdict: the feature does what it claims, but I would not call this clear to merge.** The two Must-fixes in the review above reproduce independently for me, and both are about what happens after the spawn rather than the spawn itself.

### Tiers

| Tier         | Result                                                                                                                      |
| ------------ | --------------------------------------------------------------------------------------------------------------------------- |
| 0 static     | PASS — typecheck, knip, build, 3611 tests / 0 fail                                                                          |
| 2 TUI render | PASS — the `Where` row renders at 200, 60, and 34 columns with no clipping, and the derived name updates live per keystroke |
| 3 detection  | PASS — a real Claude session lands in the new worktree and is tracked on the new branch                                     |
| 4 CLI + HTTP | PASS — `--worktree`, `--base`, and the `worktree:{}` wire field all behave                                                  |

Tiers 1 and 5 were not applicable to this diff (no changed pure-logic module lacked a covering test, and no agent definition or config schema changed). **Tiers 0, 2, 3 and 4 are the only ones I exercised, and a pass on them is not a pass on the PR** — see the two reproduced findings below, which live in the composition between this feature and prune rather than in any single tier.

### The dialog path, end to end

Typed a prompt, watched `Where` derive `fix-nav-bug` as it was typed, tabbed over, chose **New worktree**, pressed Enter. The pane opens inside `.claude/worktrees/fix-nav-bug`, `git worktree list` gains the entry, the branch exists, and the picker row shows the session as `v84repo/fix-nav-bug:fix-nav-bug`.

**A 38-second screencast of exactly that run is exported and waiting to be attached** — it is not embedded here because the inline player needs a browser upload I deliberately did not attempt. Whoever picks this up can drag it in from:

```
~/Downloads/ccmux-pr84-spawn-worktree-1785341601.mp4
```

It shows the before-state, the name deriving live as the prompt is typed, the destination choice, the spawn, the resulting picker row going working then idle, and then the concrete proof: the pane cwd inside the worktree, `git worktree list`, the branch, the `node_modules` symlink, and the copied `.env.local`.

The file-setup promise is the part that makes "behaves identically to an agent-created worktree" true rather than aspirational, and it holds:

- `node_modules` symlinked absolutely back into the main checkout
- `.env.local` (gitignored, untracked) copied in with the main checkout's content
- a control file that was untracked but **not** gitignored was correctly left out, so the dual filter really is a dual filter

Also confirmed, off camera: `--base release-x` lands on that tip (proved by a file that exists only there); create-or-open reuses on a second spawn; three **concurrent** spawns of one name yield exactly one `created:true` and two opens, one worktree, one branch; a fork into a new worktree is refused _before_ anything is created, leaving no directory and no branch; `--base` without `--worktree` exits 1 rather than being silently inert; spawning from inside a linked worktree still places the new worktree under the main checkout rather than nesting it; an existing branch of the same name is adopted rather than recreated; and a foreign directory containing a `.git` is refused with its contents intact.

### Independently reproduced from the review

**M1 — a just-created worktree is immediately prunable, and an idle or waiting agent does not protect it.** Ran `scanRepos` against a fixture straight after `createWorktree`:

```
session=working  -> skipped ("an agent is working here")
session=idle     -> OFFERED (reason=merged-locally, branchDeletion=safe, sessions=1)
session=waiting  -> OFFERED (reason=merged-locally, branchDeletion=safe, sessions=1)
```

With no session at all it comes back `merged-locally`, `dirty=false`, `branchDeletion="safe"`. The gate at `worktree-prune.ts:504` is `status === "working"` and nothing else, so an agent that has finished a turn, or one sitting on a permission prompt, is removable along with its branch. This is the feature creating its own prune candidates, and `README.md:336` currently advertises the weaker rule as the safety guarantee.

**M2 — prompt-derived names collide on realistic prompts.** Stopwords eat two of the three slots:

```
COLLIDE: "fix-the-flaky"    vs "fix-the-flaky"
COLLIDE: "add-a-test"       vs "add-a-test"
COLLIDE: "update-the-readme" vs "update-the-readme"
```

Because create-or-open is unconditional, the second spawn silently lands in the first agent's worktree on the first agent's branch. The CLI at least prints `Reusing worktree`; the dialog says nothing.

### One rendering observation of my own

The derived-name preview elides earlier than you would expect. `MAX_WIDTH` is 65, so the side-by-side budget is 27 columns and `New worktree: ` spends 14 of them, meaning any slug over 13 characters truncates _at every terminal width_. The module's own example, `fix sidebar flicker`, renders as `New worktree: fix-sidebar-…`. On a narrow rail it degrades to `Worktree: …`, showing no name at all. The worktree created is always correct, so this is cosmetic, but the row exists to answer "where did that branch name come from" and usually cannot finish the sentence. It also hides exactly the collisions M2 describes.

### Coverage gaps

- `.worktreeinclude` was exercised with a flat file and simple patterns. Negation and directory-only patterns are delegated to `git ls-files`, so they are git's semantics rather than this PR's, and I did not probe them.
- The "registered but directory missing" recovery path was not triggered live; unit tests only.
- I did not run the prune **write** half against a live agent. M1 above is the read half (what prune offers); that the removal then stops the session is read from `worktree-prune.ts:756` rather than executed, since executing it means killing a real agent to prove it.

epilande:
E2E run of the dialog path (screen recording, sped up): typing a prompt while the `Where` field derives the worktree name live, selecting **New worktree**, and the spawn landing in `.claude/worktrees/fix-nav-bug` with the picker row going working then idle. The tail is the proof: the pane cwd inside the worktree, `git worktree list` gaining the entry, the new branch, the `node_modules` symlink, and the copied `.env.local`. Recorded at the reviewed head `f774848`; the review-fix commits change collision naming and prune classification, not this path.

https://github.com/user-attachments/assets/4d12b22e-6dfb-451d-a51c-554f22a37de6

PR #86, opened 2026-07-30
fix: Post-merge hardening across daemon, prune, fork, and TUI

epilande:
## Overview

Hardens the daemon, prune, spawn, fork, and dialog paths with the 21 verified findings from the post-merge review of the #67/#64/#68/#65/#66 arc, one commit per item with the item ID in the subject.

## Motivation

Five independent reviewers swept the seven commits merged between `ba37949` and `02fe34e`, a second adversarial pass re-verified every finding against the actual code with live repros, and a third model reviewed the results independently. Everything that survived is either a wrong answer waiting for a trigger (fail-open PR lookups on the destructive prune path, a fork model that refused valid destinations, a frozen TUI cell that looks authoritative) or a false safety comment the next reader would trust. F1 (the zero-commit worktree misclassification) already shipped with PR #84; this PR is the rest.

## Changes

One commit per item, in dependency order (S3 before S2 before S1; prune after #84's F1). See each commit body for the item's what and why.

### Docs (1)

- **753997e** `D` - Corrects the false `AGENTS.md` dialog touchpoints (the trio was folded into `focusedOptionField()`) and `architecture.md`'s launch-pane claims (resolved per spawn, self-exclusion is sidebar-only)

### Spawn and fork (5)

- **eea67db** `S3` - Refuses `\` and `$'...'` in `promptCommand` templates; both guard bypasses were re-reproduced executing a canary in sh, bash, and zsh before fixing
- **db51d6a** `S2` - Adds a `{path}` placeholder with the same strict single-quoted-context guard and escaping as `{prompt}`; the red proof runs a naive substitution and proves the payload executes, then proves the shipped escaping yields the path byte-exact
- **ba795f0** `S1` - Fork resumes by absolute transcript path (`claude --resume '{path}' --fork-session`) instead of the false directory-scoped model; the dead `cwd !== session.cwd` guard is removed; `{id}` stays a first-class placeholder so config can revert to id-based resume if the undocumented path form ever breaks; refuses with a 400 before any pane when the transcript is missing or unreadable
- **a2a902a** `S14` - Caps spawn prompts at 256 KiB (mirroring `MAX_INVOKE_PROMPT_BYTES`) with a 400 before any pane exists, and kills the pane when a post-creation tmux spawn throws instead of exiting non-zero
- **4a1b373** `O6` - Validates `detach` on the wire (`"false"` the string used to mean detached); audit of the body extraction found no other unvalidated boolean field

### Foundation and project derivation (4)

- **dbbc6a5** `S5` - Derives `mainRepoRoot` for submodule and `--separate-git-dir` layouts via `--show-toplevel`; a worktree of a bare repo stays null by explicit decision, and the wrong `EnrichedSession` invariant doc is fixed
- **58cf63d** `O1` - Drops the `?? session.project` fallback that fabricated a `<repo>/<worktree>` label from a cwd basename
- **079c467** `S4` - Restores the `$HOME` boundary when git itself answers: a literal `~/.git` dotfiles setup no longer collapses every directory under home into one group (the bare-repo dotfiles pattern is unaffected, and tested as such)
- **71909bf** `O3` - Makes the old-git warning reachable (real old git emits five lines, the length gate returned first) and documents git 2.31 as the effective minimum in the README

### Prune hardening (6)

- **2879b9f** `S7` - The `gh` PR lookup returns a discriminated result; when PR state cannot be determined the worktree is skipped with a visible reason instead of silently reading as "no PR". With PR state unknown, nothing can reach `pr-merged` (the force-delete path)
- **62c0f86** `S7` follow-up - Closes the fail-open escape hatch the review round found: remotes that are not recognizably github.com (GitHub Enterprise domains, ssh aliases, `insteadOf` shorthands) now read as unknowable; only a repo with zero remotes takes the permissive branch. `git config --local` prevents global-config fallthrough from masquerading as the zero-remote case
- **4f90f7b** `S8` - Moves the dirty re-check after agent shutdown, immediately before the rename, so the SIGTERM write window is actually covered; the comment and README line now tell the truth
- **b40b4cb** `S9` - SIGTERM, verify, escalate to SIGKILL, verify again; a process that still answers refuses the candidate instead of being deleted underneath. Implemented in-module rather than reusing `lifecycle.ts`'s `killAndVerify` to keep the existing injectable test seams
- **3c89823** `S10` - Normalizes the candidate path on both dirty opt-in compares, so a case-insensitive filesystem can no longer produce a "was not opted in" dead end
- **04df5e3** `O7` - Over-cap `paths`/`allowDirty` lists 400 with the count and the cap instead of silently truncating; the raw length is checked before any per-entry realpath work

### TUI and dialog (5)

- **0764fc8** `S6` - The standalone `branch` column label is a thunk, so the cell tracks checkout changes instead of freezing at mount-time value; pinned by a two-render regression test
- **72eae3c** `S11` - `setLastSpawnAgent` flushes pending UI state even when the agent is unchanged (the default branch), so a pending toggle survives the spawn's `process.exit`
- **ce685a2** `S12` - The spawn guard is held through persistence and dialog close, so an Enter during the state write cannot spawn a second pane; a failed spawn still re-enables the dialog
- **51d4b63** `S13` - Both group new-session entry points (key and context menu) resolve through `newSessionContext`, and project groups resolve to `mainRepoRoot` instead of a nondeterministic sibling worktree
- **bcd6793** `O5` - Any built-in with an overridden `executable` stays in the agent list (a `BUILTIN_BINARIES` map subsumes the Claude-only exemption). One existing test pinned the old wrong behavior and was rewritten; see the commit body

### Daemon reads (1)

- **854144f** `O2` - `scanTranscriptHead` uses an expanding read (modeled on `readLogTail`'s doubling window, 8 MiB ceiling) so a 264 KB to 626 KB cwd-bearing line no longer falls past a fixed cap; measured against a 496-transcript corpus during review

## Review round

The five prune commits were written under the July 30 API incident and got a dedicated adversarial review pass afterward. It confirmed four of the five safety properties outright and found one Must-fix (the S7 escape hatch above, fixed in **62c0f86** with the GHE case proven end to end before and after), plus nits folded into the same commit (an orphaned `sleep 30` per test run, a double negation).

Deliberately not taken, recorded so they are not re-litigated: O4 is a disclosure on issue #68, not code (a naive patch-id upgrade would widen F1); O8 (reason strengthening inside the confirm window) needs a wire change for a bounded consent-drift issue and stays open.

## Breaking Changes

None. The `{path}` fork template is a behavior improvement with a declarative `{id}` fallback; all nine built-in `promptCommand` templates are asserted to still pass the tightened guard.

## Testing

- [x] Baseline: `bun run typecheck` clean, `bun test` 3725 pass / 0 fail (155 files), re-verified independently at HEAD
- [x] Every behavioral fix carries a red proof: its test was run against the unfixed code and shown to fail before the fix was applied
- [x] Prune, live: isolated daemon (own port, `CCMUX_HOME`, private tmux socket) over throwaway git fixtures with a failing `gh` stub. S7 skips github.com and GHE remotes with visible reasons while a zero-remote repo still classifies from local evidence; F1's zero-commit worktree is still excluded (and provably evaluated, not filtered earlier); S10's case-mismatched opt-in removes; O7 400s at 501 paths and accepts 500
- [x] S9, honestly: the SIGKILL escalation is unreachable live because the classifier never offers a worktree with a bound session (proven live: starting an agent in a candidate flipped it to a skip on the next scan). It is defense-in-depth, covered by unit tests including a real SIGTERM-trapping process
- [x] Fork, live on Claude Code 2.1.220: a session anchored in a linked worktree forked into the main checkout; the fork knew the source's fact, took a new fact the source transcript never saw, wrote a distinct transcript (different inode, nlink 1), and the source session stayed alive; an unreadable transcript 400s with no pane created. The `--resume '{path}'` escaping was proven byte-exact through sh, bash, and zsh against a path containing quotes, `$(...)`, and backticks
- [x] Spawn, live: an over-cap prompt and a string `detach` both 400 with zero panes created; a normal spawn on the same daemon succeeds
- [x] TUI, live: the picker's branch column updated in place (`main` to `feature/live-check`) in a still-running picker; the new-session dialog was captured un-clipped with a project group resolving to the repo root
- [x] Foundation: real-git fixtures (mkdtemp + actual `git init`/`submodule add`/`worktree add`) for submodule, `--separate-git-dir`, bare-repo-worktree, and a `$HOME` that is itself a git repo

One pre-existing observation from live testing, not addressed here: an idle pane-tracked session gets no unprompted SSE delta from a bare branch change (nothing broadcasts on git-info drift alone), so the S6 fix was exercised through a real `session_updated` delta triggered via the mark-seen endpoint.

## Merge strategy

Per the work plan: this PR is deliberately 22 single-item commits, and a single 22-item squash spanning the daemon, the TUI, prune classification, and shell escaping is a bad bisect target. Prefer a merge commit, or squash by area if preferred.


epilande:
ccmux-verify-pr e2e recording (Claude on the picker, hooks mode; screen recording, sped up). Verifies: the spawn path (`spawn-command.ts`, `server.ts`) and live state rendering (`store.ts`, `SessionItem.tsx`) through idle -> working -> waiting -> idle. Representative run; these core changes affect all agents. The fork (S1) and prune (S7-S10) behaviors were verified live off-camera; see the PR body Testing section.

https://github.com/user-attachments/assets/454220a0-9cb1-4d0a-ae2e-58b66a6d5da8

PR #87, opened 2026-07-31
test(tui): Await sidebar hydration instead of sleeping

### LINKED ISSUE #74: [Bug]: Flaky sidebar-hydration tests in App.test.tsx gate on a fixed sleep
## Summary

Two tests in `src/tui/App.test.tsx` fail intermittently under full-suite load while passing in isolation. Both gate on a fixed `await new Promise((r) => setTimeout(r, 10))` standing in for "the hydration effect has flushed", which is exactly the shape that breaks when the machine is busy.

- `sidebar hydration with null state does not clobber active-pane default` (test at ~line 280, gate at ~line 309)
- `sidebar hydration with non-null state overrides active-pane default` (test at ~line 322, gate at ~line 353)

## Evidence

Observed during the review of #64/#67 on branches that touch **zero** files under `src/tui/`, so this is pre-existing and unrelated to those changes:

- One full-suite run: 1 fail; three immediately following runs: 0 fail.
- Reproduced independently at 78 pass / 1 fail on two of three isolated runs, then green 6/6 in isolation, 2/2 on the full suite, and 4/4 under deliberate CPU contention.

The load-dependence is the tell: a 10ms sleep is sometimes enough for the hydration effect to flush and sometimes not, depending on what else the runner is doing.

## Why it matters

It makes merge gating noisy. A green suite is supposed to be a signal, and an intermittent failure that no one can reproduce on demand trains people to re-run rather than investigate, which is how a real regression eventually gets waved through.

## Proposed fix

Await the hydration signal itself rather than sleeping. Options, roughly in order of preference:

1. Await the actual condition (poll the store/renderer state until hydration has applied, with a generous timeout), so the test is fast when the machine is idle and still correct when it is loaded.
2. Expose a resolved promise or effect-flush hook from the hydration path that the test can await deterministically.

Bumping the sleep is not a fix; it just moves the failure threshold.

## Verification

Run `bun test src/tui/App.test.tsx` repeatedly under CPU contention (e.g. alongside a full `bun test`) and confirm both tests are stable across many consecutive runs.

epilande:
## Overview

Replaces the fixed 10ms sleeps in the two sidebar-hydration tests with a deterministic await of the hydration fetch chain itself.

Resolves #74

## Motivation

Both tests gated on `await new Promise((r) => setTimeout(r, 10))` as a stand-in for "the hydration effect has flushed". Under full-suite CPU load 10ms is sometimes not enough, so the tests fail intermittently while passing in isolation. That noise trains people to re-run the suite instead of investigating, which is how a real regression eventually gets waved through.

### Tests

- **src/tui/App.test.tsx** - The fetch mock now captures the promise it returns. Each test awaits that captured promise, then flushes two microtask turns so the chained `.then` callbacks (the `r.json()` parse, then the `applySidebarSelection` call) have run. App's `.then` callbacks attach to the same promise before the test's await, so ordering is guaranteed by microtask semantics rather than wall-clock time. No production code changed.

## Breaking Changes

None

## Testing

- [x] `bun test src/tui/App.test.tsx` green 10 consecutive runs in isolation, plus 5 more while a full `bun test` ran concurrently as CPU load
- [x] Full `bun test`: 3747 pass / 0 fail (verified independently across multiple runs)

One unrelated pre-existing flake surfaced once under load during verification: `App sidebar mode > debounces flash during rapid navigation` (App.test.tsx:887). It is one of the ~20 other `setTimeout`-based settles this issue explicitly scoped out, and this diff does not touch it. Flagging it here as a candidate follow-up.

### DIFF
--- a/src/tui/App.test.tsx
+++ b/src/tui/App.test.tsx
@@ -323,12 +323,24 @@ describe("App", () => {
 
   it("sidebar hydration with null state does not clobber active-pane default", async () => {
     const originalFetch = globalThis.fetch;
-    globalThis.fetch = (async () => ({
-      json: async () => ({
-        selectedSessionId: null,
-        selectedHeaderKey: null,
-      }),
-    })) as unknown as typeof fetch;
+    // Capture the promise our mock fetch returns so the test can await the
+    // hydration chain directly instead of guessing at a sleep duration. The
+    // fire-and-forget chain in App.tsx is
+    // fetch(...).then(r => r.json()).then(data => {...}).catch(...); App's
+    // .then callbacks are attached to this same promise before the test's own
+    // await, so once `hydrationFetchPromise` settles, App's `r.json()` call
+    // has already been made (see the extra microtask flush below).
+    let hydrationFetchPromise: Promise<unknown> | undefined;
+    globalThis.fetch = (() => {
+      const promise = (async () => ({
+        json: async () => ({
+          selectedSessionId: null,
+          selectedHeaderKey: null,
+        }),
+      }))();
+      hydrationFetchPromise = promise;
+      return promise;
+    }) as unknown as typeof fetch;
     try {
       await renderApp(120, 20, { sidebar: true, groupBy: "none" });
       sseCallbacks!.onInit(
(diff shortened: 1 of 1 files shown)


epilande:
Independent review pass (fresh-context reviewer): merge with nits, no Must-fix. All three Should-fixes applied in 484eaee: a `toBeDefined()` guard so the null-state test can't pass vacuously, the capture gated on the `/sidebar-state` URL (onMount fires a `/server-info` fetch first), and three microtask flushes with a comment stating the measured cold-start worst case. Both mock blocks collapsed into one helper. Re-verified: 134/134 stable across repeated runs, full suite 3747/0.

PR #88, opened 2026-07-31
fix(tui): Account for wrapped content in dialog and menu heights

### LINKED ISSUE #82: [Bug]: A context-menu label that wraps silently breaks the menu's height and clamping
## Summary

`ContextMenu` is a fixed 22 columns wide and computes its height as `items.length + 2`. A label long enough to wrap renders **two** rows while still counting as one item, so the box height and the viewport clamping are both wrong for the whole menu — silently, with no error.

Found while adding a menu item during #71: the intended label "Move changes to worktree" wrapped in a live capture. It was shortened to "Move changes" to avoid the bug, so nothing ships broken today, but the trap remains for the next long label.

## Why this is worth hardening

This is the same class as the dialog-height bug that `fieldRows` was introduced to catch in #78: a hand-maintained size that type-checks fine and only misbehaves at render time. That one was fixed by deriving the height so a missing entry fails compilation. The menu has the same shape of problem and no such guard.

It is also invisible to the existing tests, which assert menu *content*. A wrapped label still contains its text, so every assertion passes while the box is drawn wrong — the same blind spot that let a broken dialog border through in #69 until a live capture caught it.

## Suggested directions

Any of these would close it; the first is cheapest:

1. Derive the height from the rendered row count rather than `items.length`, so a wrapping label is accounted for.
2. Truncate labels to the available width with an ellipsis, so a long label degrades legibly instead of silently changing the layout.
3. Assert at render time (or in a test helper) that every label fits `MENU_WIDTH`, so an over-long label fails loudly during development.

Option 3 alone would have caught this one at authoring time.

## Verification

- Add a menu item whose label exceeds the available width and confirm the box height matches the rendered rows.
- Add a frame-integrity assertion in the menu tests — every box row ends with the border character — mirroring the check added for the new-session dialog in #69, which catches this class rather than one instance.

### LINKED ISSUE #85: [Bug]: New-session dialog clips its bottom rows when the Agent field shows a wrapped error
## Summary

The new-session dialog computes its height from `fieldRows`, which budgets the Agent field as `max(1, visibleAgents().length)`. That accounts for the agent *list* but not for an **error message rendered in that field**, which can wrap to several rows. The dialog then comes up short and its last rows fall outside the border.

Pre-existing — reproduced on `main`, not introduced by #69.

## Reproduction

Run the picker against a daemon built from an older revision, so the Agent field renders the stale-daemon error (`Daemon is out of date - run \`ccmux daemon restart\``). At a narrow width that message wraps to three rows while the field is budgeted for one, so the height is two rows short.

Observed:

- On `main` at ~34 columns: the third placement option (`3 Split down`) is missing.
- On the #69 branch, the same clipping removes the **`New worktree` destination option** entirely — a functional loss rather than a cosmetic one, since the field's whole purpose disappears.

## Why the tests do not catch it

`NewSessionDialog.test.tsx` renders with a healthy agent list, so the error branch is never on screen when the height is exercised. The height itself is derived correctly from the field list — `fieldRows` is honest about the *fields*; nothing accounts for a field's content wrapping beyond its budgeted rows.

This is the same blind spot that let a broken right border through in #69 until a live capture caught it, and it is the reason that PR added a frame-integrity assertion (every box row ends with `│`). That assertion would likely catch this too if the error state were rendered.

## Suggested direction

- Have `fieldRows.agent` account for the rendered height of an error message, not just the agent count — i.e. measure the wrapped line count at the current width rather than assuming one row.
- Add a component test that renders the dialog in the agents-error state at a narrow width and asserts, via the existing frame-integrity check, that no row is clipped.

## Verification

Render the dialog with `agentsError` set at the widths the responsive branches target (roughly 34, 63, full) and confirm every field row and every option is present and inside the border.

epilande:
## Overview

Fixes the two wrapped-content sizing bugs in the TUI: the new-session dialog clipping its bottom rows when the Agent field shows a wrapped error, and the context menu's height and viewport clamp silently breaking when a label wraps.

Resolves #85
Resolves #82

## Motivation

Both surfaces sized themselves from a row count that assumed their content fit on one row. At 34 columns the stale-daemon error left the dialog three rows short, so `3 Split down` and the entire `Where` field (the worktree destination) fell outside the border. The menu's `items.length + 2` fed the viewport clamp while a wrapping label rendered two rows, pushing the bottom border off screen near the bottom edge. Both failed silently: content assertions pass because the text is all there, so only live captures caught them.

### New-session dialog (#85)

- **NewSessionDialog.tsx** - Adds `wrapText()` (greedy word wrap) and pre-wraps the agent error to the content width, one pinned row per line, capped at the rows the field actually has (with an ellipsis when cut). `fieldRows.agent` budgets from that same line count, so the height is derived from what is drawn rather than estimated. `Loading agents...` now goes through `truncateText` like every other string in the dialog.
- **NewSessionDialog.test.tsx** - Error state at 34 columns asserting frame integrity plus every row below the error (`Split down`, `Where`, `Worktree`, `Directory`); error-taller-than-screen cap; `wrapText` unit tests.

> [!NOTE]
> The issue suggested estimating wrapped rows with `Math.ceil(len / width)`. That was probed against opentui's native wrapping over a 121-case corpus and no char-count formula was a reliable upper bound (it word-wraps, carries boundary spaces to the next row, and sometimes hard-breaks a word early). An under-prediction is exactly the clip being fixed, so the dialog wraps the text itself: pre-fitted lines cannot wrap again, making the budget honest by construction.

### Context menu (#82)

- **ContextMenu.tsx** - Labels truncate to the space left beside their key hint, item rows are pinned to one row, and the outer box now sets `height={menuHeight()}` so the clamp's input matches the real box. No label authored today changes; a test pins the four longest real label/hint pairs and asserts none gain an ellipsis.
- **ContextMenu.test.tsx** - Net-new border/clamp coverage: over-long label keeps `items.length + 2` height with a legible ellipsis, and a bottom-edge placement keeps both borders on screen.

### Shared test machinery

- **test-helpers.tsx** - `expectFrameIntegrity()` (every box row closes with `│`), replacing two inline copies and guarding this class of bug for future components.

## Breaking Changes

None

## Testing

- [x] Full `bun test`: 3756 pass / 0 fail; `bun run typecheck` clean (verified independently)
- [x] Pre-fix repro confirmed for both: with the fix stashed, the new dialog assertions catch the missing `Split down` and `Where` rows at 34 columns, and the menu test catches the collided label and lost bottom border
- [x] Live capture of the error state via a stub daemon 404ing `/agents`: the wrapped stale-daemon message renders across 4 rows at 34 columns with every field present and the border intact (also clean at 30 and 40 columns)
- [x] Live capture of the context menu via injected SGR right-click: full labels, bottom-edge clamp flush with the last row
- [x] Live picker capture of the healthy dialog at 200 and 60 columns from this branch's build

One pre-existing cosmetic issue observed while verifying, untouched here: at 30 columns the `Where` row's compact option label clips as `Worktre` because `optionLabel` has no truncation of its own. The border stays intact; candidate follow-up.

### DIFF
--- a/src/tui/components/ContextMenu.test.tsx
+++ b/src/tui/components/ContextMenu.test.tsx
@@ -139,9 +139,8 @@ describe("ContextMenu sizing", () => {
     const items = itemSpies(["Attach", LONG, "Kill"]);
     const { frame } = await renderMenu({ items });
 
-    expectFrameIntegrity(frame);
-    const { height, lines, top } = boxBounds(frame);
-    expect(height).toBe(items.length + 2);
+    expectFrameIntegrity(frame, items.length + 2);
+    const { lines, top } = boxBounds(frame);
     // Legibly cut rather than merely clipped, and still beside its hint.
     const row = lines[top + 2]!;
     expect(row).toContain("Move changes to");
@@ -160,7 +159,7 @@ describe("ContextMenu sizing", () => {
     ];
     const { frame } = await renderMenu({ items });
 
-    expectFrameIntegrity(frame);
+    expectFrameIntegrity(frame, items.length + 2);
     for (const item of items) expect(frame).toContain(item.label);
     expect(frame).not.toContain("…");
   });
@@ -172,10 +171,12 @@ describe("ContextMenu sizing", () => {
     const size = { width: 40, height: 10 };
     const { frame } = await renderMenu({ items, x: 9999, y: 9999, size });
 
-    expectFrameIntegrity(frame);
-    const { top, bottom, height } = boxBounds(frame);
+    // The height assertion is the one that fails pre-fix here: a box taller
+    // than `menuHeight()` loses its bottom border off the viewport, so there
+    // is no `└` to measure to.
+    expectFrameIntegrity(frame, items.length + 2);
+    const { top, bottom } = boxBounds(frame);
     expect(top).toBeGreaterThanOrEqual(0);
-    expect(height).toBe(items.length + 2);
     // Flush against the bottom row of the viewport, not past it.
     expect(bottom).toBe(size.height - 1);
     // Every item is still on screen, first and last included.
--- a/src/tui/components/ContextMenu.test.tsx
+++ b/src/tui/components/ContextMenu.test.tsx
@@ -2,6 +2,7 @@ import { describe, it, expect, afterEach, mock } from "bun:test";
 import { testRender } from "@opentui/solid";
 import { MouseButtons } from "@opentui/core/testing";
 import { ContextMenu, type ContextMenuItem } from "./ContextMenu";
+import { expectFrameIntegrity } from "./test-helpers";
 import { theme } from "../theme";
 
 type Setup = Awaited<ReturnType<typeof testRender>>;
@@ -48,6 +49,14 @@ async function renderMenu(
   );
   await setup.renderOnce();
   return { frame: setup.captureCharFrame(), items, onClose };
+}
+
+/** The box's own rows, border included, from a captured frame. */
+function boxBounds(frame: string) {
+  const lines = frame.split("\n");
+  const top = lines.findIndex((l) => l.includes("┌"));
+  const bottom = lines.findIndex((l) => l.includes("└"));
+  return { top, bottom, height: bottom - top + 1, lines };
 }
 
 function locate(frame: string, label: string) {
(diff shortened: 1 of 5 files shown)


epilande:
Independent review pass: merge with nits, no Must-fix. The reviewer re-proved both fixes by reverting the components in a scratch copy (all four new tests fail pre-fix). Nits applied in 2159de4: `expectFrameIntegrity`'s docstring now states what it actually guards (it caught zero of the four bugs here; the height/content assertions do the real work) and gained an optional `expectedHeight` asserted from the corner rows; the empty-string `agentsError` now falls back via `||` instead of rendering a blank red row. The reviewer's deeper finding, code-unit vs display-column measuring shared by `wrapText` and `truncateText`, is tracked in #91 rather than patched here. Re-verified: full suite 3756/0.

PR #89, opened 2026-07-31
fix(prune): Name ignored directories in the prune run log

### LINKED ISSUE #81: [Bug]: Prune deletes gitignored directories silently, with no mention in the run log
## Summary

The prune run log names ignored **files** it is about to delete, but says nothing about ignored **directories** — which are deleted with no opt-in and no mention anywhere the user can see.

Found while verifying #80. Correctly attributed there as pre-existing #68 behavior rather than a regression, but worth its own issue.

## Current behavior

`readDirtyState` classifies ignored entries in two ways:

- An ignored **file** (e.g. `.env`) is surfaced: it appears on the candidate row, in both confirmation steps, and in the run log as `deleting ignored: 1 ignored file (.env)`.
- An ignored **directory** (git reports it with a trailing slash) is dropped entirely as regenerable build output, per the deliberate policy at `docs/architecture.md:205`.

The directory case is right for `node_modules/` and `dist/`, which is what the policy was written for. It is wrong for a gitignored directory holding work — a local scratch directory, a data dir, an untracked experiment — which is deleted with no warning and no line in the log saying it happened.

## Why the file/directory split does not hold

The distinction the policy actually wants is "regenerable" versus "not", and the file/directory boundary is only a proxy for it. `node_modules/` is regenerable; `notes/` is not, and both are directories. A user cannot tell from the confirmation which of the two they are about to lose, because neither is mentioned.

## Concrete case

A worktree containing a gitignored directory of work, pruned with no dirty opt-in:

- pre-#80 and post-#80 behave identically — the directory is deleted either way
- `ignoredFiles` is empty, because the trailing slash filters directories out
- the run log's `deleting ignored` line never appears
- nothing on the row, in the confirm, or in the log mentions it

## Suggested direction

Not necessarily to gate on it — that would fire on nearly every worktree and train reflex approval, which is the alarm-fatigue reasoning the current policy is built on. Options worth weighing:

1. Name ignored directories in the run log at least, so the deletion is recorded even if it is not gated.
2. Distinguish by name: exempt a configured `worktree.symlinkDirectories` entry and a small known-regenerable set, surface the rest.
3. Surface a count on the row (`+2 ignored dirs`) without adding an opt-in.

Whatever the choice, the user should be able to see it happened.

## Verification

- Confirm an ignored directory of work is deleted today with no mention on the row, in either confirmation, or in the run log.
- Confirm the chosen surfacing does not fire on a plain `node_modules/` in a way that produces the reflex-approval problem the current policy avoids.

epilande:
## Overview

Names gitignored directories in the prune run log when a removal deletes them, closing the gap where they were deleted with no record anywhere the user can see.

Resolves #81

## Motivation

`readDirtyState` dropped every trailing-slash entry from `git status --ignored=matching` as presumed-regenerable build output. That is right for `node_modules/` and wrong for a gitignored `notes/` holding real work, which prune deleted with no mention on the row, in either confirmation, or in the run log.

This is log-only on purpose. Gating on an ignored directory would fire on the `node_modules/` in essentially every worktree and train the reflex approval the ignored-file policy exists to avoid, so the dirty gate and both confirmation steps are untouched. The deletion is now recorded somewhere the user can read afterwards, which is the whole of what was missing.

### Read half

- **worktree-git.ts** - `DirtyState` gains `ignoredDirs`; `readDirtyState` routes `!!` entries by git's trailing slash instead of discarding the slashed ones. The slash (not an `isDirectory()` stat) is the discriminator because it is reliable in the direction that matters: git prints it for a real directory and never for a symlink to one, so a `worktree.symlinkDirectories` setup link (the #80 case) can never be miscounted as a directory of work.

### Run log

- **worktree-prune.ts** - `PruneCandidate` carries `ignoredDirs`; the existing summary formatting is factored into `summarizeIgnored()` so `describeIgnoredFiles` keeps byte-identical output on the row and confirmations, and a new `describeIgnoredDeletion(files, dirs)` feeds both log emission sites (dry run and real run). Files come first in the line: they are the ones no backup can bring back.

  ```
  deleting ignored: 1 ignored file (.env), 2 ignored dirs (notes/, data/)
  would delete ignored: 1 ignored dir (notes/)
  ```

### Tests and docs

- **worktree-prune.test.ts** - 13 new tests: collection, log emission in both modes, gating unchanged (an ignored directory still does not flip `dirty`, still needs no opt-in, and removal proceeds cleanly with no `allowDirtyPaths`), and the #80 symlink cases pinned both ways (directory-pattern symlink produces no ignored entries at all; bare-name-pattern symlink stays an ignored FILE, verified against real git behavior).
- **PruneDialog.test.tsx** - `ignoredDirs: []` in the candidate factory, plus format-hook rewraps.
- **docs/architecture.md** - the "Ignored files are shown, not gated" paragraph now documents the run-log-only surfacing instead of claiming directories are dropped.

## Breaking Changes

None

## Testing

- [x] Full `bun test`: 3759 pass / 0 fail; `bun run typecheck` clean (verified independently)
- [x] Gating unchanged, proven by test: merged worktree with a gitignored `notes/plan.md` reads `dirty === false` and prunes with no dirty opt-in, while the run log names `notes/`
- [x] No noise regression: a plain `node_modules/` still appears nowhere on the row or confirmations, only in the run log of an actual removal


epilande:
Independent review pass: merge with nits, no Must-fix; the reviewer verified the core claims against real git (byte-identical file summaries, gating unchanged, both #80 symlink pins). Both Should-fixes applied in e1a076a: C-quoted directories (`!! "notes dir/"`, non-ASCII names) are now classified as directories via `isCollapsedDirectory` (a bare `endsWith("/")` filed them as ignored FILES, putting them on the confirmation surfaces this PR keeps directory-free), and the run log emits one step per kind so sidebar-width truncation can't cut away exactly the directory names. Re-verified: full suite 3761/0.

PR #90, opened 2026-07-31
fix(spawn): Switch the client for cross-session --target

### LINKED ISSUE #75: [Bug]: Cross-session spawn --target reports success but never switches the client
## Summary

`POST /spawn` (and `ccmux spawn --target <pane>`) accepts any pane on the tmux server. When the target pane lives in a **different session** from the caller and `--detach` is not passed, the daemon creates the pane and reports success, but the user is never switched to it.

The final step uses `select-window -t <newPaneId>`, which sets the active window *within* the target session. Moving an attached client between sessions requires `switch-client` instead.

Verified during the review of #64: running `select-window` against a pane in another session changed that session's active window while the attached client stayed put.

## Why it was deferred

Found during #64 (spawn split direction / target pane / per-agent prompts) and deliberately left unfixed there. A proper fix needs client-TTY resolution inside the daemon: the daemon is not attached to the caller's client, which is the same problem `src/commands/switch.ts` solves by passing `-c <tty>`. That is more new surface than a Low-severity issue warranted inside a batch that was otherwise security fixes, so #64 documented the limitation instead.

Newly reachable only because `--target` now accepts an arbitrary pane; before that, spawns always landed in the daemon's own notion of "current".

## Current behavior

Documented in the README as a caveat: pair a cross-session `--target` with `--detach`.

## Proposed fix

Have the spawn path detect that the target pane's session differs from the caller's, and switch the caller's client with `switch-client -t <target>` using the caller's TTY, mirroring what `switch.ts` already does. The CLI already has the caller's context; the missing piece is threading the client TTY through the spawn request so the daemon can act on the right client.

## Verification

- Attach a client to session A, spawn with `--target` pointing at a pane in session B without `--detach`, and confirm the client actually lands on the new pane.
- Confirm the same-session case is unchanged.
- Confirm `--detach` still suppresses any switching.

epilande:
## Overview

Makes a cross-session `ccmux spawn --target <pane>` actually move the caller's client to the new pane, instead of reporting success while the client stays put.

Resolves #75

## Motivation

`--target` accepts any pane on the server, including one in a different tmux session. The spawn created the pane there, but the final `select-window` only changes which window is current within that session; moving an attached client between sessions needs `switch-client`, and the daemon has no client of its own to run it on. The README documented "pair it with `--detach`" as a caveat; this removes the caveat.

### TTY threading (CLI → daemon)

- **spawn.ts** - `callerClientTty()` resolves `#{client_tty}` (the attached client's device, not the pane's own pty) and sends it as `callerTty` on `POST /spawn`. Resolved only when an explicit `--target` was given without `--detach`, behind the same cross-server gate `callerPane` uses, so an ordinary spawn makes zero extra tmux or HTTP calls. Also hoists the single `/server-info` round-trip both placement fields share.
- **tmux-client.ts** - New `resolveCurrentTmuxClientTty()`.

### Decision seam (daemon)

- **spawn-command.ts** - `normalizeClientTty()` validates the wire field against a strict `/dev/...` device-path pattern (400 otherwise; the value travels straight into a tmux argv). `buildSpawnFocusArgv()` is the pure decision: `null` on detach, `switch-client -c <tty> -t <pane>` when the placement session provably differs from the caller's, `select-window` for everything else, including anything unprovable, so a missing tty or a vanished caller pane falls back to today's behavior rather than moving a client that never asked.
- **server.ts** - Captures the placement and caller session ids from the pane probes (a second probe runs only when a target and a tty are both present) and drives the focus step through `buildSpawnFocusArgv`.

### Docs

- **README.md** - The cross-session caveat now documents the switch, with `--detach` as the opt-out.

## Breaking Changes

None. Requests without `callerTty` (older CLIs, the picker, fork) take the exact pre-change path.

## Testing

- [x] Full `bun test`: 3764 pass / 0 fail (+26 tests); `bun run typecheck` clean (verified independently)
- [x] Unit coverage for the decision seam (same-session / cross-session / detach / each unprovable fallback), the tty validator, the daemon route argv for all cases including bad-tty 400s, and the CLI threading gates
- [x] Live e2e on an isolated tmux server (`-L ccmux75`, own daemon on port 22690, scoping proven before start) with a real attached client and a tmux argv recorder, so every case is verified by both client movement and exact command:
  - Cross-session `--target`: `switch-client -c /dev/ttys085 -t %2`, client followed A → B (new-window and `--split h` both)
  - Same-session `--target`: byte-identical old path, single probe, no `switch-client`
  - `--detach`: no tty lookup, no focus command, client stayed on A while the pane appeared in B
  - Ordinary spawn: identical tmux call sequence to pre-change
  - Pre-fix behavior reproduced by hand on the same server (`select-window` leaves the client on A)

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,10 @@ the end of your session, which leaves every existing window index alone; pass
 `--target` accepts any pane on the server, including one in a _different_ tmux
 session: ccmux creates the pane there and moves your client over to it, the
 same jump `ccmux switch` makes. Pass `--detach` to leave your view where it is.
+That jump needs a client of its own to move, so it only happens when you run
+`ccmux spawn` from inside the session you are attached to; run from outside
+tmux, or from a detached session, it creates the pane without switching (use
+`ccmux switch` afterwards).
 
 `--prompt` starts the agent interactively with the prompt already submitted.
 It is supported for the agents whose interactive-with-prompt invocation ccmux
--- a/README.md
+++ b/README.md
@@ -284,9 +284,9 @@ the end of your session, which leaves every existing window index alone; pass
 `--target <pane-id>` to insert one directly after that pane's window instead
 (tmux renumbers the windows after it), or `--target none` to let tmux place it.
 
-Targeting a pane in a _different_ tmux session creates the pane there but does
-not move you to it, so pair that with `--detach`
-(see [#75](https://github.com/epilande/ccmux/issues/75)).
+`--target` accepts any pane on the server, including one in a _different_ tmux
+session: ccmux creates the pane there and moves your client over to it, the
+same jump `ccmux switch` makes. Pass `--detach` to leave your view where it is.
 
 `--prompt` starts the agent interactively with the prompt already submitted.
 It is supported for the agents whose interactive-with-prompt invocation ccmux
--- a/src/commands/spawn.test.ts
+++ b/src/commands/spawn.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, mock, afterEach } from "bun:test";
+import { describe, it, expect, mock, spyOn, afterEach } from "bun:test";
 
 // Neutralize ensureDaemon so the action never spawns/probes a real daemon.
 // Counted as well as neutralized: argument validation that runs BEFORE it is
@@ -23,6 +23,7 @@ interface SpawnBody {
   split: unknown;
   target?: string;
   callerPane?: string;
+  callerTty?: string;
   detach: boolean;
   worktree?: { name?: string; base?: string };
 }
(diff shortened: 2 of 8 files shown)


epilande:
Independent review pass found one Must-fix, verified live: `#{client_tty}` resolved from a pane in a session with no attached client falls back to the most-recently-active client of any session (tmux `cmd_find_best_client`), so a spawn from a detached session could switch a client in an unrelated session — the session comparison couldn't see it because it compares panes, not the client. Fixed in 78a464e: the daemon now verifies the tty is a client of the caller's session (`list-clients -t`) before ever emitting `switch-client`, falling back to `select-window` otherwise; a failed focus command is now logged instead of silently reporting success. E2E re-run on the isolated rig: the bystander repro no longer moves the client while the legit cross-session switch still works, and the same-session path is unchanged. Re-verified: full suite 3772/0.

PR #92, opened 2026-07-31
fix(tui): Measure text in display columns, not code units

### LINKED ISSUE #91: [Bug]: TUI text utilities measure code units, not display columns
## Summary

The TUI's text-measuring utilities count UTF-16 code units, not display columns, so any string containing wide glyphs (CJK) is mis-measured at exactly the places that exist to make text fit: `truncateText` (src/tui/utils/format.ts:33, pre-existing) and `wrapText` (src/tui/components/NewSessionDialog.tsx, added in #88).

Found during the review of #88. Not a blocker there: every string those call sites receive today is ASCII in practice, and emoji happen to survive (2 code units and 2 columns). But the failure mode is ugly when hit.

## Reproduction

Render the new-session dialog at 34 columns with a Japanese `agentsError` (reachable via `Failed to resolve agents: ${errorMessage(err)}`, which can embed a filesystem path or user-config value):

```
wrapText (code units)          renderer wrapping (pre-#88)
│ >Agent  エージェントを  │    │ >Agent  エージェントを  │
│         した            │    │         解決できません  │
│         デーモンを再起  │    │         でした デーモン │
```

A "15-column" line is really 30 columns, the renderer clips half of it, and the message reads as gibberish rather than as a truncation. `truncateText` has the same assumption everywhere it is used (columns, labels, subtitles), it just degrades less visibly.

## Suggested direction

One shared display-width helper (wide ranges count 2), used by both `truncateText` and `wrapText`, rather than per-call-site patches. If @opentui/core exports a string-width utility (its renderer clearly measures wide glyphs correctly), reusing that beats maintaining our own East Asian Width table.

## Verification

- `wrapText`/`truncateText` unit tests with CJK and emoji inputs asserting column-true output.
- Re-render the dialog error case above and confirm a contiguous, legible prefix.

epilande:
## Overview

Adds shared display-width helpers (`displayWidth`, `sliceToWidth`) and rebuilds the TUI's text-measuring utilities on top of them, so truncation and wrapping count terminal columns instead of UTF-16 code units.

Resolves #91

## Motivation

Prompts, transcript snippets, and error messages flow through `truncateText`, `truncateHighlighted`, and `wrapText`, and all three measured code units. A CJK string over-fills its cell (each glyph is 2 columns but counted as 1), a ZWJ emoji like a family burns 11 units of budget for 2 visible columns, and a naive `slice()` can cut mid-surrogate and render a replacement glyph. The new-session dialog repro in #91 turned a Japanese error message into half-clipped gibberish.

### Shared helpers

- **src/tui/utils/format.ts** - Adds `displayWidth` (built on `Bun.stringWidth` with `ambiguousIsNarrow: true`, verified against OpenTUI's renderer by probing ambiguous glyphs like `…` and `▎` in a fixed-width box) and `sliceToWidth` (grapheme-safe prefix via `Intl.Segmenter`, never splits a cluster). `truncateText` and `truncateHighlighted` now measure and slice in columns; a private `sliceTailToWidth` handles the leading-context suffix case.

### Wrapping

- **src/tui/components/NewSessionDialog.tsx** - `wrapText` measures in columns and breaks over-wide words on grapheme boundaries, with a termination guard for a single cluster wider than the column (emits the word whole rather than spinning or splitting the cluster).

### Tests

- **src/tui/utils/format.test.ts** - Column-true assertions for CJK, ZWJ emoji, flags, and mixed strings; verifies no output ends in a lone surrogate.
- **src/tui/components/NewSessionDialog.test.tsx** - Wrapping invariants (`displayWidth(line) <= width` for every line) plus the issue's repro as a component test: the dialog at 34 columns with a Japanese `agentsError` renders a contiguous, legible message with the border intact. Five of the new tests fail against the pre-fix code.

Out of scope (deferred follow-up): the raw `.length` budget arithmetic in `session-columns.ts` and `SessionItem.tsx`.

## Breaking Changes

None

## Testing

- [x] Full suite: 3821 pass, 0 fail; typecheck clean (verified independently of the implementation run)
- [x] New tests proven to catch the bug (revert check: 5 fail on pre-fix code)
- [x] Live picker + new-session dialog rendered in a detached tmux session at 200x50 and 60x30 (borders, ellipsis truncation, active indicator intact)

### DIFF
--- a/src/tui/components/NewSessionDialog.test.tsx
+++ b/src/tui/components/NewSessionDialog.test.tsx
@@ -2,6 +2,7 @@ import { describe, it, expect, afterEach } from "bun:test";
 import { testRender } from "@opentui/solid";
 import { NewSessionDialog, optionWindow, wrapText } from "./NewSessionDialog";
 import { expectFrameIntegrity, squish } from "./test-helpers";
+import { displayWidth } from "../utils/format";
 import type { SpawnableAgent } from "../../lib/spawnable-agents";
 import type { NewSessionDraft } from "../store";
 
@@ -112,6 +113,45 @@ describe("wrapText", () => {
 
   it("gives up rather than looping when there is no width to wrap into", () => {
     expect(wrapText("anything", 0)).toEqual(["anything"]);
+  });
+
+  /**
+   * Issue #91: the widths above are display columns, so a line of CJK fills
+   * its column exactly like an ASCII one. Measured in code units these lines
+   * were twice their claimed width and the renderer clipped half of each.
+   */
+  it("wraps wide glyphs to the column they actually occupy", () => {
+    const message = "エージェントを 解決できません でした デーモンを 再起動";
+    const lines = wrapText(message, 19);
+    for (const line of lines) {
+      expect(displayWidth(line)).toBeLessThanOrEqual(19);
+    }
+    expect(lines.join(" ")).toBe(message);
+  });
+
+  it("breaks an over-wide CJK word on a glyph boundary", () => {
+    const lines = wrapText("解決できませんでした", 9);
+    for (const line of lines) expect(displayWidth(line)).toBeLessThanOrEqual(9);
+    expect(lines.join("")).toBe("解決できませんでした");
+    // An odd budget stops a column short rather than splitting a glyph.
+    expect(lines[0]).toBe("解決でき");
+  });
+
+  it("never splits an emoji cluster across lines", () => {
+    const family = "👨‍👩‍👧‍👦";
+    const lines = wrapText(family.repeat(6), 5);
+    for (const line of lines) {
+      expect(displayWidth(line)).toBeLessThanOrEqual(5);
+      // Whole families only, so nothing renders as a replacement glyph.
+      expect(line.replace(new RegExp(family, "gu"), "")).toBe("");
+    }
+    expect(lines.join("")).toBe(family.repeat(6));
+  });
+
+  it("emits an unsplittable glyph rather than spinning on it", () => {
+    // A column too narrow for one wide glyph: it has to overflow, but the
+    // loop must still terminate.
+    expect(wrapText("日本", 1)).toEqual(["日本"]);
   });
 });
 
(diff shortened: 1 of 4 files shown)


epilande:
Adversarial review pass complete. Verdict: merge as-is, no Must-fix or Should-fix findings.

Independent verification highlights:

- **Core premise confirmed**: `Bun.stringWidth` with `ambiguousIsNarrow: true` matches OpenTUI's native "unicode" measurer on all 16 glyph classes probed (CJK, hangul, halfwidth kana, ZWJ family, flags, skin tones, keycaps, VS16, combining marks, and the ambiguous `…` `▎` `α` `→` `①`), verified by rendering each inside a box sized to `displayWidth` with a sentinel column.
- **ASCII parity is exact**: old vs new implementations produce 0 differences over an exhaustive sweep (every prefix x every budget, every span position for `truncateHighlighted`, 4224 `fitProjectCell` configurations). Existing callers see byte-identical output.
- **Fuzzing clean**: 90,000 random inputs across mixed scripts; no content loss, no over-budget lines except the documented unsplittable-glyph case, no hangs.

Deferred Optional findings (recorded, not blocking):

1. Per-call cost is 200-600x the old code-unit math, but end to end a worst-case 40-session full-refresh render moves ~2% (~0.3ms). A 4-line ASCII fast path exists if this ever shows up in a profile.
2. The `wrapText` unsplittable-glyph guard emits the whole word (unbounded overflow) at widths no production path can reach (`contentWidth()` never drops below 9). One-cluster-per-line would cap overflow at one column if a future caller passes width < 2.
3. The ambiguous-width unit test passes even without the `WIDTH_OPTIONS` pin (Bun's default is already narrow), so it documents intent but does not defend the pin.
4. `session-columns.ts` now mixes code-unit gates with column-based truncation. Net effect on a 34k-configuration CJK sweep is strongly positive (7570 cells overflow less, 366 overflow 1-2 columns more inside already-overflowing cells; ASCII bit-identical). Tracked as a follow-up issue.

PR #94, opened 2026-07-31
fix(tui): Budget the project cell in columns, not code units

### LINKED ISSUE #93: [Bug]: Project-cell budget arithmetic still measures code units
## Summary

#92 made the shared text helpers column-true (`displayWidth`, `sliceToWidth`, and the rewritten `truncateText` / `truncateHighlighted` / `wrapText`), but deliberately left the raw `.length` budget arithmetic in the project-cell pipeline untouched. Those sites now mix metrics: `fitPrefix` (src/tui/components/session-columns.ts:612) gates on code units then truncates in columns, and `fitProjectCell` (session-columns.ts:677-726) sums `.length` of already-fitted parts against a column budget. Related sites: `branchLabelFor` (:578), `shortenBranchLabel` (:604), `fitDirname` (:630), `projectCellWidth` (SessionItem.tsx:216), the `padStart` right-align (SessionItem.tsx:316), and the `promptFloor` reserve (SessionItem.tsx:846).

## Impact

ASCII is bit-identical to pre-#92 behavior. For wide glyphs (CJK directory names, branch names) the mismatch causes silent column overspend that pushes cell content past its budget. A 34k-configuration CJK sweep during the #92 review measured the current mixed state as a net improvement over pre-#92 (7570 cells overflow less, 366 overflow 1-2 columns more, all inside cells that already overflowed by 1-7), so this is a polish item, not a regression.

Example regression case from the sweep: budget 8, dirname `プロジェクト`, branch `main`, worktree. Pre-#92 rendered `プロジェ…` at 9 columns (1 over); current renders `プロ…` plus `:ma~+` at 10 columns (2 over) because the shorter code-unit gate frees room for a branch label that used to be dropped.

## Suggested direction

Convert the pipeline's `.length` reads to `displayWidth` so the shrink logic and the renderer agree end to end. The fiddly part is that `fitProjectCell`'s shrink order is tuned to single-character margins, so each conversion wants a wide-glyph test case alongside (the existing assertions in session-columns.test.ts that sum `.length` of fitted parts would also need `displayWidth`). The `padStart` right-align needs a column-aware pad.

## Verification

- Wide-glyph cases in session-columns.test.ts asserting fitted cells never exceed their column budget.
- Re-run the CJK sweep shape from the #92 review and confirm the overflow-more bucket drops to zero.

epilande:
## Overview

Converts the project-cell layout pipeline from UTF-16 code-unit budgets to display-column budgets, so emoji and other wide glyphs in directory names, branch names, and prompts measure and align exactly like ASCII.

Resolves #93

## Motivation

#92 made the shared truncation helpers column-true, but the budget arithmetic feeding them still counted code units, so the two halves of the pipeline disagreed. A cell holding an emoji directory or branch name could overflow its column budget and push neighbouring cells out of alignment, and the right-aligned pane cell padded by code units, shoving an emoji session name right of its ASCII neighbours. This finishes the job: one metric end to end.

### Column budgets

- **src/tui/components/session-columns.ts** - All budget reads in `fitProjectCell` and its helpers (`branchLabelFor`, `shortenBranchLabel`, `fitPrefix`, `fitDirname`, `trailingLabelsWidth`) now measure with `displayWidth` and slice with `sliceToWidth`. The shrink order is untouched; only the measurement changed.
- **src/tui/components/SessionItem.tsx** - `projectCellWidth` sums fitted parts in columns, the `promptFloor` reserve measures the normalized prompt in columns, `alignText` pads via the new column-true helper, and the highlighted-branch gate now agrees with `branchLabelFor`'s cap (left in code units it would have let a wide highlighted branch render past it).
- **src/tui/utils/format.ts** - Adds `padStartWidth`, the column-true counterpart of `String.padStart`.

### Tests

- **src/tui/components/session-columns.test.ts** - Existing width assertions converted from `.length` sums to `displayWidth`; wide-glyph cases per helper (emoji dirname, ZWJ, CJK, emoji branch) plus a sweep-style property test asserting fitted cells never exceed their column budget and never emit a lone surrogate.
- **src/tui/components/SessionItem.test.tsx** - Right-align parity: a row with an emoji cell ends on the same column as an ASCII one.
- **src/tui/utils/format.test.ts** - `padStartWidth` coverage.

Sweep results (throwaway scripts, not committed): 73,200 ASCII configurations old vs new produced 0 differences, and a 109,800-configuration wide-glyph sweep went from 14,620 over-budget cells to 0.

Known limit: tmux itself measures ZWJ sequences (family emoji) wider than the 2 columns OpenTUI and `Bun.stringWidth` agree on, so a ZWJ directory name can still look ragged in a real terminal. That is a terminal-vs-renderer disagreement outside the layout's control; single-codepoint emoji and CJK render correctly.

## Breaking Changes

None

## Testing

- [x] Full suite: 3839 pass, 0 fail (+18 new tests); typecheck clean (verified independently of the implementation run)
- [x] ASCII parity: 73,200 configurations, 0 differences vs main
- [x] Wide-glyph sweep: 14,620 over-budget cells before, 0 after
- [x] Live picker frames (mock daemon rows, isolated tmux server): emoji and CJK cells truncate to exactly 12 columns and the pane cell right-aligns flush with ASCII rows

### DIFF
--- a/src/tui/components/SessionItem.test.tsx
+++ b/src/tui/components/SessionItem.test.tsx
@@ -5,6 +5,7 @@ import { createSignal } from "solid-js";
 import { SessionItem, alignText } from "./SessionItem";
 import { TickContext } from "../store";
 import { mockEnrichedSession } from "./test-helpers";
+import { displayWidth } from "../utils/format";
 import type { EnrichedSession } from "../../types";
 
 type Setup = Awaited<ReturnType<typeof testRender>>;
@@ -1144,6 +1145,15 @@ describe("alignText", () => {
     expect(alignText("", 3, "right")).toBe("   ");
     expect(alignText("", 3, "left")).toBe("");
   });
+
+  it("pads a wide glyph by the columns it draws", () => {
+    // `📚:1.0` is 7 code units but 6 columns, so a code-unit padStart would
+    // add 5 spaces and leave the cell a column left of its ASCII neighbours.
+    expect(alignText("📚:1.0", 12, "right")).toBe("      📚:1.0");
+    expect(displayWidth(alignText("📚:1.0", 12, "right"))).toBe(
+      displayWidth(alignText("dev:1.0", 12, "right")),
+    );
+  });
 });
 
 describe("SessionItem right-aligned fields", () => {
@@ -1158,5 +1168,49 @@ describe("SessionItem right-aligned fields", () => {
       160,
     );
     expect(frame).toContain("     dev:1.0");
+  });
+
+  // Both dirnames below are 13 characters; the CJK one draws 26 columns and
+  // has to give some back at this width, while the ASCII one fits whole.
+  // Under the code-unit budget both measured 13 and neither truncated, so the
+  // CJK row drew past its cell. One render per test so each gets torn down.
+  const dirnameRow = (dirname: string) =>
+    renderItem(
+      {
+        session: mockEnrichedSession({
+          cwd: `/Users/e/Code/${dirname}`,
+          paneCwd: `/Users/e/Code/${dirname}`,
+          gitBranch: "main",
+          tmuxTarget: "dev:1.0",
+          lastActivityAt: "2024-01-15T11:59:00Z",
+        }),
+      },
+      56,
+    );
+
+  it("leaves a 13-character ASCII dirname whole at 56 columns", async () => {
+    expect(await dirnameRow("abcdefghijklm")).toContain("abcdefghijklm:main");
+  });
+
+  it("truncates a 13-character CJK dirname at the same width", async () => {
+    expect(await dirnameRow("プロジェクトディレクトリ名前")).toContain(
+      "プロジェクトディレクトリ…:main",
+    );
+  });
+
+  it("right-aligns a wide-glyph pane cell to the same column as an ASCII one", async () => {
+    // Six columns of target in a 12-column cell: six spaces of pad. A
+    // code-unit pad counts the emoji's surrogate pair twice and stops at
+    // five, drawing the cell one column left of every ASCII row.
+    const frame = await renderItem(
+      {
+        session: mockEnrichedSession({
+          tmuxTarget: "📚:1.0",
+          lastActivityAt: "2024-01-15T11:59:00Z",
+        }),
+      },
+      160,
+    );
+    expect(frame).toContain("      📚:1.0");
   });
 });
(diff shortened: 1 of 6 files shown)


epilande:
Surface demo (shown, not driven): screen recording of the picker built from this branch, sped up, filtered to two throwaway sessions. What to look at: the `e2e94-📚-emoji-project-demo` row renders its wide-glyph project cell column-true next to the `e2e94-ascii-control` row, with the agent, version, pane, and age columns aligned across both rows and the prompt truncating at the shared right edge. Agent-agnostic layout change; no agent lifecycle involved.

Note: the video rasterizer cannot draw color emoji, so the book emoji appears as a placeholder box occupying its correct two columns; a real terminal renders the glyph itself (see the PR description's live captures).

https://github.com/user-attachments/assets/bb3cb836-73c0-42a2-ad63-e40b69579acf

epilande:
Adversarial review pass complete. Verdict: merge with fixes; all fixes applied in 6e8824b + 5d4eae5 and re-verified.

Independent verification highlights:

- **ASCII parity is exact**: 349,920 old-vs-new `fitProjectCell` configurations produced 0 differences, so plain-ASCII rendering is byte-identical to main.
- **The wide-glyph invariant is absolute**: across 829,440 emoji/ZWJ/CJK configurations at production-reachable budgets, the converted pipeline never exceeds its column budget; base overflowed 107,146 of them.
- **Completeness audit**: every remaining `.length` in both files classified as non-layout or provably ASCII, except one genuinely missed site (fixed below).

Should-fixes found and applied:

1. `shortenBranchLabel` could emit a content-free `:~`/`:~+` label when no grapheme fit the single body column (216 of 17,280 wide-glyph configs). It now drops the label entirely; a sweep confirms 0 remaining `:~` shapes and 0 new overflow.
2. The wide-glyph reserve test was vacuous (its `🚀` fixture measures 11 in both metrics). Swapped to `Bash(検索)` (8 code units, 10 columns) and revert-checked: reverting the conversion now fails the test.
3. `abbreviateTarget` still sliced code units, producing a lone-surrogate replacement glyph for emoji pane names and a 16-column render in the 12-column box for CJK ones. Converted to column-true slicing; emoji/CJK/ZWJ/ASCII shapes pinned as tests, 2,000-config ASCII sweep shows 0 differences.

Also added revert-catching tests for the two load-bearing conversions the coverage audit flagged as untested (`projectCellWidth` and the prompt-floor reserve); both were confirmed to fail against a revert and pass against the fix.

Deferred (recorded, not blocking): a tiny-budget (0-2) behavior difference in the repoPrefix arm unreachable behind the 12-column floor, and a microbenchmark-only perf delta (~0.18ms worst case per 60-row frame, both call sites memoized).

Final state: 3847 pass / 0 fail (+8 tests), typecheck and knip clean.

PR #96, opened 2026-07-31
feat(worktree): Move uncommitted changes into a fresh worktree

### LINKED ISSUE #71: [Feature]: Move uncommitted changes into a new worktree
## Overview

Adds Move changes: relocate a checkout's uncommitted changes into a fresh worktree and continue the work there, leaving the original checkout clean. Available from the session context menu ("Move changes to worktree") and the CLI (`ccmux spawn --worktree --with-changes`).

Depends on #69 (worktree spawning).

## Motivation

Work started in the main checkout has a way of outgrowing a quick edit: an experiment turns into a real change sitting dirty on `main`. Getting it onto its own branch in its own worktree is a fiddly manual sequence (stash, worktree add, pop, verify, reset), and fumbling it risks the changes. The same applies to a worktree whose session went one direction while the uncommitted work belongs in another.

### Mechanics

- Stash the source checkout's changes, create the worktree through the spawn creation engine, `stash pop` inside it, then reset the source checkout clean.
- Untracked files move by default (agents create new files constantly; excluding them would silently leave work behind), with a per-run choice: `move` (default), `copy` (new worktree gets them, source keeps them), or `leave` (they stay in the source only). Gitignored files are not part of this; the engine's file setup (symlinks/includes) covers those.
- The source is only reset after the pop succeeds. On pop failure or conflict, remove the created worktree and keep the stash intact, reporting the stash ref, so no state is ever lost.
- Refuse to run when the source checkout has an operation in progress (merge, rebase, cherry-pick).

### Entry points

- Session row context menu: "Move changes to worktree", shown only when the row's checkout is dirty. Opens the new session dialog (#65) prefilled: agent from the row, destination locked to new worktree, changes-move enabled, plus an "Untracked files" field (move / copy / leave, default move). Name and prompt stay editable before confirming.
- CLI: `ccmux spawn <agent> --worktree [name] --with-changes [--untracked <move|copy|leave>]`, run from or targeting the dirty checkout.
- Naming follows the spawn rules: explicit name, else deterministic slug from the prompt; no name and no prompt is an error.

### After the move

- The new worktree gets a fresh agent session like any other spawn; the source session and its checkout continue untouched, now clean.

## Out of scope

- Partial selection of what moves (hunk or file level). All uncommitted changes move together in v1.
- Moving committed-but-unpushed work; that is branch surgery, not a spawn concern.
- Automatic suggestions ("this checkout is dirty, move the changes?"). Manual invocation only.

## Verification

- Unit tests: stash/create/pop/reset sequencing, each untracked mode (move, copy, leave), conflict rollback preserving the stash, in-progress-operation refusal.
- Live e2e per the AGENTS.md workflow: dirty a checkout with tracked and untracked changes, run the action from the menu and the CLI, verify the changes land in the new worktree, the source is clean, the agent starts, and the row groups correctly. Separately force a pop conflict and verify rollback leaves the stash recoverable and the source untouched.

### LINKED ISSUE #83: [Feature]: Let the new-session dialog show and edit the derived worktree name
Follow-up to #69.

## Current behavior

The new-session dialog's **Where** field previews the worktree name derived from the prompt, live as you type. Two limitations, both deliberate at the time:

1. **The name is preview-only.** It cannot be edited in the dialog. An explicit name is available from the CLI via `ccmux spawn --worktree <name>`.
2. **At the dialog's 63-column width the preview truncates**, e.g. `fix-sidebar-…`, so the full slug is not readable before pressing Enter.

## Why they were deferred

Editing needs a second focusable text input inside a single field, which is a change to the field-list model (`NEW_SESSION_FIELDS`) that #69 was explicitly scoped not to restructure. The truncation fix is a layout change that was not worth rushing as the last act of that issue.

Neither is harmful today: the slug is deterministic from the prompt the user just typed, the created worktree's name is visible immediately afterward, and an explicit name is one CLI flag away.

## Why it is still worth doing

Committing to a name you cannot fully read is a small papercut that will be hit most by exactly the users the feature targets — someone typing a long prompt and letting the name derive itself. And "derived, take it or use the CLI" is a slightly awkward seam once the dialog is otherwise the primary way to spawn.

## Suggested directions

- Render the derived name on its own row rather than inline, so the full slug fits at typical widths.
- Or keep it inline but truncate from the middle, so both ends stay readable.
- If editing is wanted, treat the field-list model change as the substance of the work rather than a detail — that list drives focus order, option-key routing, rendered rows, and (since #78) the derived dialog height.

## Verification

Render the dialog at the widths the responsive branches target (roughly 34, 63, and full) with a long prompt, and confirm the full derived name is legible or degrades in a way that keeps both ends readable.

epilande:
## Overview

Adds Move changes: relocate a checkout's uncommitted work into a fresh worktree and continue there, leaving the original clean. Also makes the new-session dialog's derived worktree name legible and editable.

Resolves #71
Resolves #83

## Motivation

Work started in one checkout has a way of outgrowing it: an experiment turns into a real change sitting dirty on `main`. Getting it onto its own branch in its own worktree is a fiddly manual sequence (stash, worktree add, apply, verify), and fumbling it risks the changes. And once the dialog is the primary way to spawn worktrees, committing to a derived name you cannot fully read or change is a papercut that hits exactly the users the feature targets.

### Move engine

- **worktree-move-changes.ts** - The transaction: stash, create (through an injected seam backed by the spawn handler), apply into the new checkout, then drop. Apply-then-drop with by-SHA references keeps the work reachable on every failure path; there is deliberately no reset of the source. The whole transaction is locked per repository (worktrees share one stash stack), a stash entry only counts as ours if `refs/stash` moved, and `--index` preserves the staged/unstaged split when the target index allows it.
- **worktree-create.ts** - The seam reports whether the worktree was created or opened, and creating a worktree now writes `**/.claude/worktrees/` to the host repo's `.git/info/exclude` (the same entry Claude Code writes), so sibling checkouts are never treated as movable work.

### Wire + CLI

- **spawn-command.ts / server.ts** - The spawn request's `worktree` object gains `withChanges` and `untracked` ("move" | "copy" | "leave"), so changes-without-a-worktree is unspellable. The move runs before any pane exists; every failure body carries `reason`, `stashSha`, and `sourceRestored`, and a failure after a completed move still reports where the work went. `GET /sessions/:id/dirty` powers the menu gate and answers about the directory the move will actually use.
- **spawn.ts** - `ccmux spawn <agent> --worktree [name] --with-changes [--untracked <mode>]`, validated before the daemon starts. Prints counts, the real source path, and stash recovery advice that git actually accepts; refuses loudly when the running daemon predates the feature instead of silently spawning without the move.

### Picker

- **App.tsx / store.ts** - "Move changes to worktree" on dirty rows opens the dialog in move mode: destination locked (enforced in the store, not just the UI), an Untracked field, and a note naming the checkout being emptied. Failed moves with recovery state raise a persistent dialog with the stash instructions instead of a 4-second toast; successful moves report what moved.
- **NewSessionDialog.tsx** - The worktree name lives on its own row: derived live from the prompt while untouched (with an `auto` hint, since a derived name may come back suffixed), frozen and slugified in place once edited. An untouched name is omitted from the spawn, preserving derive-with-suffix semantics; only an edited name travels as explicit. The dialog now fits its rows to the terminal, degrading by dropping optional rows before interactive ones.

## Design decisions

- **A move refuses a pre-existing worktree.** Explicit names elsewhere mean create-or-open; for a move, opening an existing checkout is where every data-loss path lived (review proved a failed apply would have force-removed the user's worktree). Moving into existing checkouts is a different, riskier feature.
- **Copy mode excludes gitignored content**, matching move mode, so `.env` files and build artifacts never propagate to the new checkout. Gitignored files are the engine's file-setup conventions' job.
- **No reset step.** Per the issue thread: `git stash push` already leaves the source clean, and apply-then-drop makes the drop the single commit point.

## Review

Five independent review lanes (four Opus + one Codex) over the full diff, then two red-test-first fix rounds: 23 findings fixed, each reproduced failing before the fix, one refuted by measurement. Live e2e against a real daemon and a real Claude session covers the picker move end to end on disk (staged/unstaged split preserved, ignored files stay put, stash empty after), CLI copy mode, both refusal paths, and the name editing flow. Recordings in the comments below.

## Breaking Changes

None. The wire changes are additive; a new CLI against an older daemon errors loudly instead of degrading.

## Known limitations

- A rolled-back move deliberately leaves its stash entry as the backup; entries accumulate only across failures and the CLI prints the drop command.
- The OpenTUI input overflow for long typed values at sidebar widths predates this branch and is documented in AGENTS.md.

## Testing

- [x] 4011 tests pass across 157 files, typecheck clean
- [x] Every behavioral review fix red-proofed (failing test first)
- [x] Live e2e: picker move, on-disk verification, CLI copy mode, refusals, name editing (recordings below)
- [x] Dialog verified at 200x50, 63, and 34 columns, and at degraded heights down to the floor

> [!NOTE]
> Prefer a merge commit (or per-area squash) over a single squash: the branch is 35 commits with per-finding fixes, and a single squash makes a poor bisect target.


epilande:
Screen recording, sped up. Shows the picker's "Move changes" flow: filtering to a session, opening the right-click menu, the move-mode dialog, the worktree name deriving live as you type, and confirming the move.

https://github.com/user-attachments/assets/0db0d9e9-46b4-464d-b05c-91dfa8d650bf

epilande:
Screen recording, sped up. Shows the #83 worktree-naming flow: the derived worktree name appearing automatically, editing it to caps and spaces, slugifying it on Tab, and confirming; also shows the worktree marker and grouping in the picker.

https://github.com/user-attachments/assets/9680389f-bb84-4d7e-a601-5ab213b93b88

PR #97, opened 2026-08-01
feat(worktree): Fork a session into a fresh worktree

### LINKED ISSUE #70: [Feature]: Fork a session into a new worktree

epilande:
## Overview

Adds a worktree destination to Fork: a session forks into a freshly created worktree cut from its own branch, both from the CLI (`--fork` with `--worktree`) and from the picker ("Fork into worktree" in the row menu, through the new-session dialog in a fork mode).

Resolves #70

## Motivation

Fork continues a conversation without disturbing the original, but until now both sessions shared a checkout, and the fork-plus-new-worktree combination was refused outright pending live verification. A fork is usually the start of a divergent attempt, which is exactly the work that wants its own branch and checkout. The old refusal's premise (transcript resolution being directory-scoped) turned out to be false, and the shipped path-form resume never depended on it; the corrected design is on the issue.

### Engine

- **worktree-create.ts** - `readCheckoutHead` reads the source checkout's branch (sha when detached, null when unborn), and `slugForFork` derives `<source-branch>-fork` with the suffix budgeted inside the 40-char slug cap, so the result is a fixed point of `slugify` and the dialog preview equals the created name. A caller-derived name keeps derived semantics: numbered `-2`, `-3` on collision, never opening an existing checkout.
- **server.ts** - The fork+worktree refusal is gone. With no explicit name the worktree is named from the source branch; with no explicit base it is cut from the SOURCE checkout's branch rather than the main checkout's (gated on source and destination sharing a repo). `worktree.withChanges` on a fork is refused before anything is created: it would stash the working tree the source session is still running in.

### CLI

- **spawn.ts** - `ccmux spawn <agent> --fork <id> --worktree [name]` works end to end; `--with-changes` with `--fork` is refused before the daemon is even started. A new CLI talking to an older daemon that still refuses the combination now names the stale daemon and the `ccmux daemon restart` fix, instead of printing a message that reads as "this feature does not exist".

### Picker

- **App.tsx / store.ts** - Forkable rows in a repo gain "Fork into worktree" under "Fork". It opens the new-session dialog in a fork mode: no agent or prompt rows (a fork continues the source's), destination locked to a worktree, and a Name row previewing the derived name with the `auto · -2 if taken` hint. An untouched name is omitted from the POST so the daemon derives and numbers it; an edited name travels explicit. A detached HEAD is not previewed as a branch name, and a branch that slugifies to nothing asks for a typed name instead of promising one. Forking from the sidebar no longer steals focus, matching the dialog's other modes. `F` and plain Fork are byte-identical to before.
- **NewSessionDialog.tsx** - The fork shape has its own row budget (pinned by unit tests in both directions), and on the no-slug path the placeholder text is budgeted before the hint so narrow widths keep the explanation rather than the caveat.

### Docs

- **README / docs/agent-adapters.md / docs/architecture.md / AGENTS.md** - All three surfaces documented; the adapters doc records the live verification and the dialog mechanics description now matches what the types actually guarantee.

## Design decisions

- **A derived fork name never opens an existing checkout.** `<source-branch>-fork` numbers past collisions, so a second fork of the same session gets its own worktree instead of silently joining the first. An explicit name keeps create-or-open, same as any other spawn.
- **Base is the source checkout's branch, passed explicitly.** The engine's default base is the main checkout's branch, which is the wrong history for a fork whose source lives in a linked worktree, the same bug class the move fixed by resolving the source HEAD.
- **The preview and the daemon share one slug function.** What the Name row shows is what gets created; on success the toast reports the daemon's actual (possibly numbered) name.

## Review

Five independent review lanes (four Opus with real-git repros, one Codex) over the full diff, then two red-test-first fix rounds: 11 findings fixed, each reproduced failing first, and the load-bearing tests mutation-checked (the `-fork` suffix truncation was found independently by three lanes; the cross-repo base gate and the fork row budget were proven untested by mutation before gaining tests). Live e2e against a real daemon and real Claude sessions: the fork recalled a phrase planted before forking, the source transcript stayed byte-identical, a second fork numbered `-2`, the picker flow created exactly what it previewed, the `withChanges` refusal left no side effects, and a detached source derived `<sha12>-fork`. Recording in the comments below.

## Breaking Changes

None. The wire change is additive, and a new CLI against an older daemon explains the restart instead of failing mutely.

## Known limitations

- A fresh fork does not appear in the picker until it takes its first turn: Claude Code writes the forked transcript lazily, so the daemon has nothing to bind until then. Pre-existing behavior for any promptless spawn, noted here because forks make it visible.
- Resizing the terminal while the fork dialog is open can drop the `auto` hint on the no-branch path until the dialog is reopened (OpenTUI relayout quirk, cosmetic).
- A source branch that slugifies to nothing (for example a non-Latin name) needs a typed worktree name; both the dialog and the daemon's refusal say so.

## Testing

- [x] 4104 tests pass across 157 files, typecheck clean
- [x] Every behavioral review fix red-proofed (failing test first); key tests mutation-checked
- [x] Live e2e: CLI fork with history recall, second-fork numbering, picker dialog flow, wire refusal, detached HEAD naming (recording below)
- [x] Stale-daemon hint verified against a genuinely pre-#70 daemon build


epilande:
Live e2e recording: the picker's Fork into worktree flow (right-click menu, dialog with the derived name preview, Enter, and the toast reporting the created worktree).


https://github.com/user-attachments/assets/0f8cecc7-531c-4d29-8cbb-19c386601c8f

PR #98, opened 2026-08-01
feat(tmux): Select which tmux server ccmux uses

### LINKED ISSUE #95: [Bug]: multiple tmux sockets
### Preflight

### ccmux version

1.2.2

### Agent

Claude Code

### Are hooks installed?

Yes

### Title
No way to select the tmux socket; daemon silently binds to whatever `$TMUX` it inherits

### Problem
The daemon scans one tmux server, chosen entirely by the ambient `$TMUX` of the
process that happened to spawn it. Since it's spawned detached and auto-starts on
the first ccmux command, that's effectively luck. There's no flag, no config key,
and no env var to say which server to watch.

If the first ccmux command runs outside tmux, the daemon falls through to the
`default` socket. When that socket is stale — tmux leaves the inode behind when a
server dies — `getServerSocketPath()` gets a non-zero exit, returns null, and
`list-panes -a` finds nothing. The user sees "No active sessions" with no
indication anything is wrong.

### Repro
1. Run all tmux sessions on a named socket: `tmux -L mysock new-session -d`
2. Ensure a stale `/tmp/tmux-$(id -u)/default` exists from a dead server
3. From a plain login shell (not inside tmux): `ccmux`
4. Empty list. `curl -s localhost:2269/server-info` → `{"socketPath":null,...}`

### Workaround
    ccmux daemon stop
    TMUX="/tmp/tmux-1000/mysock,0,0" ccmux daemon start -b

### Suggested fix
- `--socket <path>` / `--label <name>` on `ccmux daemon start`, or a
  `CCMUX_TMUX_SOCKET` env var, threaded through as `-S`/`-L` on the spawned
  tmux commands.
- Surface the failure: when `display-message` exits non-zero, say which socket
  path was tried rather than reporting an empty session list.

### Environment
ccmux <version>, tmux <version>, Linux <distro>

### Steps to reproduce

 - use gascity (which maintains separate tmux server)
 - use ccmux

### OS and tmux version

debian

```shell

```

epilande:
## Summary

Adds a way to pin ccmux to a specific tmux server, and surfaces socket failures instead of a silent empty session list.

The daemon binds to whatever tmux server the ambient `$TMUX` of the process that auto-started it points at. Run your sessions on a named socket (e.g. a tool that keeps its own tmux server) and launch ccmux from a login shell, and the daemon lands on the stale default socket showing "No active sessions" with no indication anything is wrong.

- New override, in precedence order: `ccmux daemon start --socket <path>` / `--label <name>`, the `CCMUX_TMUX_SOCKET` env var (leading `/` means an `-S` path, anything else an `-L` label), and the `tmuxSocket` config key
- The daemon always honors the override; clients honor it only outside tmux (when attached, the ambient server wins and the existing cross-server guard covers mismatches)
- Every tmux invocation now goes through a central argv builder (**src/lib/tmux-exec.ts**) that injects `-S`/`-L` only when an override is active. With no override, argv is byte-identical to before (regression-tested)
- `/server-info` gains an additive `socketError` field, pane-scan failures record the attempted socket in scan health, and the picker, sidebar, and `ccmux show` print `tmux server unreachable at <path>` instead of an empty list
- `getServerSocketPath()` no longer caches its first success forever; a failed pane scan, or one that finds no server at all, invalidates it so a tmux restart onto a different socket re-resolves

Deliberately out of scope: multi-server aggregation. The setting selects the one server ccmux uses; running two daemons side by side stays unsupported (shared marker dir and log watchers).

## Related issue

Resolves #95

## Testing

- [x] `bun test` passes (4227 tests across 159 files)
- [x] Verified the affected TUI surface (picker) in a detached tmux session, both the new diagnostic and the normal session list
- [x] Verified end to end with a real agent session (see caveat in notes)

## Review fixes

A review pass with independent verification of every finding landed as 993c98a:

- Registers `tmuxSocket` as a config key, so the documented `ccmux config set tmuxSocket <value>` works (was `Unknown key`)
- Reads a non-string `tmuxSocket` in `ccmux.json` as unconfigured instead of throwing through `tmuxArgv` into every ccmux surface
- Surfaces a tmux server that dies after its socket was cached: pane scans report the no-server condition (`scanTmuxPanesOrThrow` returns `{ panes, noServer }`) and invalidate the `/server-info` cache, without touching scan health, session reaping, or the fail-closed discovery arm. Verified live: die, detect, recover, plus 12 consecutive no-server scans with no degraded flip
- Resolves a relative `--socket` against `CCMUX_CALLER_PWD` rather than the install root
- Makes the `switch` and `server-guard` tests hermetic when a socket override is configured on the developer's machine
- Documents the argv-builder invariant in AGENTS.md's Do Not list

## Notes for reviewers

Capture of the new empty-state diagnostic (picker against a daemon pinned to a dead socket):

```
 ● Sessions (0)
 tmux server unreachable at /tmp/nonexistent-socket-95
```

Live verification highlights:

- A daemon started with `CCMUX_TMUX_SOCKET=ccmux95` resolves `/private/tmp/tmux-501/ccmux95` in `/server-info` and installs its focus hooks on that server only; the default server's hooks stay pointed at the existing daemon. A real Claude session on the named socket is discovered with correct status.
- Label socket dirs are realpath'd deliberately: on macOS a label resolves under `/private/tmp/...`, while `-S /tmp/foo` stays verbatim. Without the realpath, `isSameTmuxServer` would compare `/tmp` against `/private/tmp` and refuse every pane.
- Sidebar hook bodies are NOT prefixed: the resize hook already interpolates `#{socket_path}` at fire time (server-correct by construction), and the auto-open body is a tmux command string where a `tmux` prefix would parse as a command name.
- One honest gap: marker-to-pane binding on a named socket was not positively proven end to end. Isolated-home test harnesses turn out to never bind panes on any socket (control-proven harness artifact), and the same build binds all 20 sessions on the default server, which rules out a binder regression.


epilande:
E2E verification: a real Claude session driven through idle -> working -> waiting (permission prompt) -> idle on the picker, filtered to the test session. Screen recording, sped up. Every step here (pane discovery, marker binding, state transitions) runs through this PR's central tmux argv builder in passthrough mode (no socket override configured), the default path all existing setups use. Representative run; the change affects all agents. The socket override itself (pinned daemon on a named socket, unreachable-socket diagnostic) is covered by unit tests plus the live checks in the PR description.

https://github.com/user-attachments/assets/b7438602-4a76-4a84-b7ce-f17add61019a

elmpp:
thank you!

PR #99, opened 2026-08-02
feat(tui): Collapse the new-session dialog's option fields into dropdowns

epilande:
## Summary

The new-session dialog's option fields (Agent, Placement, Where, Untracked) previously rendered their choices as stacked rows, so the Agent field alone pushed the rest of the form down by one row per detected agent. This series collapses every option field into a one-row dropdown pill and recomposes the dialog as a full-width form with a proper button row.

- **`DropdownField.tsx` (new):** a reusable `DropdownTrigger` (full-width pill: held value, caret pinned right, raised background that heats up on focus) and `DropdownOverlay` (absolute, z-raised bordered list that floats OVER the rows beneath instead of pushing them down, windowing itself via `optionWindow` with absolute number keys). The two are separate components because an absolute child only z-sorts against its own siblings in OpenTUI, so the overlay must be a late child of the dialog box itself.
- **`new-session-options.ts` (new):** one accessor (`newSessionOptions`) is the single source of truth for what each field offers and which option it holds, shared by the key routing, the pills, the overlay, and the store's value dispatch.
- **Keys:** space/l/right open the focused pill's list; while open the overlay owns every key (j/k move, enter/space/l/right confirm, h/left/esc cancel, 1-9 direct-pick). j/k still cycle a collapsed pill and 1-9 still direct-pick without opening. Which dropdown is open is one record on the draft (`dropdown: { field, index } | null`), so two can never be open at once.
- **Full-width form:** label column + `CONTROL_GAP`, text inputs in padded shell boxes on the same raised tokens, vertical spacers between fields, and a right-aligned Cancel / confirm button pair in the macOS order (primary rightmost, carrying the mode's verb: Spawn / Move / Fork). The buttons are click-only duplicates of Enter/Escape, deliberately not Tab stops.
- **Height budget:** the field spacers and the button row join `planDialogRows`' give-up ladder (spacers first, buttons second), so the dialog degrades in the same fixed order at tight heights; the overlay lives outside the budget entirely. Collapsing the agent field also deleted the stacked-options machinery, shrinking move-mode's floor from 18 to 14 rows.
- **Hints:** `newSessionHintSegments` is now the single source for the dialog's key hints, with state-specific copy (a focused pill teaches `space open`; an open overlay teaches `enter/space select`).

## Testing

- `bun run typecheck` clean; full `bun test` green (4125 pass / 0 fail across 157 files on this base, file count checked against disk).
- Plan tests cover every mode's row budget and the give-up ladder (spacers drop, then buttons, then error shrink); component tests assert row order; overlay anchor and `optionWindow` have their own unit tests.
- Verified live in detached tmux sessions at 200x50 and 60x10 (the tight-height drop) across all three dialog modes.
- Screencast of the dialog walkthrough in the PR comments below.


epilande:
ccmux-verify-pr surface demo (new-session dialog on picker). Shows (not driven): src/tui/App.tsx, src/tui/store.ts, src/tui/new-session-options.ts, src/tui/components/NewSessionDialog.tsx, src/tui/components/DropdownField.tsx, src/tui/components/Footer.tsx. Option fields collapsed into dropdown pills with a floating overlay, full-width form, right-aligned Cancel/confirm buttons.

Screen recording, sped up. Walkthrough: open the dialog, open the Agent dropdown (overlay floats over the rows below), j/j + enter to pick Codex, same flow on Placement, type a Prompt, peek the Where dropdown, esc to cancel. Recorded against an isolated daemon with an empty session list.

https://github.com/user-attachments/assets/fbb277a8-17a0-4569-818b-848148b96930

PR #100, opened 2026-08-03
feat(tui): Polish the row menus and fold Fork into the dialog

epilande:
## Overview

Polishes the picker's row menus and folds the two fork actions into one dialog-driven flow: calmer colors, sentence-case labels, a deliberate item order, and a new <kbd>m</kbd> hotkey that makes every menu action keyboard-accessible. Follow-up to #97 and #99.

## Motivation

- The menus used five colors across eight rows and mixed Title Case with sentence case, so the rows read as noise rather than meaning.
- "Fork" (instant, in place) and "Fork into worktree" (dialog) were two entries for one decision. The destination is the only choice a fork has, so one entry that always asks is fewer things to learn, and every fork gets a chance to name its worktree.
- Several menu items ("Fork into worktree", "Move changes", "Open agent view") existed only in the right-click menu, making them mouse-only dead ends.

### Menu appearance and order

- **App.tsx** - Neutral text for every row except Attach (green) and Kill / Kill group (red); sentence-case labels throughout ("New session", "Pin to top", "Prune worktrees", "Kill group"); session menu reordered to Attach, New session, Fork, Review diff, Move changes, Restart, Kill, with the destructive action last on the background-row menu too
- **PruneDialog.tsx** - Heading follows the "Prune worktrees" rename

### Fork consolidation

- **App.tsx** - One "Fork" entry (hint <kbd>F</kbd>); the menu item and the <kbd>F</kbd> key both open the new-session dialog in fork mode; `submitFork` merges the old instant-fork and fork-into-worktree submit paths, keeping the in-flight guard, the 15s timeout, and the detached-HEAD name handling
- **store.ts / new-session-options.ts / NewSessionDialog.tsx** - Fork mode unlocks the existing destination dropdown instead of being pinned to a worktree: "This checkout" is the default, so <kbd>F</kbd> then <kbd>Enter</kbd> posts the exact payload the old instant fork sent, and "New worktree" reveals the name field as before; a source outside a repository opens with the destination locked to its own checkout

### Keyboard access

- **App.tsx / store.ts / ContextMenu.tsx / SessionList.tsx** - <kbd>m</kbd> toggles the selected row's menu (session menu on rows, group menu on headers), anchored at the row's own screen line; <kbd>j</kbd>/<kbd>k</kbd> move a highlight, <kbd>Enter</kbd> activates, <kbd>Esc</kbd>/<kbd>m</kbd> close, and any other key dismisses the menu and keeps its ordinary meaning
- **store.ts** - The highlight is keyed to a stable item id rather than a row position: "Move changes" arrives asynchronously and Fork can vanish on an SSE update, and neither is last anymore, so an insertion above the lit row must not move the highlight onto a neighbour; an item that disappears takes its highlight with it and <kbd>Enter</kbd> does nothing
- **HelpOverlay.tsx** - New <kbd>m</kbd> row; the "Other" section moves to the right column and the sidebar's compact list drops the blank row between entries to fit

## Breaking Changes

None

## Testing

- [x] `bun test` passes (4159 tests across all 157 files)
- [x] Verified the picker and sidebar in detached tmux sessions: menu colors captured hex-exact via `capture-pane -e`, <kbd>m</kbd>-opened menus anchored on the selected row, highlight movement and clamping, the fork dialog in both destinations, and the locked no-repo case
- [x] Verified against real agent sessions and the live daemon (fork dialog over a forkable Claude row; fork payloads for both destinations covered by App tests rather than submitted against real sessions)

## Notes for reviewers

- Sidebar <kbd>F</kbd> now follows the dialog convention (toast, no jump to the new pane) instead of the old instant fork's jump; both destinations behave the same way.
- When "Move changes" arrives async, Restart and Kill shift down one row while the menu box stays put. That wobble is inherent to putting Kill last, and it is safe because neither input path addresses items positionally.
- The "Kill Group?" confirmation dialog keeps its Title Case, matching its "Kill Session?" / "Restart Session?" family; sentence-casing that family is a possible follow-up.


epilande:
Screen recording of the menu changes on this branch (sped up).

What to look at:

- **`m` opens the row menu** on the selected row, anchored to it, with the first item lit. `j`/`k` move the highlight, Enter runs it, Esc or `m` closes. The menus were right-click only before, which left Restart, Kill and Move changes with no keyboard route at all.
- **The new order and palette.** Attach, New session, Fork, Review diff, Move changes, Restart, Kill, with the destructive action last. Only Attach (green) and Kill (red) carry color now; everything else is plain text.
- **`m` on a group header** opens the group menu: New session, Pin to top, Pin to bottom, Prune worktrees, Kill group.
- **One Fork entry.** `F` opens the dialog instead of forking on the keystroke. It starts on **This checkout** with no Name row, so Enter there is the old one-shot fork unchanged; switching **Where** to a new worktree reveals the Name row with the derived `<branch>-fork` preview.

The clip cancels out of the fork dialog rather than submitting it. The submit paths (both destinations, their payloads, and the highlight staying on its own item when "Move changes" arrives late) are covered by tests.

https://github.com/user-attachments/assets/a3e0d7df-33f2-43b2-8e86-bdc8ba1f69d5

PR #105, opened 2026-08-04
feat(tui): Full Worktrees panel replacing prune-only W (#102)

### LINKED ISSUE #102: feat(tui): Turn W into a full Worktrees panel (replaces PruneDialog)
## Problem

`W` is mnemonically "worktrees" but all it does is prune. The dialog can only remove, and its read path is prune-shaped all the way down: the scan classifies a worktree only when it has a proven removal reason and silently drops everything else (open PR, detached HEAD, or simply nothing to say). A healthy worktree with no session attached is therefore invisible in the entire product: you can't see it, jump to it, or spawn an agent into it from ccmux. Repos are also discovered only through live sessions, so a repo whose agents all exited disappears along with its stale worktrees, which is exactly when you want to reclaim them.

Creation is well covered (`n` dialog worktree destination, move changes #71, fork #70, `ccmux spawn --worktree` #69). The gap is seeing and acting on worktrees that already exist.

## Proposal

Turn `W` into a **Worktrees panel** that replaces PruneDialog, with prune as one action inside it. Scope line: worktree lifecycle as it relates to agent sessions, not a general git GUI.

### Panel

- Modal overlay in PruneDialog's shape, with a compact sidebar variant.
- Repo-scoped like today (selected row's repo; all repos when nothing matches), plus a **Tab** toggle inside the panel to widen to all repos without closing it.
- Repo discovery: session-derived repos plus the repo containing the TUI's cwd (fixes the zero-session invisibility).
- Each repo group lists the **main checkout first** as a context anchor (never prunable), then worktrees sorted: active sessions, idle, prunable last.
- Per row: branch, dirty counts, **ahead/behind** (new data; extend the `%(upstream:track)` parse beyond `[gone]`), PR badge (including healthy open-PR worktrees the current scan silently drops), attached session with status, and the prune reason where one exists. Worktrees the prune scan skips (locked, agent working) show their skip reason inline, replacing today's "N not offered" count line.

### Loading

Two-phase, since fetch and PR resolution can take seconds:

1. Instant render from local-only data (worktree list, branches, dirty state, attached sessions).
2. Enrichment streams in (fetch, upstream-gone, PR state, prune classification); rows become selectable for prune as their reason arrives. One re-sort when classification completes, cursor following its row, so the list settles once instead of churning.

Protocol: a new `GET /worktrees` endpoint for phase 1; phase 2 reuses `GET /worktrees/prune-candidates`, merged by path. `POST /worktrees/prune` is untouched (it already re-derives everything server-side).

### Actions

- **Enter** is context-sensitive: jump to the attached session, or, when there is none, open the new-session dialog in a new "existing worktree" mode (directory locked to the row's path, agent and prompt fields shown). Occupied rows can only jump; there is deliberately no second-session-in-a-worktree path.
- **space / a / D / confirm**: prune selection and removal, carried over from PruneDialog.
- **y**: copy the worktree path.
- **d**: diff review, based on merge-base with the repo's default branch.

The "existing worktree" spawn mode is reachable from the panel only; the `n` dialog's destination dropdown is unchanged.

### Surrounding surface

- `ccmux worktree list` CLI subcommand as a thin formatter over the new endpoint.
- Footer gains a low-rank `W worktrees` hint (currently only the help overlay mentions `W`).
- Labels rename from "Prune Worktrees" to "Worktrees" (dialog title, help overlay, group context menu).

## Non-goals

- General git operations (rebase, merge back, rename, lock/unlock).
- A worktree picker inside the `n` dialog's destination dropdown.
- A configured repo registry; discovery stays sessions + cwd.

## Delivery

Single PR: backend read path (list-all scan mode, `GET /worktrees`, ahead/behind), the panel replacing PruneDialog, the new-session dialog's fourth mode (via `NewSessionShape`, so both policy functions must handle it, plus its `planDialogRows` counts and tests), and the CLI subcommand.

epilande:
Closes #102.

Turns `W` from prune-only into a full **Worktrees panel**: every worktree a repo has (main checkout included) with branch, dirty counts, ahead/behind, attached sessions with live status icons, PR badges, and prune as one action inside the panel instead of the whole panel.

## What's in here

**Backend read path**

- New local-only `GET /worktrees` (`worktree-list.ts`): all worktrees per repo with branch, dirty counts, upstream ahead/behind (extends the `%(upstream:track)` parse beyond `[gone]`), and attached sessions. No fetch, no `gh`, by contract, so it paints instantly.
- Prune scan now also reports healthy open-PR worktrees (`open` bucket) instead of silently dropping them, and repo discovery adds the caller's cwd repo so a repo with zero live sessions is still visible (including bare-repo main roots).
- Live subagents working in `agent-*` worktrees are attributed to those worktrees (`SubagentState.worktreePath` from `agent-<id>.meta.json`, synthetic background sessions), which also closes a prune-gate blindspot: a worktree only a subagent occupies is no longer classified as removable. A companion cwd-drift guard (`adoptsLoggedCwd`) stops a parent session from being dragged onto its teammate's worktree row by transcript cwd entries.
- Last-moment pane-occupancy guard before removal: a pane whose cwd is inside the worktree and isn't a bare shell (editors count as occupants) blocks that removal; the caller's own pane is exempt. Fail-soft if tmux is uninspectable (logged, removal proceeds).
- `ccmux worktree list [--repo]` CLI as a thin formatter over the endpoint. Client cwd travels as `CCMUX_CALLER_PWD` because `bin/ccmux` cds to the package root.

**Panel** (`WorktreesPanel.tsx`, a `git mv` of PruneDialog)

- Two-phase load: phase 1 renders the local list immediately and stays fully navigable; phase 2 (`GET /worktrees/prune-candidates`) merges in by path with ONE re-sort (cursor follows by path, scrolling recomputed in visual lines). A muted animated `· ◑ scanning` title suffix marks the in-flight scan so the re-sort reads as completion, not a glitch; phase-2 failure degrades to a read-only panel with one error line.
- Two-line rows with a continuous rail, plain-word detail phrases (`2 modified · 4 untracked`, `branch gone`, `agent working here`), session-list status icons (spinner and all), facts stated once (a merged PR suppresses the badge and tracking noise; an occupied row's liveness skip is folded into its session summary). Removable rows cluster under a per-repo `├─ removable · N ──` rule and are the only rows with `[ ]`/`[x]` checkboxes.
- Context-sensitive **Enter**: jump to the attached session (revalidated against the live store, not the phase-1 snapshot), or open the new-session dialog in a new "existing worktree" mode (fourth `NewSessionShape` mode; directory locked to the row). **x** select/remove with a centered confirmation dialog that pluralizes for real and never silently no-ops; **D** includes dirty rows; **y** copies the path via OSC 52 + pbcopy (both, deliberately); **d** reviews the BRANCH against its merge-base with the default branch (falls back to a working-tree review when no merge-base exists); **Tab** widens repo scope.
- Compact sidebar variant, footer `W worktrees` hint, labels renamed from "Prune Worktrees".

## Review process

Built in three parallel lanes with disjoint file surfaces, then a 4-lane review fan-out (3× Opus + an independent Codex pass). All 3 Must-fix findings were fixed and re-verified: scroll position computed in visual lines instead of row indexes (cursor could leave the screen while keys acted on it), the send-review confirmation no longer buried under the panel overlay, and the CLI cwd fix above (Codex-only find). All Should-fixes are in. Several rounds of live-usage feedback drove the presentation redesign, the empty-`x` behavior, the checkbox glyphs, and the scanning indicator.

## Verification

- `tsc --noEmit` clean; `bun test` **4454 pass / 0 fail** across 161 files.
- Live-verified against the real daemon (hotkeys + PATH + daemon on this build for a full day of use): both phases and widths, all three status states animating, subagent attribution, removal flow end to end, old-daemon degradation (404 → error state).
- Renames staged via `git mv` (PruneDialog → WorktreesPanel), history preserved.

## Known limitations and notes for the record

- The panel is a fetch-on-open snapshot: status icons can go stale if it's left open (cosmetic; follow-up candidate).
- OSC 52 copy success is unconfirmable by design (a terminal can accept the sequence and copy nothing), hence the dual-channel write.
- The Tab scope-widen toggle is unit-tested but was never exercised on the live rig (isolated-rig detection quirk).
- The prune endpoints' repo discovery intentionally widened (additive `cwd`), so a repo with no sessions is now reachable; the classification gate, not the repo list, remains the safety boundary.
- Pre-existing (verified on main, not introduced here): a narrow race between the prune scan's session gate and removal, mitigated by the new pane-occupancy last-moment guard; the session-derived discovery path still can't list bare-repo worktrees (`worktreeFacts` returns a null main root there — separate change).
- `App sidebar mode > debounces flash during rapid navigation` flaked once during development and passed on retry; pre-existing, unrelated to this branch.


epilande:
Screen recording of the panel (sped up): `W` opens it scoped to the repo, the title shows a scanning spinner while the phase-2 scan (fetch + PR resolution) runs in the background with the list fully navigable, then the removable worktrees settle into their checkbox section with PR badges and skip reasons. Then: selecting a removable worktree with `space`, opening the removal confirmation with `x`, and cancelling it.

https://github.com/user-attachments/assets/8540c224-69db-47d9-b96f-661640c56871

PR #107, opened 2026-08-05
feat(daemon): Read a session's last response (ccmux last)

epilande:
## Summary

Foundation for session handoff: read a session's last response (or last N turns) from its transcript, addressable by a fuzzy session reference.

- **`ccmux last <ref>`**: prints the referenced session's last assistant response bare on stdout (pipe-friendly; resolution notes go to stderr). `--turns N` (1-20) renders the last N assistant turns with the user prompts between them (`assistant:` / `user:` labels, oldest first). `--json` emits the full envelope.
- **`GET /sessions/:ref/transcript`**: the daemon endpoint behind it. Frozen response contract: `sessionId`, `agentType`, `turns[]`, `source: "transcript" | "pane"`, `truncated`, plus an additive `resolution` field (`tier`, `exact`, `proximity`).
- **Session resolver** (`src/daemon/session-ref.ts`): exact session id, then pane id, then window coordinate, then `self` (the caller's own pane via `$TMUX_PANE`), then agent-type and project-name matches scoped by tmux locality (same window > same session > global). Ambiguity is a refusal listing every candidate (id, coordinate, agent, status, cwd) with exit 1, never a guess.
- **Transcript readers** (`src/daemon/transcript-readers/`): a per-agent registry seam, with Claude, Codex, and Copilot JSONL readers sharing `foldJsonlTurns`. Later PRs add the remaining agents through the same seam.
- **Caps**: per-turn 20k tail-preserving cap; lines over 256KiB are skipped before parsing (measured single tool-output lines run 146K to 665K chars). Only assistant text is returned, tool outputs are never inlined.

Sessions without a readable transcript fall back to a pane capture (`source: "pane"`).

## Testing

- Full suite passes, typecheck clean; new unit tests for the resolver, the transcript read, the readers, and the CLI command.
- Verified end to end against live Claude and Codex sessions. Demo video in the first comment.


epilande:
Demo (24s): `ccmux last`, the JSON envelope, and how a reference resolves. Beats: exact id prints the response bare; `--turns 3 --json` shows the frozen envelope and the 2N-1 turn shape; a fuzzy `codex` ref resolves scoped by tmux locality with the note on stderr; an ambiguous `claude` ref refuses with both candidates and exit 1.

https://github.com/user-attachments/assets/17c69139-7dec-4bcc-bc69-dc9223fb14ce

PR #109, opened 2026-08-05
feat(daemon): Hand off a session's last response to another session (ccmux handoff)

epilande:
## Summary

The handoff itself: send one session's last response to another session as a submitted prompt, over the read foundation from #107 and the delivery hardening from #108.

- **`ccmux handoff <from> <to>`** / **`POST /handoff`**: composes a provenance header, an optional `--note`, and the source's last response, then delivers it to the target's pane as one submitted prompt. `from` and `to` take the same fuzzy refs as `ccmux last` (`self` works from inside an agent's own shell). `from == to` is refused.
- **Provenance header**: ``[ccmux handoff] from: <session-id> (<agent> · `<cwd>` · branch <b>) at <time>``, followed by the note on its own line. The header leads the composed text, so a payload can never open with `/` or `!` (slash-command defuse), and the source id doubles as a pull-more pointer: the receiver can run `ccmux last <id> --turns N` for more context. The cwd's backticks are load-bearing: they keep the header itself from tripping Cursor's whitespace-slash unsafe-reply guard. The genuine header is the only column-0 `[ccmux handoff]` line in the message; any payload line that tries to look like one arrives quoted with `> `.
- **Delivery policy**: an idle target gets the text immediately (deliveries to one target are serialized, and each send uses a unique tmux buffer, so concurrent handoffs cannot interleave or clobber each other); a working target queues the handoff (`pendingHandoff` rides the wire additively, the TUI shows a ⇄ badge) and delivery fires on the next transition into idle, with guards re-run at dequeue. Content guards that are decidable at enqueue time (the target agent's unsafe-reply pattern) refuse up front instead of queueing something that can never deliver. Transient failures at dequeue (a tmux hiccup, the target busy again) re-queue with a bounded 3-attempt budget inside the 30-minute TTL; deterministic refusals drop. A waiting target is refused outright, because a handoff must never answer a permission prompt. Composition happens at enqueue time, so a queued handoff holds the bytes that existed when it was sent.
- **`--spawn`**: instead of a target, open a new session (same spawn plumbing as the picker) with the composed handoff as its opening prompt. The composed text is pre-checked against the spawn prompt byte budget, with a clear refusal naming the fix (fewer `--turns`) instead of a spawn-internal error.
- **Caps and hygiene**: one 65,536-char cap on the composed text for both the direct and queued paths (head-dropped, `truncated` flagged); the final composed text is control-char stripped, and the header's own facts (cwd, branch) are additionally newline-stripped, so a hostile note, path, or payload cannot smuggle escape sequences or forged header lines into a pane.

Known limitations, deliberate for now: the pending-handoff queue is in-memory, so a daemon restart drops a queued handoff after the sender was told "queued" (documented in the docs PR of this series; persistence is a follow-up if it bites in practice). And a Cursor target still refuses a handoff whose payload mentions absolute paths, because Cursor's live-verified unsafe-reply guard matches whitespace-then-slash anywhere; the refusal is now honest (409 at enqueue, never a silent drop), and relaxing it would need live verification that a tmux paste cannot trigger Cursor's command palette the way typed input does.

## Testing

Full suite passes, typecheck clean. Endpoint tests cover the refusals (not-found, ambiguity, waiting target, self-handoff), the queue-then-deliver path with no double delivery, and cap/strip behavior. Demo videos in the first comment: delivery into an idle composer plus the waiting refusal and `send --stdin`, queue-then-deliver with the ⇄ badge, an agent handing off its own conclusion via `self`, and `--spawn`.


epilande:
Demos of the delivery contract, in order:
1. deliver-refuse-stdin (21s): a handoff lands in an idle composer as one submitted prompt; the same command against a target holding a permission prompt gets the verbatim refusal with exit 1 and the dialog untouched; a piped `ccmux send --stdin` lands as a single two-line message.
2. queue-on-busy (22s): handoff to a working target queues, the ⇄ badge sits on the row, and the queued handoff types itself in when the turn ends, badge clearing with no second command.
3. handoff-self (23s): an agent runs `ccmux handoff self <target> --note "pull more with ccmux last <id> --turns 3"` from its own pane; the receiver immediately acts on the note's pointer.
4. spawn (15s): `ccmux handoff <from> --spawn` opens a new session with the composed handoff already submitted as its opening prompt.

https://github.com/user-attachments/assets/601c7cbc-3f91-41a2-98f7-ae4bb5bcedf6


https://github.com/user-attachments/assets/cb698b2b-9269-4078-a39f-c862f5734ed1


https://github.com/user-attachments/assets/d54ca3b3-f130-49e5-8233-b5bd4d5f6ad3


https://github.com/user-attachments/assets/1a64c7b1-d84b-49b5-bcc7-88c98b01612f

PR #112, opened 2026-08-05
feat(tui): Hand off a session's last response from the row menu

epilande:
## Summary

The TUI face of #109: hand a session's last response to another session from the row menu.

- **Hand off** in the row context menu arms pick-target mode: the picker becomes a target picker, with a banner naming the source (`⇄ Hand off from <agent> · <project> · pick a target`) and the footer hints switching to the pick verbs. `j`/`k` aim, Enter sends through `POST /handoff`, Esc cancels the whole gesture.
- **Key swallowing while aiming**: every non-pick key is inert during pick mode, so a stray `x` cannot kill the row you are pointing at and `W` cannot open a panel over the aim. Verified on camera in the demo.
- **Outcome toasts**: delivered (`Handed N chars to <agent> · <project>`), queued (with the explanation that it delivers when the turn ends), and every daemon refusal (waiting target, self-handoff, ambiguity with candidates) surfaced as a toast rather than silence.
- **⇄ badge**: a row whose session holds a `pendingHandoff` shows the badge straight off the wire field, and it clears with the SSE update after delivery.
- **Store fix that this surfaced**: `updateSession` merged wire sessions into the store, so an optional field that a later update omitted (exactly what `pendingHandoff` does after delivery) stuck forever; it now reconciles, letting omitted fields clear. This is cross-cutting and benefits every optional wire field.

The Enter-on-target gesture fires the handoff immediately in this PR; a later PR in the series inserts the turns + note dialog between aim and send.

## Testing

Full suite passes, typecheck clean. Component tests cover pick-mode entry/exit, key swallowing, each toast branch, and badge render; the queue badge and pick flow are shown live in the demo video below.


epilande:
Demo (56s), covering this PR plus the Copy dialog and labels that land later in the series: `m` opens the row menu with Copy and Hand off; the Copy dialog's turns selector and two-digit entry; Esc copies nothing; then Hand off arms pick-target mode with the banner and footer hints, a stray `x` while aiming is swallowed (both rows survive), and Enter delivers with the toast and the target flipping to working.

https://github.com/user-attachments/assets/e4a18c32-b84b-40e4-8efe-aae117cc8e91

PR #127, opened 2026-08-06
fix(tui): Restore terminal modes on process.exit quits

### LINKED ISSUE #125: TUI leaves mouse tracking (?1003h/?1006h) enabled on exit/detach — terminal spews SGR mouse sequences
### Summary

After using the ccmux TUI (e.g. `ccmux` picker / `ccmux sidebar`) and returning to the shell — either by quitting it or by **detaching while it's still running** — the terminal is left in mouse-reporting mode. Any subsequent mouse movement then emits SGR mouse sequences (e.g. `35;71;29M35;70;29M…`) straight into the shell, corrupting the prompt / command line.

ccmux enables mouse tracking for its UI (`?1000h`/`?1002h`/`?1003h` any-motion + `?1006h` SGR) but does **not** disable those modes when it exits or is detached / `SIGHUP`'d, so the host terminal stays in mouse mode.

### Repro
1. A terminal that reports mouse events (e.g. Ghostty).
2. Run `ccmux` (bound to a named tmux socket via `tmuxSocket`, though that isn't required to repro).
3. Quit ccmux, **or** detach the tmux client while ccmux is still running.
4. At the returned shell prompt, move the mouse → the line fills with `<btn>;<col>;<row>M` SGR sequences.

### Expected
On exit **and** on `SIGHUP`/`SIGTERM`/detach, ccmux should restore the terminal modes it changed — i.e. emit `\e[?1000l\e[?1002l\e[?1003l\e[?1006l` (and generally undo any DEC private modes / alternate screen it enabled).

### Workarounds
- `reset`, or `printf '\e[?1000l\e[?1002l\e[?1003l\e[?1006l'`.
- There is no config option to turn mouse off — `ccmux config get mouse` / `mouseEnabled` / `mouseTracking` all report "(not set)".

### Environment
- ccmux built from `main` (commit `c7594d5`).
- Linux x64, terminal **Ghostty**, over SSH, used with a named tmux socket (`tmux -L <name>`, via the `tmuxSocket` config key).

epilande:
## Summary

Restores the host terminal's modes when the TUI quits. Every user-visible quit path (<kbd>q</kbd>/<kbd>Escape</kbd>, <kbd>Enter</kbd>-to-jump, spawn-enter) exits via `process.exit()`, which skips OpenTUI's `beforeExit` cleanup hook, so the mouse-tracking modes (`?1000h ?1002h ?1003h ?1006h`) and the `?1049` alternate screen were never restored. Anywhere the TUI shares a real terminal with the user's shell (picker run directly in a pane, persistent pickers), every subsequent mouse move spewed SGR sequences like `35;71;29M...` into the prompt. tmux popups masked the leak by discarding the popup's pty.

`launchTUI` now always creates the renderer itself (previously the picker path let `@opentui/solid` create it internally, leaving no handle) and installs a `process.on("exit")` listener that calls `renderer.destroy()`, whose native shutdown emits the full disable block. `destroy()` is synchronous and idempotent, so it is safe from an `exit` listener and composes with OpenTUI's signal handlers, which already cover `SIGHUP`/`SIGINT`/`SIGTERM` correctly (verified by counter-test: a SIGHUP'd picker restores cleanly on both old and new builds). The sidebar focus dance keeps its ordering: focus is stolen before renderer creation, since creation fires the capability probes.

## Related issue

Resolves #125

## Testing

- [x] `bun test` passes (4930 pass / 0 fail)
- [ ] Verified end to end with a real agent session (if this touches detection, state, the daemon, or the CLI)

Verification used tmux's per-pane mode flags as ground truth:

- Before the fix, picker + <kbd>q</kbd>: `mouse_any=1 mouse_sgr=1 alt_screen=1` left armed (reproduces #125 exactly).
- After the fix, picker + <kbd>q</kbd>: `mouse_any=0 mouse_sgr=0 alt_screen=0`.
- Sidebar direct launch renders and <kbd>q</kbd> kills its pane as designed; `ccmux sidebar --toggle` still spawns unfocused with no capability-probe gibberish in the shell pane, and a second toggle tears it down.

## Notes for reviewers

- The issue's proposed mechanism (missing `SIGHUP` handling) is not the actual gap: OpenTUI registers handlers for `SIGHUP` and friends, and they restore correctly. The gap is `process.exit()`, which emits only `exit`, and OpenTUI listens on `beforeExit`.
- Upstream OpenTUI has adjacent open issues (anomalyco/opentui#904 / #905, a brief echo window during destroy ordering) that are distinct from this bug and not addressed here.
- A `mouse: false` config preference (also noted in #125) would be a separate small feature.
- E2E recording to follow as a comment.

### DIFF
--- a/src/tui/index.tsx
+++ b/src/tui/index.tsx
@@ -119,10 +119,17 @@ export async function launchTUI(options: TUIOptions = {}): Promise<void> {
   // Quit paths exit via process.exit(), which skips OpenTUI's `beforeExit`
   // cleanup (its signal handlers do fire, but `process.exit` emits only
   // `exit`), leaving mouse tracking and the alternate screen armed on the
-  // host terminal (issue #125). destroy() is synchronous and idempotent,
-  // so it is safe here even when a signal-driven destroy already ran.
+  // host terminal (issue #125). destroy() is idempotent, so it is safe here
+  // even when a signal-driven destroy already ran, and it restores the
+  // terminal synchronously as long as no frame callback awaits real I/O
+  // (destroy() during a live render pass defers the restore to the render
+  // loop, which never resumes once process.exit is in flight). The catch
+  // matters: destroy() runs every Solid onCleanup before the native restore,
+  // and a throw there would skip the restore and reinstate #125.
   process.on("exit", () => {
-    renderer.destroy();
+    try {
+      renderer.destroy();
+    } catch {}
   });
 
   await render(
--- a/src/tui/index.tsx
+++ b/src/tui/index.tsx
@@ -100,18 +100,30 @@ export async function launchTUI(options: TUIOptions = {}): Promise<void> {
 
   // Sidebar spawned via -d into an unfocused pane needs the focus dance
   // to keep terminal capability probe replies from leaking into the
-  // user's shell. Picker/non-sidebar TUIs run in the user's own focused
-  // pane, so the dance is a no-op there.
-  let rendererOrConfig: CliRenderer | typeof config = config;
+  // user's shell. Focus must be stolen before the renderer exists, since
+  // creating it fires the probes. Picker/non-sidebar TUIs run in the
+  // user's own focused pane, so the dance is a no-op there.
+  let restoreTarget: string | null = null;
   if (options.sidebar && process.env.TMUX_PANE) {
-    const restoreTarget = await findRestorePane();
+    restoreTarget = await findRestorePane();
     if (restoreTarget) {
       await selectPane(process.env.TMUX_PANE);
-      const renderer = await createCliRenderer(config);
-      arrangeSidebarFocusDance(renderer, restoreTarget);
-      rendererOrConfig = renderer;
     }
   }
+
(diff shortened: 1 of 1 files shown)


epilande:
Screen recording of the fix (sped up, with cuts): quitting the picker with `q` on the current main build vs this branch, in the same tmux pane, using tmux's per-pane mode flags as ground truth.

1. Main build (`ccmux-main`): after quitting, the pane is left with mouse tracking still armed. The flags read `mouse_any=1 mouse_sgr=1`. The quit also strands the pane on the leaked alternate screen; a blind `printf '\033[?1049l'` (the workaround from #125) was needed off-camera just to make the prompt visible again, and the mouse modes survive even that.
2. This branch: `q` returns straight to a visible prompt, and the flags read `mouse_any=0 mouse_sgr=0 alt_screen=0`.

The picker is filtered to a throwaway e2e Claude session; cuts hide unrelated local sessions.

https://github.com/user-attachments/assets/9f57ac14-c23d-4b95-b43f-52c7f6a59200

PR #129, opened 2026-08-06
feat(tui): Unify dialog styling and hint placement

epilande:
## Summary

Brings the Hand off and Copy dialogs onto the new-session dialog's visual language, so every dialog reads as one system. Hand off was the outlier: it is a form (two fields plus a send) but was drawn in neither of the two dialog families, with its own title colour, its own `>` focus marker, a bare turns row next to a shelled note row, no buttons, and its source as an unlabeled subtitle.

**Hand off**

- Bare bold `Hand off` title (the New session title's rule), focus marked with the same `▎` blue bar, both fields in the shared surface/border control shells (click to focus).
- Right-aligned `Cancel` / `Send` buttons in the shared quiet-surface / mauve-primary style, click twins of Enter and Escape, budgeted as one droppable unit.

**Endpoint rows**

- Both ends of the handoff are named by `From` / `To` rows that read like the session list: project:branch context dimmed, agent name in its brand colour, and the tmux `session:window.pane` target undimmed. Same-project handoffs differ by nothing but the pane, so the pane survives narrowing last: the context middle-truncates first and the agent drops whole before the pane gives an inch (`fitHandoffEndpoint`, pure and tested).
- From sits above To, the way the text will travel. The row budget is unchanged: To stays in the height floor (it names the irreversible half), and From still gives way first at short heights.

**Pick banner**

- The hand-off pick banner is now one mauve sentence: `⇄ Hand off from ccmux:1.1 · pick a target`. The agent · project label is gone (the source row was just picked from this very list, and the dialog that follows names both ends in full), and so are the enter/esc hints, which duplicated the footer's pick arm. A source with no pane falls back to the agent · project label rather than pointing at nothing.

**Row menu accelerator**

- <kbd>m</kbd> then <kbd>h</kbd> triggers Hand off from the row menu. `ContextMenuItem` grows an optional `key`, a menu-local accelerator handled while the menu is open; the other hints name list-wide keys that fall through and close the menu, a path `h` cannot take because on the list it collapses a group.

**Copy**

- Title colour aligned, the turns selector drawn in the focused control shell, and a `Cancel` / `Copy` button row added.

**Hints follow the new-session dialog's split.** The picker's Footer carries each open dialog's key hints (new `handoffDialogMode` / `copyDialogMode` arms, authored once as shared segment lists so the two surfaces cannot drift), and only the sidebar's dialogs, which have no footer, draw their own hint row: confirm-first, green `enter` / red `esc`, middle segment given up when compact.

**Smaller fixes while in there**

- `NoticeDialog` title joins the shared colour; severity stays in the message lines.
- `ConfirmationDialog` switches from 50%-offset/negative-margin centering to the arithmetic centering every other dialog uses (the odd-height off-by-one documented in `NoticeDialog`), and clamps its fixed 50-column width so it fits a 30-column sidebar instead of overflowing.

The `Y confirm / N cancel` language on the confirm dialogs is left alone deliberately: that is the one-question family's own consistent convention, shared with the Worktrees panel's removal confirm.

## Testing

- [x] `bun test` passes (4942 tests)
- [ ] Verified end to end with a real agent session (if this touches detection, state, the daemon, or the CLI) - N/A, TUI only

## Notes for reviewers

Hand off in the picker (footer carries the hints):

```
│ Hand off                                         │
│                                                  │
│ ▎Turns  Last response                            │
│                                                  │
│  Note   note (optional) · sent in the header     │
│                                                  │
│  From   ccmux:main  claude  ccmux:1.1            │
│  To     ccmux:main  codex   ccmux:7.4            │
│                                                  │
│                                  Cancel    Send  │
│                                                  │
└──────────────────────────────────────────────────┘
 enter send · j/k turns · tab note · esc cancel   <- Footer
```

Copy in the sidebar at 40 columns (no footer, so the dialog carries its own compact hint row):

```
 ▼┌──────────────────────────────────┐
 ●│ Copy from claude · ccmux         │
  │                                  │
 ●│  Last response                   │
  │                                  │
 ─│                  Cancel    Copy  │
 ▼│                                  │
 ●│ enter copy · esc                 │
  └──────────────────────────────────┘
```

Row budgets: both dialogs give rows up in the shared order (blank rows, then buttons, then the From row, then hints), and the plans now take a `keyHints` flag so the picker's dialogs do not budget a row for hints the footer is already showing. The footer reverts to the default hint line the moment the dialog closes, verified live in both surfaces.


epilande:
Screen recording of the reworked dialogs, sped up. Nothing was sent or copied: both dialogs are cancelled with Esc.

**Picker.** `m` on a row opens the row menu, where `h` now goes straight to Hand off. That drops you into pick-a-target mode on the list itself (mauve banner, move the highlight, Enter). The dialog opens with a bare "Hand off" title, the Turns selector, the Note field, then From above To, each endpoint named by project:branch, agent, and pane, coloured per token. The footer carries the hints, so the dialog draws none of its own. Same path again for Copy (`y`), which gets the turns selector and Cancel/Copy.

https://github.com/user-attachments/assets/7540ecbb-cfd9-408a-a933-e180f9da3b78

**Sidebar.** The same two dialogs at 40 columns. There is no footer here, so each dialog draws its own hint row instead (`enter send · esc`, `enter copy · esc`), and the From/To rows shed the agent token as the width runs out.

https://github.com/user-attachments/assets/a8275403-7a0a-4932-b4cd-b8f3ff5903ea

PR #132, opened 2026-08-11
fix: rearm watcher tree after null events

LiadOz:
## Summary

Reopen the native watcher’s root walk and sweep gates when `fs.watch`
delivers an event without a path.

This prevents rapid root-level additions and removals from being missed when
the filesystem reports colliding directory timestamps.

## Testing

- [x] `bun test` passes
- [ ] Verified end to end with a real agent session (if this touches detection, state, the daemon, or the CLI)

## Notes for reviewers

The regression test deterministically models a root-directory signature
collision. On current main, the watcher misses the added and removed files.
With this change, it detects both correctly.

The picker was also verified using an isolated tmux server, isolated ccmux
daemon, and a seeded Codex session.

### DIFF
--- a/src/daemon/log-tree-watcher.test.ts
+++ b/src/daemon/log-tree-watcher.test.ts
@@ -208,6 +208,41 @@ describe("LogTreeWatcher (native)", () => {
     expect(unlinked).toContain(gone);
   });
 
+  it("reopens nested gates on a null event when directory signatures collide", async () => {
+    const project = join(root, "proj-a");
+    mkdirSync(project);
+    const gone = join(project, "gone.jsonl");
+    writeFileSync(gone, "x\n");
+    await startWatcher(1);
+
+    const evt = watcher as unknown as Internals;
+    evt.handleEvent(null);
+
+    const fresh = join(project, "fresh.jsonl");
+    writeFileSync(fresh, "x\n");
+    unlinkSync(gone);
+
+    // Model a filesystem whose timestamp granularity lets the rapid pair of
+    // mutations retain a cached child signature. Reopening only the root
+    // gates cannot discover either change below that still-closed child.
+    const current = statSync(project, { bigint: true });
+    const collidingSig = {
+      mtimeNs: current.mtimeNs,
+      ctimeNs: current.ctimeNs,
+    };
+    const projectNode = evt.rootNode.childDirs.get(project) as TreeNode & {
+      walkSig: typeof collidingSig;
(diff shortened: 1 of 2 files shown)


epilande (commented):
The `DirSig` comment is wrong. Measured on HFS+: 200 rapid creates in one directory produce exactly 1 distinct `(mtimeNs, ctimeNs)` signature, and the add+delete case collides. APFS gives 200/200. The comment conflates what triggers a ctime update with its stored resolution; ctime has the same resolution as mtime on a given filesystem.

The fix reopens only the root's gates, and that's the one level with no session files. Depths are `claude=1`, `copilot=2`, `codex=4`. On a null event the walk descends into cached children, and each child's colliding signature still swallows the change. The new test is at the root; the existing null-event test uses a `proj-a` subdir, which matches reality.

Let's clear the whole tree:

```suggestion
      // A null event carries no path-local evidence, so cached directory
      // signatures cannot safely close the gates. On filesystems whose
      // timestamp resolution is coarser than the mutation sequence (HFS+,
      // legacy ext4 with 128-byte inodes), a rapid add/remove pair can
      // leave a directory's (mtime, ctime) unchanged. Clear every cached
      // gate, not just the root: no built-in keeps session files at the
      // root (claude depth 1, copilot 2, codex 4), so a root-only reset
      // would miss every .jsonl that matters.
      clearSigs(this.rootNode);
      this.walk(this.root, this.rootNode, 0);
      this.sweep(this.root, this.rootNode);
      return;
```

With a helper next to `newDirNode`:

```ts
/** Reopen every cached gate under `node` for the null-event reconcile. */
function clearSigs(node: DirNode): void {
  node.walkSig = null;
  node.sweepSig = null;
  for (const child of node.childDirs.values()) clearSigs(child);
}
```

Cost isn't a concern: the null-event reconcile already stats every cached directory, since `walk` descends unconditionally and `sweep` always recurses.

Two more:

- Fix the `DirSig` comment in this PR. It now reads as measured-false directly above code asserting the opposite.
- Move the test's colliding node into a subdirectory.

Separate follow-up, not for this PR: the named-directory-event path runs the same gates and has the same hole.

LiadOz:
Fixed in `8203dad`. Null-event reconciliation now recursively reopens every cached directory gate. The regression uses a nested `proj-a` node and fails with the previous root-only implementation. I also corrected the `DirSig` documentation so timestamps are described as an optimization rather than a guarantee.

PR #133, opened 2026-08-12
fix: track Codex through code mode host

LiadOz:
## Summary

Ignore Codex's internal `codex-code-mode-host` process during agent discovery.
                                           
The host is a same-TTY child of the real Codex process and also matches the
Codex process rule. Wrapper collapsing therefore retained the internal host
and discarded the real Codex process, breaking authoritative PID-based      
session tracking.                                                                      

## Testing

- [x] `bun test` passes

## Notes for reviewers

The regression test fails on current main by selecting the internal host PID
instead of the real Codex PID.
                                                                                                                                                                              
Verified against live Codex 0.147.0 processes. An isolated daemon test also                                                                                                   
reproduced the same-TTY parent/host hierarchy, retained the real Codex PID,                                                                                                   
filtered the host PID, and rendered the resulting session in the picker.


epilande (commented):
Verified the diagnosis. `codex-code-mode-host` matches the unanchored codex rule via the word boundary at the hyphen, runs as a same-tty child, and `dropWrapperParents` evicts the parent through the child's ppid link. Marker reap path checks out too.

Missing from the docs: this is only reachable on the `ps`-tty platforms.

macOS uses `FD_TTY_DISCOVERY`, which harvests the tty from lsof fds 0/1/2. The host runs fully piped, so it resolves `tty: null` and `withTty` drops it before `dropWrapperParents` runs:

```
real codex 75899: fd0=/dev/ttys122  fd1=/dev/ttys122  fd2=/dev/null
host       79914: fd0,fd1,fd2 = all pipes  ->  no tty
```

Linux takes the tty from `ps`, which reports the inherited controlling terminal regardless of stdio, so the host survives and evicts. Resolver on those exact rows: Linux keeps `[79914]`, macOS keeps `[75899]`.

Two related facts:

- The host is code-mode-conditional, not per-session. 2 of 7 live codex processes had one. That plus the platform gate is why our 0.146 pass came back green.
- Anchoring `processMatch` to `/^codex$/` isn't a viable alternative, since ChatGPT.app's `codex app-server` basename is literally `codex`.

Let's scope the docs paragraph in `docs/agent-adapters.md`:

```suggestion
- Codex 0.146 also starts an internal `codex-code-mode-host` process in the user's project directory, conditionally, when code mode engages. Its basename matches the same unanchored process rule and it is a same-tty child of the real Codex binary, so ordinary wrapper collapsing would retain the host and discard Codex. `discoverAgentProcesses` filters that exact internal executable before wrapper collapsing (`isCodexCodeModeHostCommand` in `processes.ts`). The eviction is reachable only on the `ps`-tty platforms (Linux): on macOS `FD_TTY_DISCOVERY` harvests the tty from fds 0/1/2 and the host runs fully piped, so it resolves no tty and the `withTty` filter drops it first. The filter is defense-in-depth there, and load-bearing on Linux, where keeping the real Codex PID matters: `SessionStart` markers record the Codex hook runner's parent PID, and stale-marker cleanup compares that PID with process discovery; selecting the host would reap every live Codex marker after two scans and disable subsequent `PermissionRequest` / `Stop` marker updates.
```

Which platform did you verify on? On macOS the pre-fix daemon looks healthy, so I want to know whether the tracking breakage was observed or inferred.

LiadOz:
Updated in `ad42695`. Verification was on Ubuntu 24.04 with live Codex 0.147 processes. The parent/host process shape was observed directly, and the unpatched resolver behavior was reproduced in the regression fixture. The deployed daemon already had our local patch, so complete marker reaping was inferred rather than observed in production. The documentation now explains the Linux `ps`-TTY path and macOS defense-in-depth behavior.

PR #136, opened 2026-08-16
fix(daemon): remove killed sessions on confirmed exit

### LINKED ISSUE #135: [Bug]: Picker keeps killed sessions visible until the next scan
## Preflight

## ccmux version

1.3.0

## Agent

Multiple / not agent-specific

## Are hooks installed?

Not sure

## What happened?

After pressing `x` to kill a selected session in the ccmux picker and confirming the action, the agent process is terminated but the session row remains visible for a few seconds.

The row eventually disappears after the daemon's next reconciliation scan.

## Steps to reproduce

1. Start a supported agent in a tmux pane.
2. Run `ccmux picker`.
3. Select the active session.
4. Press `x` and confirm the kill action.
5. Observe that the agent process exits, but its row remains in the picker temporarily.
6. Wait a few seconds for the row to disappear.

## OS and tmux version

macOS, Darwin 25.5.0, tmux 3.7b

## Logs and additional context

Expected behavior: after the kill request succeeds, the session should disappear from the picker immediately, or as soon as ccmux knows the pane/process is gone.

Actual behavior: the UI keeps showing a stale session row until the daemon's next reconciliation scan, making it appear as though the kill action has not completed.

This appears to affect normal session kills from the picker and does not seem specific to one agent.

deferred:
## Summary

Remove a killed session's row daemon-side, once the process is confirmed dead, instead of leaving it on screen until the next reconciliation scan.

`handleKillSession` snapshots the row's pid, sends SIGTERM, then polls liveness (signal 0, 100ms, capped at 2s) and calls `removeSession` only on confirmed death. Removal happens in the daemon precisely so every attached client learns about it through the ordinary `session_removed` broadcast, rather than each one dropping the row locally and guessing whether the kill worked.

An already-dead process (ESRCH) removes the row immediately, with no wait. A process that outlives the cap **keeps** its row and reports the new additive `killed: false`, so a surviving agent is never hidden from the picker; the scan loop's liveness cleanup owns it from there. `ccmux kill` surfaces that case rather than printing a success it has been told is untrue.

Before removing, the handler re-reads the row and skips a *different* non-null pid. The snapshot fixes which process was waited on, but the row is addressed by id, and a pane-tracked id outlives the process it names (`createPaneTrackedSession` mutates in place when a new agent appears in that pane).

`handleRestartSession`'s inline 5s poll now shares the same `waitForExit` helper.

### Deliberately not covered

`kill-all` is a separate synchronous sweep with no wait and no removal, so bulk-killed rows still linger until the scan. Group kill is not the bulk path (the TUI fires per-row `/kill` concurrently), so it does get the immediate removal.

`waitForExit` is a third copy of a poll `worktree-prune.ts` already has as `stillAliveAfter`. Unifying them touches prune, kill, and restart, so it belongs in its own PR.

## Related issue

Closes #135

## Notes

Supersedes the original client-side approach in this branch (@deferred), which removed the row in the TUI on a successful kill request. Moving the removal into the daemon fixes it for every attached client at once, and gates it on the process actually being gone rather than on the request being acknowledged.

## Testing

- [x] `bun run typecheck`, `knip`, `bun run build`, `bun test` (4972 tests) all pass
- [x] Verified live against a real Claude session, on an isolated daemon and tmux server:
      `POST /sessions/:id/kill` returned `{"success":true,"killed":true}` in 1334ms and the
      row was gone from `/sessions` on the response, not at the next scan. `ccmux kill`
      printed the success line and exited 0 with the row removed.

The `killed: false` path (a process that outlives the cap) is covered by the test suite
against a real `trap "" TERM` process rather than live: producing a SIGTERM-immune agent on
demand is not something the harness can arrange honestly.

New `src/daemon/server.kill.test.ts` drives **real child processes** rather than a `process.kill` spy: a SIGTERM-able `sleep`, a `trap "" TERM` immune process (proving the 2s cap and that the row is retained), an already-reaped pid (ESRCH), and a row the reconciler re-points at a different live process mid-wait (proving the removal guard). Each asserts through to the actual `session_removed` SSE frame.


src/tui/App.tsx
```diff
@@ -1919,22 +1919,30 @@ export function App(props: AppProps) {
...
+          // and invites a retry.
           if (session?.trackingMode === "background") {
             store.actions.showToast("Stopping agent...");
+            return;
+          }
+          // The daemon's next scan will emit the authoritative removal event,
+          // but hiding a normal row now avoids making the user wait for it.
+          // Invoke-backed rows must remain until their invoker reports teardown.
+          if (session && session.originInvocationId === null) {
+            store.actions.removeSession(id);
```
epilande: nit: A 200 here only proves SIGTERM delivery. `handleKillSession` folds ESRCH into success and neither removes the session nor broadcasts, so if the agent survives SIGTERM (trapped handler, wedged event loop, often the very rows we kill), the removed row can't come back: `updateSession` drops unknown ids, `session_created` only re-fires on a visibility transition, and SSE won't reconnect against a healthy daemon. A long-lived sidebar would hide a live, possibly-waiting agent until restart (a popup picker heals on reopen). It's also client-local, so other surfaces still wait for the scan.

Not a blocker. The happy path converges fine and the trigger is rare. I'm thinking of a daemon-side follow-up: `handleKillSession` reuses `handleRestartSession`'s exit-poll, then `sessionManager.removeSession` on confirmed death so the `session_removed` broadcast reaches every surface, and a SIGTERM-surviving row correctly stays visible.

src/tui/App.tsx
```diff
@@ -1919,22 +1919,30 @@ export function App(props: AppProps) {
...
-          // invites a retry.
+          // A background row is removed only once the supervisor drops it from
+          // the roster, so without this toast `x` reads as having done nothing
+          // and invites a retry.
           if (session?.trackingMode === "background") {
             store.actions.showToast("Stopping agent...");
+            return;
+          }
+          // The daemon's next scan will emit the authoritative removal event,
+          // but hiding a normal row now avoids making the user wait for it.
```
epilande: nit: The 200 doesn't guarantee the process died (SIGTERM delivered; ESRCH is folded into success), so the scan's removal event is conditional. Let's not present it as a guarantee:

```suggestion
          // Once the process is gone, the daemon's next scan will emit the
          // authoritative removal event; hiding a normal row now avoids
          // making the user wait for it.
```

src/tui/App.test.tsx
```diff
@@ -1189,6 +1189,11 @@ describe("App kill/restart dispatch routing", () => {
       await killSelected();
       expect(calls.some((c) => c.url.includes("/sessions/s1/kill"))).toBe(true);
       expect(calls.some((c) => c.url.includes("/invoke/"))).toBe(false);
+      // The successful kill response removes normal rows immediately instead
+      // of waiting for the daemon's next reconciliation scan.
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      await setup.renderOnce();
+      expect(squish(setup.captureCharFrame())).not.toContain(squish("myapp"));
```
epilande: nit: This pins the new branch, but nothing pins the two preserved ones: deleting the `originInvocationId` guard, or just the background branch's early `return`, leaves the suite green, since the sibling tests assert only fetch URLs and toast text (toast state is separate from the sessions list, so a background row could be toasted *and* wrongly removed without any test noticing). Let's add the positive counterpart to two siblings:

```ts
expect(squish(setup.captureCharFrame())).toContain(squish("myapp"));
```

after the settle in the background-toast test, and the same (plus the settle + `renderOnce` this test already uses) in the Claude invoke-cancel test.

src/tui/App.tsx
epilande: This is where the edge lives (details in the review summary): once this row is removed locally, nothing can bring it back if the process survives SIGTERM. `updateSession` drops unknown ids (`store.ts:1489`), the daemon only re-emits `session_created` on a visibility transition (`server.ts:3524`), and SSE won't reconnect against a healthy daemon. The agent keeps running, possibly in `waiting`, with no row on this surface.

epilande (commented):
The diagnosis is right: the daemon never removes a killed session, so the row waits on the next scan.

One edge with removing the row client-side: the 200 only means SIGTERM was sent, not that the process died. If the agent survives the signal, the row is gone but the agent is still running, and nothing brings the row back (the store ignores updates for ids it doesn't have). It also only fixes the surface that pressed `x`; a sidebar still waits for the scan.

Let's fix it in the daemon instead: after SIGTERM, poll briefly for the process to exit (the restart handler already has this loop), then `sessionManager.removeSession` on confirmed death. The `session_removed` broadcast then updates every surface right away, and if the process survives, the row correctly stays. The guards here also become unnecessary, since invoke and background rows never reach that path in the daemon.

Would you be up for reworking this PR that way? If you'd rather not, no worries, I can pick it up in a follow-up instead, but I'd hold off on merging the client-side removal since the daemon fix replaces it.

epilande:
Went ahead and pushed the daemon-side version onto your branch (plus a merge of main) so this can land: after SIGTERM the kill handler polls briefly and removes the session on confirmed death, so the `session_removed` broadcast updates every surface right away, and a process that survives the signal keeps its row. The client-side removal is reverted since this replaces it, and there's a new `server.kill.test.ts` covering the kill path with real processes.

Merging once CI is green. Your diagnosis in #135 was spot on.

deferred:
Awesome, thank you! Loving your project, helps a lot in my daily work 👍

PR #140, opened 2026-08-17
fix(setup): honor PI_CODING_AGENT_DIR

deej-io:
## Summary
- Resolve Pi's agent directory from `PI_CODING_AGENT_DIR`, including Pi-compatible `~` expansion.
- Install the ccmux Pi extension under that directory and document the override.
- Add resolver and import-time regression coverage.

## Testing
- `bun test src/lib/config.test.ts src/daemon/adapters/pi/hook-adapter.test.ts`
- `bun run typecheck`
- `bun run build`
- `PI_CODING_AGENT_DIR=<temp> CCMUX_HOME=<temp> ./bin/ccmux setup --agent pi`

`bun test` still has pre-existing failures in this checkout: four cross-file plugin mock-load errors and the `tag shadowing` worktree-prune test timing out. The latter also reproduces in an untouched `main` worktree; the affected Pi and plugin tests pass in isolation.


epilande:
Thanks for the fix!

PR #141, opened 2026-08-19
fix: apply theme text to unstyled headings

ricardoroche:
## Summary
- apply the active theme foreground to the picker Sessions heading
- apply the same foreground to group and session preview headings
- prevent OpenTUI white fallback from bypassing configured light themes

## Verification
- bun run typecheck
- bun test
- rendered the picker in a detached tmux session and asserted the Espresso Paper header foreground is #4a4542

### DIFF
--- a/src/tui/components/GroupPreview.tsx
+++ b/src/tui/components/GroupPreview.tsx
@@ -235,7 +235,7 @@ export const GroupPreview: Component<GroupPreviewProps> = (props) => {
     >
       <box height={3} flexDirection="column">
         <box flexDirection="row" gap={1}>
-          <text>
+          <text fg={theme.text}>
             <b>{props.header.label}</b>
           </text>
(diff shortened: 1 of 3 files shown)


epilande (approved):
Verified beyond reading the diff: rendered the picker from both `main` and this branch under `catppuccin-latte` in isolated tmux panes and compared the raw `capture-pane -e` output. "Sessions" and the group preview heading flip from `38;2;255;255;255` to the theme's `text` color; the dimmed path is unchanged. Also confirmed the rationale in the `@opentui/core` bundle: the default fg really is opaque white (`RGBA.fromValues(1, 1, 1, 1)`), so an unstyled `<text>` bypasses the palette exactly as described.

`theme.text` is the token every sibling heading already uses; these three were the outliers. Good catch.

One note, not on this diff: the same fallback survives in `SessionItem.tsx` via `dimColor(ctx, undefined)`, so session row project names still render white. Tracking in #142.

PR #143, opened 2026-08-19
fix: apply theme text to session row project names

### LINKED ISSUE #142: Session row project names ignore theme text (OpenTUI white fallback)
#141 fixed the unstyled headings, but the same bug class survives in `SessionItem.tsx`, on the picker's highest-traffic text.

`dirnameColor()` at `src/tui/components/SessionItem.tsx:427-430` resolves to `dimColor(ctx, undefined)` for every non-active row, and `dimColor` passes the `undefined` through, so the project name falls back to OpenTUI's default fg: opaque white (`_defaultOptions.fg = RGBA.fromValues(1, 1, 1, 1)` in `@opentui/core`). Same at `:451`, where `baseColor={dimColor(ctx, undefined)}` styles the search-highlight text.

On a light theme (e.g. `catppuccin-latte`) this is white text on a light background. Verified with `tmux capture-pane -e`: the row dirnames emit `38;2;255;255;255` while the group headers next to them emit the theme's `text` color.

Fix is the same one-line change #141 made for `Header.tsx`: replace both `undefined` fallbacks with `theme.text`.

One design question to settle while fixing: the ternary at `:428-430` gives active-but-unselected rows `theme.text` and everything else `undefined`. Once both branches are `theme.text` it collapses. Under dark themes `theme.text` and white differ only subtly, so the distinction may have been decorative all along; if it was intentional, the non-active branch needs a different token (e.g. `subtext`) rather than `undefined`.

Also worth a sweep: any other helper with an optional color param that can pass `undefined` through to a renderable. A grep for `dimColor(ctx, undefined)` currently surfaces exactly these two sites.

epilande:
Fixes #142. Completes what #141 started: session row project names (the picker's highest-traffic text) still passed `undefined` through `dimColor()`, landing on OpenTUI's opaque-white default fg and ignoring the configured palette. On light themes that is white text on a light background.

## Changes

- `SessionItem.tsx`: both `undefined` fallbacks become `theme.text`, the dirname span and the search-highlight `baseColor`. The `dirnameColor()` helper is deleted; its ternary (active-and-unselected rows got `theme.text`, everything else white) dated to the v1.0.0 import and was visually indistinguishable on dark themes, so it collapses rather than survives with a substitute token. Dimmed rows still get `theme.border` through the untouched `dimColor` pass-through.
- `SessionItem.test.tsx`: regression tests asserting the actually rendered fg via `captureSpans()` (the pattern from `SessionItem.pr-color.test.tsx`; `captureCharFrame()` is text-only). Plain row expects `theme.text`, dimmed row expects `theme.border`. The plain-row assertion was verified to fail against the old code with received `[255,255,255,255]`.

## Verification

- `bun run typecheck` clean; full suite 4945 pass / 0 fail (2 new).
- Live render per AGENTS.md: picker launched from this branch in a detached tmux pane under an isolated `CCMUX_HOME` with `catppuccin-latte`, captured with `capture-pane -e`. Every row's project name now emits the theme's text color (`38;2;76;79;105`); a sweep of the frame finds zero readable white runs left. Before the fix, the same capture showed every non-active row dirname as `38;2;255;255;255`.
- Sweep for the same bug class: `grep` for helpers passing an optional color through as `undefined` surfaces no remaining sites in `src/tui`.

### DIFF
--- a/src/tui/components/SessionItem.test.tsx
+++ b/src/tui/components/SessionItem.test.tsx
@@ -1,9 +1,11 @@
 import { describe, it, expect, afterEach } from "bun:test";
 import { testRender } from "@opentui/solid";
 import { MouseButtons } from "@opentui/core/testing";
+import { RGBA, type CapturedFrame } from "@opentui/core";
 import { createSignal } from "solid-js";
 import { SessionItem, abbreviateTarget, alignText } from "./SessionItem";
 import { TickContext } from "../store";
+import { theme } from "../theme";
 import { mockEnrichedSession } from "./test-helpers";
 import { displayWidth } from "../utils/format";
 import { HANDOFF_BADGE } from "./session-columns";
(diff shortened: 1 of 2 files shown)


epilande:
Theme showcase (static): the picker rendered under Catppuccin Mocha and Catppuccin Latte with this fix applied, list only, against two throwaway test sessions. The Latte tile shows what this PR changes: session row project names render in the theme's text color. Before this fix they were painted opaque white, invisible on Latte's light base (see the capture evidence in the PR description). Static render, nothing driven.

ccmux-pr143-theme-showcase-1787123006

PR #146, opened 2026-08-20
feat(opencode): Surface question-tool waits as waiting/question

### LINKED ISSUE #137: [Feature]: Distinguish OpenCode question prompts
## Preflight

- [x] I searched existing issues and this has not already been requested.

## Problem / motivation

When OpenCode asks a user question through its question picker, ccmux does not reliably show that the session needs input. The session can appear as `working` or `idle`, so it is easy to miss an unanswered question.

## Proposed solution

Support OpenCode question lifecycle events and display an active question as:

- status: `waiting`
- attention type: `question`
- picker label: `question`

The state should remain waiting until OpenCode reports that the question was answered or rejected. A terminal fallback for the visible question picker would cover existing sessions or missed plugin events.

## Alternatives considered

Terminal-only detection can miss the picker during redraws and does not provide a reliable lifecycle signal. Treating questions as generic permissions loses the distinction between an answer request and tool approval.

## Relevant agent(s)

- OpenCode

## Additional context

OpenCode exposes `question.asked`, `question.replied`, and `question.rejected` events. Its question picker presents answer choices and waits for explicit user input.

epilande:
## Overview

Surfaces OpenCode question-tool prompts as `waiting` sessions with a `question` attention type, so an unanswered question picker shows up in the picker and notifications instead of sitting invisibly at `working`.

Resolves #137

## Motivation

OpenCode's `question` tool blocks the turn on a pending deferred exactly like a permission, but `session.status` stays `busy` for the whole wait and fires nothing at ask time (verified live on 1.18.15). The session pinned at `working` for as long as the picker was open, so it was easy to miss that an agent was waiting on an answer.

### Plugin (event path)

- **src/plugins/opencode/plugin.js** - Handles `question.asked` / `question.replied` / `question.rejected` (the unprefixed v1 names the plugin bus publishes) and writes a new `waiting_question` marker state, carrying the question text in `permission_context`. A missed reply self-heals on the next `session.status`. The rename-preservation guard now also covers `waiting_question`, since question asks trigger a title rename mid-wait.

### Daemon projection

- **src/daemon/session-markers.ts** - Widens the marker state union with `waiting_question`
- **src/daemon/adapters/opencode/aggregate.ts** - Both waiting states count toward the worst-of fold; the attention type follows whichever the newest waiting marker carries, and mixed permission+question waits set `ambiguousWait` like two permissions would
- **src/daemon/cascade-evaluator.ts** - `markerStatusState` projects `waiting_question` to `waiting`/`question`, so the canonical map cannot drift from the aggregate's

### Terminal fallback

- **src/lib/agents.ts** - A picker-footer rule (`matchAll` on `esc dismiss` + `enter `) covers a question already open when the plugin loads. The SDK has no `question.list`, same gap as `permission.list`. Anchors captured live on 1.18.15 and re-verified on 1.18.19; `esc dismiss` is rendered by the question picker alone across OpenCode's TUI, and the footer disappears with the picker so the match never lingers as scrollback.

### Docs & tests

- **docs/agent-adapters.md**, **AGENTS.md** - Event list and OpenCode caveats updated
- **src/plugins/opencode/plugin.test.ts**, **aggregate.test.ts**, **cascade-sources.test.ts**, **terminal-detector.test.ts** - Question lifecycle, mid-wait rename preservation, mixed-wait attention, projection, and a verbatim picker fixture

## Notes

- Question notifications deliberately carry no Reply action: Escape REJECTS the open question rather than cancelling back to the composer, so a typed reply has no safe abort key. The question text lands in the marker's `permission_context`, but nothing daemon-side consumes it yet; a richer notification body is a natural follow-up.
- Existing installs keep their older plugin until `ccmux setup --agent opencode` re-runs; until then the terminal fallback lifts the stale `working` marker via standard `upgradeOnly` mechanics.

## Breaking Changes

None

## Testing

- [x] `bun run typecheck` clean, full suite 4964 pass / 0 fail
- [x] Live e2e (isolated daemon + tmux server): idle -> `waiting`/`question` (question text in the marker) -> answered -> idle on the event path
- [x] Live e2e on the terminal-fallback path: plugin removed, zero markers, state derived from the pane alone on OpenCode 1.18.19
- [x] Recorded e2e posted below: both paths on OpenCode 1.18.19, idle -> working -> `waiting [Question]` -> answered -> idle


epilande:
E2E screen recording (sped up) against OpenCode 1.18.19: the question tool flips the picker row idle -> working -> waiting with the Question label, then back to working/idle after answering. First segment is the plugin event path (marker state `waiting_question`), second is the terminal fallback with the plugin removed (no markers, state derived from the pane alone).

https://github.com/user-attachments/assets/ccb81a39-e3bb-4dfc-86a5-117c8f3acac6

PR #154, opened 2026-08-22
feat(review): Review the whole branch with D

### LINKED ISSUE #149: feat(review): Branch-vs-base diff mode for d
## Problem

`d` reviews the working tree. Agents in worktrees commit as they go, so that diff is often empty, while the actual review question is what the branch changed relative to its base. Answering it today means leaving the TUI for a `git diff main...` incantation.

## Proposal

- Pass a diff target through to hunk (`hunk diff <base>`), which already accepts one.
- Record the resolved base ref at worktree creation and reuse it; fall back to merge-base with the default branch for worktrees ccmux didn't create.
- Context-sensitive default: worktree sessions open branch-vs-base, the main checkout opens working-tree. One key toggles modes either way.

## Non-goals

- In-TUI staging or hunk manipulation; review stays delegated to hunk.

## Delivery

Single small PR: base-ref persistence, target plumbing in the `d` flow, toggle key, help/footer text.

epilande:
Resolves #149

`d` and `D` are a fixed pair, not a default and a flip. `d` always reviews what is **uncommitted**, exactly as it did before this PR. `D` (Shift+D) always reviews everything the checkout has changed **since it forked**, which is the question worth asking about an agent working in a worktree: it commits as it goes, so its working tree is often empty while the branch is the whole point.

One key means one thing on every row, so nothing on screen has to announce which mode is about to run.

## How the base is known

- `ccmux spawn --worktree` records the resolved base as `branch.<name>.ccmux-base` in the repo config at creation (only for a branch it cut, never on reuse). Git owns the lifecycle: the record is deleted with the branch and renamed with it.
- Only a stable ref name is stored. A context-dependent expression (`HEAD`, `@{-1}`) would re-resolve at review time against the wrong checkout, so it is pinned to a SHA instead, or the write is skipped.
- `resolveMergeBase` tries the recorded ref alone first and only falls through to the default-branch heuristic when the record is absent or no longer resolves, so a worktree cut off a release branch reviews against the release branch, not `origin/main`. hunk still receives the merge-base SHA, never the ref name, for the documented behind-branch reason.
- The Worktrees panel's `d` gains the recorded-base accuracy for free.

`D` needs a fork point to have anything extra to show. A checkout carrying no commits of its own beyond its base has none, so `D` there falls back to the working tree on its own, which is why the fixed mapping needs no special case for row type. A main checkout with unpushed commits *does* have a fork point, and `D` showing those commits is the point.

## Keys

- Both terminal spellings of the capital are matched (name `"D"`, and name `"d"` with shift set), per the trap documented in WorktreesPanel. Ctrl+D/U preview scroll is unchanged.
- When `d` finds nothing uncommitted, the toast points at the other key. The hint is driven by an `empty` discriminant on the result, not by matching error prose, and is composed only at the picker's call site: the Worktrees panel and `ccmux review` have no `D` to point at.
- No toggle on the Worktrees panel: Shift+D is taken there (dirty-row opt-in). The row context menu keeps its single "Review diff" entry on `d`.

## Verification

- 5013 tests pass, 0 fail; typecheck and knip clean.
- New tests: base recorded on `-b` create / not on reuse / SHA for a context-dependent base; recorded base beats the heuristic, falls back when absent or unresolvable, skipped on detached HEAD, null when HEAD is the base; both terminal spellings of `D`; a main checkout with commits of its own reviews its branch, using the production session shape (`worktreeRoot` filled, not the `?? cwd` fallback).
- Live e2e: a worktree created with `--base release-9` stored the ref, reviewed against the release tip rather than the main merge-base, and fell back to the heuristic after the base branch was deleted.
- Recorded demo posted in the comments: `hunk`'s own header names the mode it is in, so the video shows the two keys producing genuinely different diffs rather than asserting it in a caption.


epilande:
Screen recording (sped up) of the `d` / `D` pair, built from this branch. Same row both times; only the key differs. (`tags` in hunk's header is the demo branch name: it prints the branch, then what it is diffing against.)

- `d` reviews **what is uncommitted**: header `tags working tree +3 -1`, one file.
- `D` reviews **everything the checkout changed since it forked**: header `tags 3551a38... +90 -4`, five files.

`D` is a superset of `d` rather than a separate view, because hunk compares the working tree against the base. That is why `src/format.ts +3 -1` shows up in both.

The worktree was created with `ccmux spawn --worktree`, so the base `D` compares against is the one this PR records at creation time (`branch.tags.ccmux-base` = `refs/heads/main`). On a checkout carrying no commits of its own beyond that base there is no fork point, so `D` shows the working tree, the same as `d`. A main checkout with unpushed commits does have one, and `D` shows those too.

Recorded in an isolated environment (its own daemon, port, tmux socket, and config dir), so the rows on screen are the demo's own sessions rather than a live fleet.

https://github.com/user-attachments/assets/ab515919-45dd-41e3-8725-b13806b4d9ce

PR #155, opened 2026-08-22
feat(spawn): --pr and --issue flags

### LINKED ISSUE #150: feat(spawn): --pr and --issue flags
## Problem

Putting an agent on a PR or issue takes several manual steps (fetch, branch, worktree, file setup, launch), and the main checkout is often occupied by live sessions. There is no one-line path from "this PR/issue needs work" to a tracked session in an isolated worktree.

### `ccmux spawn <agent> --pr <n>`

- Resolve via `gh pr view --json`; fetch `pull/<n>/head`; branch named from the head ref.
- Writable fork PRs, matching `gh pr checkout` semantics: set `branch.<name>.remote`/`.pushRemote` to the fork URL and `branch.<name>.merge` to `refs/heads/<headRefName>`, so pushing fixes and `git pull` both work.
- Then the existing worktree creation path (naming, locking, file setup).
- Optional `--prompt`, seeded with PR title and URL.
- If the PR's branch is already checked out in another worktree, refuse and point at it (git refuses the fetch anyway; the error should be ours).
- Directory naming must not collide with `.claude/worktrees/pr-<n>` (Claude Code's own fetch-only PR worktrees).

### `ccmux spawn <agent> --issue <n>`

- Normal spawn-from-base worktree; branch and worktree named from the issue; prompt seeded with title and URL only (agents fetch the body themselves).

## Non-goals

- A standalone no-agent checkout verb; this is spawn-scoped by design.
- Auto-syncing the worktree with the PR afterwards; freshness stays `git pull` in the worktree.

## Delivery

Single PR covering both flags; they share the gh metadata and naming plumbing.

epilande:
Resolves #150

One line from "this PR/issue needs work" to a tracked session in an isolated worktree:

```bash
ccmux spawn claude --pr 154     # worktree on the PR's own branch, tracking set up, push just works
ccmux spawn claude --issue 150  # spawn-from-base worktree named after the issue, prompt seeded
```

## --pr

- Resolved via `gh pr view` (new `gh-spawn-source.ts`: injectable runner, explicit env so tests can PATH-stub gh, 15s timeout since this sits in the spawn request path). Fetches `pull/<n>/head`, cuts the local branch as the PR's own head ref so `git push` works out of the box.
- Writable fork PRs, matching `gh pr checkout` semantics: `branch.<b>.remote`/`.pushRemote` = the fork's clone URL, `branch.<b>.merge` = `refs/heads/<headRefName>`. Same-repo PRs track `origin`.
- `branch.<b>.ccmux-base` = `origin/<baseRefName>` (fetched first, verified before writing), so the picker's `D` branch review shows the PR's actual diff against its target branch.
- Worktree named `pr-<n>-<head-ref-slug>` with the prefix budgeted inside the 40-char cap; by construction never collides with Claude Code's own fetch-only `pr-<n>` directories.
- Refusals, all before anything is written, all in ccmux's own words: PR not OPEN (state named), branch already checked out in another worktree (path named), a same-named local branch that is not this PR (`branch.<b>.merge` is the evidence). A branch that IS the PR is reused fast-forward-only (`git fetch` refspec without `+`); divergence refuses and leaves it untouched. Nothing is ever force-updated or reset.

## --issue

Ordinary spawn-from-base worktree (`--base` works as usual), named `issue-<n>-<title-slug>`, prompt seeded with title and URL only.

## Implementation notes for review

- `createWorktree` gains an optional `branch` override (directory name and branch decoupled for exactly this caller); the lock, numbering, occupied-path refusals, and file setup stay shared.
- `preparePRBranch` runs the fetch and branch decision under the repo lock and releases it before `createWorktree` takes the same lock (`withRepoLock` is not reentrant; nesting deadlocks). The window between the two fails loudly via `git worktree add`.
- The seeded prompt is deliberately never threaded into `createWorktree`: `resolveWorktreeName` prefers a prompt over a derived name, so it would silently rename the worktree after the PR title. Pinned by a test.
- Prompt seeding is skipped for a custom agent with no `promptCommand` and no user `--prompt`, so the checkout still happens instead of a refusal over a prompt nobody asked for.
- `ccmux-base` is fail-soft: if `origin/<base>` cannot be made to resolve, the key is omitted rather than failing the spawn.
- CLI guards against an old daemon silently dropping the flags (200 with no `worktree` in the response errors instead of leaving the agent in the wrong directory).
- Known v1 limitation: the head fetch assumes the remote is named `origin`; a differently-named remote gets a loud, accurate error.
- Merged with #154, which added the `branch.<n>.ccmux-base` record. Its create-path call keyed the record on the worktree DIRECTORY name, which is exactly what the `branch` override here decouples, so it now passes the branch. Pinned by a test watched failing first.
- That keying fix then needed a second one, because it ACTIVATED a value that had been harmlessly orphaned. `creation.base` on the PR path is the fetched PR-head SHA, the branch's own tip, so once the record landed under the key the picker reads, `resolveMergeBase` computed a merge-base equal to HEAD and returned `null` - and `null` SETTLES the lookup rather than falling through, so `D` silently degraded to a working-tree review, which on a freshly checked-out PR branch shows nothing at all. `configurePRBranch` masks that whenever it has a base to overwrite with, so only two paths reached it: an unresolvable base ref, or a failed tracking write. The PR path now opts out of the create-path record (`recordBase: false`), leaving `configurePRBranch` the sole writer on that path and key-ABSENCE as what makes the heuristic fallback run. Explicit flag rather than inferred from the `branch` override, so a future override caller that does want the record is not silently denied it.

## Verification

- 5096 tests pass, 0 fail; typecheck and knip clean. New: 26 gh-source unit tests, 13 worktree-create (override + name budgeting, real git), 11 server e2e against a real bare origin with a real `refs/pull/7/head` and a PATH-stubbed gh (both fork and same-repo config variants, both refusals, the prompt/naming trap).
- Live, isolated daemon: `--pr 154` correctly refused (branch checked out in another worktree, path named); fork tracking config verified against a real open cross-repo PR in a throwaway clone (head SHA byte-identical to `git ls-remote refs/pull/<n>/head`, remote/pushRemote = fork URL); `--issue` produced a 40-char-capped worktree name and the seeded prompt. Nothing was created in this repo; all test state torn down.


epilande:
Screen recording (sped up) of `ccmux spawn claude --pr 155`, driving the build from this branch's own head rather than an installed ccmux. The opening frame shows which binary is under test and that it is at `5caf4db`.

The proof is git state on camera, not a caption:

- `git ls-remote origin refs/pull/155/head` and the created worktree's `git rev-parse HEAD` print the same sha, four rows apart in one uncropped frame.
- `git status -sb` reads `## feat/spawn-pr-issue...origin/feat/spawn-pr-issue`, so `git push` goes to the PR without further setup.
- `branch.feat/spawn-pr-issue.ccmux-base` is `origin/main`, the branch the PR targets, which is what the picker's `D` review diffs against.
- `git worktree list` shows the directory named `pr-155-feat-spawn-pr-issue` while the branch is `feat/spawn-pr-issue`. Deliberately decoupled: the directory stays legible while the branch stays pushable.

Happy path only. The failure path the last commit fixes (an unresolvable base ref must leave NO review base rather than a poison one) is covered by tests, not filmed.

Recorded in an isolated environment with its own daemon, port, tmux socket and config dir, against a throwaway clone, so the single row on screen is the demo's own session. The speed-up partway through the first command is a pty artifact of `gh` writing JSON to a tty under the recorder, not ccmux.

https://github.com/user-attachments/assets/d5be0b84-b25e-4ea0-ab63-d24ca0d42d42

Issue #151, opened 2026-08-22
feat(tui): Open-PR list in the Worktrees panel


epilande:
## Problem

`spawn --pr` (#150) requires knowing the PR number, which means a trip to GitHub. The Worktrees panel already shows every checkout a repo has; it should also show the PRs you could check out.

## Proposal

- New section in the Worktrees panel listing open PRs: number, title, author, draft state, CI status. Two-phase like the rest of the panel: instant paint, `gh` data merges in by number.
- **Enter** on a PR pre-fills the new-session dialog as a new `NewSessionShape` ("worktree cut from a PR head"), reusing the existing dialog, policy functions, and row budget.
- Rows for PRs already checked out locally are marked and jump/prefill to the existing worktree instead.
- The prune scan learns "PR merged/closed" as a removal reason (scan side only; `worktree list` stays local-only per its contract).

## Non-goals

- An issues list; issues stay CLI-only via `--issue`.
- Any GitHub write operations from the panel.

## Delivery

Two PRs, mirroring how the panel itself landed: main PR (PR list section, dialog shape, both policy functions, `planDialogRows` counts, tests), then a polish PR after live use. Depends on #150.

epilande:
Scoping note before implementation: the fourth bullet ("the prune scan learns 'PR merged/closed' as a removal reason") is already shipped and is being dropped from this issue's scope. `PRUNE_REASONS` (`src/daemon/worktree-prune.ts`) already contains `pr-merged` and `pr-closed`; `reasonFor` applies them in precedence order, `detailFor` words them, and `branchDeletionFor` gives `pr-merged` a force-delete while `pr-closed` keeps the branch.

The related gap that is NOT closed, recorded here rather than acted on: `spawn --pr` records `branch.<b>.ccmux-base` but not the PR number, so the scan re-derives the PR by branch name plus tip SHA. An author force-pushing after merge breaks that match and the worktree is never classified. Closing it would mean recording `branch.<b>.ccmux-pr` at spawn and preferring a lookup by number.

Two decisions on the remaining bullets, both settled by existing precedent in the code:

- **Instant paint** has no local source for open PRs, unlike the worktree list. The section renders as a header plus a pending state and the `gh` answer merges in as an independent third phase, under the same generation guard and degrade-don't-fail rule as the prune scan. Seeding from the prune scan's cached PRs was considered and rejected: it shows a set biased toward branches sessions happen to sit on, which then visibly changes when the real list arrives.
[...]

PR #158, opened 2026-08-23
feat(worktrees): Open-PR view in the Worktrees panel

epilande:
Part of #151. This is PR 1 of the two-PR delivery the issue describes; the polish pass (an optional `ccmux worktree pr list`, section collapse) stays owed.

`spawn --pr` (#150) needs a PR number, which means a trip to GitHub. The Worktrees panel already knows every checkout a repo has, so it now also lists the PRs you could check out.

## What lands

- **The panel gains a second view.** `h`/`l` switch between Worktrees and Pull Requests; `Tab` still switches scope (this repo / all repos). The two are orthogonal axes: "this repo's PRs" and "all my PRs" are both real questions.
- The PR view lists a repo's open PRs: number, title, author, draft state, review decision, CI status.
- **Enter** on a PR row opens the new-session dialog in a new mode that cuts a worktree from the PR head. The daemon derives the name and base from the PR, so the dialog shows neither a Where nor a Name row.
- A PR whose head is already checked out locally is marked, and Enter routes to that worktree instead of cutting a second one.
- `r` refreshes the list, on every phase.
- `GET /prs`, behind a per-repo cache that is also a lock.

## Why a view and not a section

The first cut appended PRs as a third section at the bottom of each repo group. Live testing killed it, for two measured reasons.

**The per-repo header did not scale.** Drawing `open PRs · N` under every group costs one line per repo. On a 13-repo machine that is 13 lines of a ~45-line viewport spent on a header that mostly answers `0`. Worse, the repo you actually work in is the one you can never see: with 8 worktrees its group starts near the fold and its PR rows are below it before scrolling begins.

**Reordering could not fix it.** Phase 3 is a network round trip that lands after phase 1 has painted, so PR rows above worktree rows would shove already-visible rows down mid-interaction. That is the layout shift the panel's pending-rides-the-header idiom exists to prevent. Within a single list, the bottom is the only sound position, which is why reachability needed a second view rather than a better sort.

The tab line turns that cost into an asset: 13 dead header lines become one live `Worktrees │ [l] Pull Requests · 7`, which is also the discoverability the buried section never had.

## Three smaller decisions

**Zero is an answer in one view and noise in the other.** A repo with no open PRs gets no line at all in the Worktrees view, and a dim `no open PRs` under its header in the PR view, where zero is the answer the view exists to give.

**A whole-request failure is stated once.** When the request itself falls over, the cause is identical for every repo, so it prints one line and the groups are dropped. Thirteen bare headers with nothing under them is the same noise in a different shape. Per-repo failures keep per-repo lines.

**The view key rides the inactive tab, not the hint line.** The hint line drops `l PRs` at 80 columns, which is the most common default terminal width, so the key would have been invisible exactly where most people run. On the tab line it survives a four-rung degradation ladder down to `[l] PRs` in 7 columns; the key is the invariant, and the active tab's own label is dropped before it.

**"Already checked out" is decided by `headRefOid` against the local branch tip, never by branch name.** Name matching walks into the namesake case `selectPRForBranch` already documents, where `--head patch-1` on a large repo answers with many PRs from many forks. Where either side does not resolve, the row is left unmarked rather than guessed. Verified in both directions against a live daemon.

## Safety

`x` acts on the selection, not on the cursor, so a PR view that left the removal keys live would confirm deletions for worktree rows that are not on screen. `space`, `a`, `A`, `D` and `x` are gated on the view; `x` says `removal lives in the worktrees view: h` rather than failing silently. Covered by a test and verified live.

## Out of scope

No issues list (issues stay CLI-only via `--issue`), and no GitHub writes from the panel. The prune scan already learned `pr-merged` / `pr-closed`, so the issue's fourth bullet was already shipped and is not re-implemented here.

## Verification

`bun test` 5197 pass / 0 fail, typecheck, knip and build clean.

Beyond the suite, in a real terminal against a live daemon: both views at both scopes, the tab line and hint line swept at 200/110/90/80/70/60/44/34 columns in both views, the SHA-not-name rule in both directions, the checked-out marking, the whole-request failure collapsing to one line, and the removal-key gate.

## Known, deliberately not fixed here

`readUpstreamStates` (`worktree-git.ts`) has the same `%(refname:short)` tag-shadowing bug this PR fixes in `readBranchTips`. Fixing it makes `upstream-gone` newly reachable for a collision-named branch, which is a behaviour change in the path that removes directories and deletes branches, and it should not ride in as a drive-by. Filed separately.


epilande:
Surface demo of the open-PR section: the surface is shown, not driven through agent states. Screen recording, sped up.

Live data, no stub. The listed row is this PR, and `checked out in pr-list-151` is a `headRefOid` match against its real head.

What to look at:

- `open PRs · 1` under the worktree list, and the title reading `15 worktrees` rather than 16: PR rows are listed, not counted as worktrees.
- The `pr-list-151` worktree row carrying `PR #158 open`, the same link from the other direction.
- Enter on a checked-out PR opening that existing worktree rather than cutting a second one, and Esc returning to its row.

Not shown here, verified separately in a terminal: cutting a worktree from a PR that is not already checked out, and the pending and unavailable header states.

https://github.com/user-attachments/assets/101b348c-9ad1-4e6b-b4cc-6e9fee26723a

epilande:
Screen recording of the reworked panel, sped up. It is a surface demo: the panel is shown by scripted navigation, not driven through live agent state.

What it walks through:

- The Worktrees view, scoped to ccmux. The tab line under the title reads `Worktrees │ [l] Pull Requests · 2`, with the key sitting on the inactive tab.
- `l` opens the Pull Requests view. #158 carries `checked out in pr-list-151`; #159 shows its branch and author and no CI phrase, since a fork PR's checks sit at `action_required` and there is nothing to render.
- `h` returns to Worktrees. No per-repo `open PRs` divider anywhere in the list, which is the line this PR deletes.
- The safety gate: a removable worktree selected with `space` in the Worktrees view, then `l`, then `x`. It flashes `removal lives in the worktrees view: h` and opens no confirm. The selection survives the view switch, and the hint line still counts `x remove 1`.
- Width: at 80 columns the tab line keeps `[l] Pull Requests`; at ~34 columns it degrades to `[l] PRs` with the key intact.

The panel is scoped to this repo for the whole recording, so the all-repos view is not part of it.

https://github.com/user-attachments/assets/f61d2f2c-e07c-4ee2-9ab2-5303e9ea7f6c

epilande:
Supersedes the recording above, which was made before the `[l]` badge came out of the tab line.

What the panel does now:

- The tab line reads `Worktrees │ Pull Requests · 2`. The view key lives in the footer (`l PRs`, `h worktrees`), not on the tab.
- `h`/`l` switch views, and each view keeps its own cursor: the worktrees cursor returns to `ccmux-pr-154`, the PR cursor to #158.
- #158's row is marked `checked out in pr-list-151`, so Enter on it opens the dialog on that existing worktree instead of cutting a second one. Enter on #159, which is not checked out, opens the same dialog in PR mode, with no Where or Name row.
- The PR view's footer drops the removal keys, because `x` acts on the selection rather than on the cursor row.

Screen recording, sped up. Recorded against a daemon scoped to this repo, so the lists are ccmux's own 15 worktrees and its 2 open PRs.

https://github.com/user-attachments/assets/5773cf07-1eab-4739-a1e8-53fa0b4adac4

PR #159, opened 2026-08-23
feat(columns): Render the last prompt as a wrapped multi-line block via `promptLines`

aiwen324:
## Overview

Adds `promptLines`, which renders a session's last prompt as a wrapped multi-line block of up to N lines between the identity row and row 2. `0` (the default) is off. `sidebar.promptLines` overrides it for the rail.

## Motivation

The `prompt` column is one truncated line, which for a real prompt is a handful of words. In the sidebar, where there is no preview pane, that is often not enough to tell sessions apart. A capped, wrapped block shows enough of the prompt to identify the session without giving it unbounded height.

```bash
ccmux config set sidebar.promptLines 3
```

```
 ● Sessions (3)
 ▼ 0 (3)
 ◑ Fix pagination off-by-one
   Retry the upload twice before
   surfacing the error, and log
   the attempt count so we can …
 myapp:main                   cc  42s
 ● Add dark mode toggle
   ~/code/parser              cc   2m
```

### Variable row height

- **grouping.ts** - `SessionHasSubtitle` (1 line or 2) becomes `SessionLineCount`, returning the real count, so `itemVisualHeight`, `toVisualLine` and `scrollTarget` work at any height
- **SessionList.tsx** - Wraps and measures the block next to the shared layout, so row heights and scroll positions can't disagree. The wrapped block is memoized per session on `(lastPrompt, width, max)`, so an unchanged session hands the row the same array and the inner list does not re-key on every tick
- **SessionItem.tsx** - Draws the lines it is handed verbatim and reports its height as their length. Turning the block on removes the `prompt` cell from both rows rather than printing the text twice; `promptDisplay: "off"` (<kbd>p</kbd>) still hides it

### Wrapping

- **format.ts** - `wrapText` moves here from `NewSessionDialog` (importing it back would have been circular via `agentColorFor`). `wrapToLines` joins it, capping the block and ellipsizing the last kept line so a clipped block never reads as a complete one

### Config

- **preferences.ts**, **config.ts** - `promptLines` and `sidebar.promptLines`, validated as an integer `0..MAX_PROMPT_LINES` (10). The key is also offered by shell completion
- **picker.ts**, **sidebar.ts**, **store.ts**, **App.tsx** - Carry the value into the TUI
- **README.md** - Documents the option, and notes that the preview pane already shows the selected row's full prompt, so the block mainly earns its keep in the sidebar

## Breaking Changes

None

## Testing

- [x] `bun run typecheck`, `bun test`, `bun run check:unused` pass
- [x] `prompt-block.test.tsx` measures the height invariant directly: a second session's identity row moves down by exactly the wrapped line count at caps of 1, 2, 3 and 5; no-prompt sessions gain no lines; a 400-word prompt stays capped; the prompt does not print twice; the wrapped block is reference-stable across reads
- [x] `wrap-to-lines.test.ts` covers capping, the ellipsis marker, width never exceeded including the marker, wide glyphs (CJK), empty input and a zero cap
- [x] Picker and sidebar rendered in a detached tmux session with the block on

---

The `title` column that was originally part of this PR is deferred to a follow-up as a `summary` column, per the review above.


epilande (commented):
Both of these are worth having, and splitting them makes sense. `promptLines` is the closer of the two to ready, so that's the one to land first.

Some data on the column first, since it changes what the column should be doing. Comparing pane titles against `lastPrompt` across 35 live sessions here: the title identified the session on 31 of them, the prompt cell on 20. The prompt cell loses to task-notification payloads, relayed messages from other agent sessions, post-compaction stdout, and short continuations like "commit and push". That's a better argument for the column than the description makes, and a reason to work through the details rather than merge it quickly.

### Naming

Could we call it `summary` rather than `title`? Next to `status`, `project`, `branch` and `agent`, "title" reads like something you set rather than something the agent generates. It's also the natural name if ccmux ever grows a user-assigned session name.

The `EnrichedSession` field can stay `paneTitle`, since that is what it holds. Only the column name changes.

### `promptLines`

The height claim holds. `itemVisualHeight`, `toVisualLine` and `scrollTarget` are imported only by `SessionList`, all three read the same `sessionLines` closure, and the row draws the array it's handed without re-wrapping. `WorktreesPanel`'s scroll math is separate and never touches session rows. The wrapping is display-width correct, and the tests measure the invariant rather than the rendered text.

1. `promptBlock` isn't memoized, so `toVisualLine`'s loop re-wraps every preceding session on each call, and each read returns a fresh array that re-keys the inner `For`. Fine at ten sessions, measurable at forty. Let's memoize on `(lastPrompt, width, max)`.

2. `<For each={Array.from({ length: rowHeight() })}>` over identical `undefined` values wants `<Index>`.

3. Let's mention the preview overlap in the README. `Preview.tsx` already renders `lastPrompt` untruncated for the selected row, so the block mainly earns its keep in the sidebar.

One on our side, not yours: `stripCommandMarkup` doesn't handle task-notification payloads or relayed session messages, and doesn't strip ANSI from the `<local-command-stdout>` branch, so a post-compaction row paints a raw escape byte today. That's a good part of the noise above, and a wrapped block makes it more visible rather than less. We'll fix that separately.

### `summary` column

Here's a survey so the per-agent behaviour isn't guesswork. Every installed agent on an isolated tmux server, one minimal turn each, capturing `#{pane_title}` at baseline, at the idle prompt, mid-turn and after:

| agent    | idle                         | while working                                 | after a turn                                | what it is                                        |
| -------- | ---------------------------- | --------------------------------------------- | ------------------------------------------- | ------------------------------------------------- |
| claude   | `✳ Claude Code`             | `✳ Claude Code`                              | `✳ Say hello`                              | generated summary                                 |
| copilot  | `GitHub Copilot`             | `say hello and nothing else - GitHub Copilot` | `Implement Hello Function - GitHub Copilot` | generated summary, app name suffixed              |
| cursor   | `Cursor Agent`               | `Cursor Agent`                                | `Hello Bot`                                 | generated summary, app name before the first turn |
| omp      | `π > probe-omp-w5`           | `π ⠧ Say hello and nothing else`              | `π > Say hello and nothing else`            | session title                                     |
| codex    | `probe-codex-x7`             | `⠏ probe-codex-x7`                            | `probe-codex-x7`                            | cwd basename                                      |
| pi       | `π - probe-pi-z3`            | unchanged                                     | unchanged                                   | cwd basename                                      |
| gemini   | `◇  Ready (probe-gemini-t6)` | `✦  Working… (probe-gemini-t6)`               | `◇  Ready (…)`                              | state plus cwd, padded to 82 chars                |
| opencode | `OpenCode`                   | `OpenCode`                                    | `OpenCode`                                  | static app name                                   |

Four set something worth showing. Codex, pi and gemini set the cwd basename, which is already the `project` column, and gemini adds its own state, which is the `status` column. Copilot, cursor and opencode all sit at their own app name until something happens, so that string wants treating as absent rather than as content.

That's what makes filtering awkward. tmux seeds `pane_title` to the hostname, so the cell currently renders the machine's hostname for anything that sets nothing. The README says those agents "leave the cell empty", which isn't the case, and `pane-title.test.ts` locks the behaviour in: `expect(paneTitleText("Mac")).toBe("Mac")` under "keeps a title that has no glyph", where `Mac` is a hostname.

This probably wants a per-agent rule on `AgentDef` rather than a filter. A filter needs hostname detection, a cwd-basename comparison, an agent-name comparison and empty handling, and still misses gemini's `◇  Ready (dir)`, which is none of those. Once one agent needs its own pattern there's a per-agent table either way. It also matches how ccmux handles other per-agent differences (`terminalRules`, `errorRules`, `readyPattern`).

Only the Claude entry needs filling in. Default the rest to no summary and we'll add copilot, cursor and omp from the survey above. Those are also why the rule wants to be more than a leading-glyph strip: copilot suffixes ` - GitHub Copilot`, omp prefixes `π >` or `π ⠧`, and cursor needs neither but wants its idle `Cursor Agent` read as empty.

The rest:

1. `SessionItem.tsx:655` has a `// ponytail:` comment describing a real defect: summary and prompt on the same row both claim `maxPromptLen` and overrun each other. The README's example config hits it, since the default `promptDisplay: inline` flattens `prompt` onto row 1. The summary then gets clipped by flexbox with no ellipsis, so a truncated cell reads as a complete one. Could we split the budget rather than document the collision?

2. `maxPromptLen`'s row selection wants to be per-field rather than per-flex-class. With summary on row 1 and prompt on row 2 (the only layout the sidebar uses, since `applyPromptDisplay` forces `row2` there), `rowHasFlexText(cols.row1)` is true, so the row 2 prompt budgets against row 1's siblings and loses its leading indent. `field !== "prompt"` was right for those two call sites specifically. The rename works everywhere else.

3. Empty titles want treating as absent as well. Gemini emits `""` briefly during boot.

4. `paneTitle` is enrichment-only, so a change doesn't trigger an SSE broadcast on its own. The picker re-inits on every open so it stays current, but the sidebar is long-lived and will hold a stale summary on an idle row. Adding it to the tracked-field comparison in `sessions.ts` would cover that, comparing the normalized value rather than the raw one: codex and omp rewrite the title on every spinner frame, and tracking the raw string would broadcast on every scan tick for their working sessions.

Both PRs touch `SessionItem.tsx` and `session-columns.ts`, so they'll conflict. Landing `promptLines` first and rebasing this one on top means the budget split only gets written once.

epilande:
I'd like to land this soon, so I'll start pushing to this branch to carry the review forward. Feel free to jump back in anytime. Plan: reduce this PR to `promptLines` and merge it with you as author, then do the `summary` column as a follow-up PR.

aiwen324:
Thanks! Please go ahead! Sorry I am distracted by other stuffs recently.

epilande:
No worries! Thanks for responding. I'll get it in 🫡

epilande:
ccmux-verify-pr surface demo (prompt block on picker and sidebar, build 4cb146e). Shows (not driven): src/tui/components/SessionList.tsx, SessionItem.tsx, session-columns.ts, prompt-block-cache.ts, src/tui/utils/grouping.ts, format.ts. Config-gated surface with `promptLines: 3`: the wrapped block with its ellipsis cap, the yield back to the one-line prompt cell during search, the `p` toggle, and the narrow-rail yield in the sidebar. Recorded against an isolated daemon with three throwaway sessions; sped up.

Picker:

https://github.com/user-attachments/assets/6d7f6a42-c10f-4308-ac82-6f9ba29ff7e5

Sidebar:

https://github.com/user-attachments/assets/ac21170e-3926-4d40-89fb-c86524175a50

aiwen324:
Thanks! I'll close this PR.

epilande:
@aiwen324 why did you close the PR? I was about to merge it 😅

aiwen324:
Oops, my bad 😅 I thought you’d already landed the feature in another PR, so I closed this one to clean up. Thanks for reopening it and carrying this forward!

PR #161, opened 2026-08-24
feat(worktrees): Draw the view tabs as clickable chips

epilande:
Follow-up to #158. The panel's view line was two labels and a `│`. It now draws both views as chips, and both take a click.

```
 Worktrees   Pull Requests · 13
```

Exactly one chip is filled, on `border`, the same ground `ContextMenu` uses for its highlighted item, so a filled block already means "this one" elsewhere in the TUI. The other sits on the panel's own `base` with its label a step dimmer. Hovering the inactive chip brightens its label rather than filling it: a fill that appears before the click lands reads as decoration, not as a control.

## Why one fill and not two

The first cut filled both, `surface` for the inactive chip and `border` for the active one. That is not portable. `border` is darker than `surface` in **dracula** and **rose-pine**, which draws the active chip recessed and the inactive one raised, and a user can override the palette arbitrarily besides. With a single fill there is nothing to invert.

It also stays one line. A rule beneath the strip would read more like a browser's tabs and costs a row of list, which is the tax this panel exists to refuse.

## What changed structurally

The builder returns a `ViewTab[]` instead of a flat segment list, because a chip is one box (that is what carries a background and a click at all) and a flat list cannot say where one chip ends and the next begins. Two details are load-bearing:

- A chip's padding lives in its **segments**, not as box padding, so its measured width is exactly its text and the degradation ladder cannot run four columns over what it thinks it is using.
- The gap between chips is its **own element**, for the opposite reason. Padding inside a background paints the gap in the chip's colour and fuses both tabs into one block.

Below the ladder's last rung, `fitTabs` walks chips left to right and drops one it cannot fill rather than painting an empty block of colour where a label belongs. The old flat version left the separator dangling there, which was odd but invisible; a background makes the same shortfall visible.

A click routes through `switchView`, not a bare `setView`, so clicking the chip already showing is a no-op.

## Verification

5218 pass / 0 fail; typecheck, knip, build clean.

Mouse is testable here, which it wasn't before: `createMockMouse` drives real clicks under `testRender`, and `CapturedSpan` carries `bg`, so the fill is assertable. All four new render assertions were mutation-checked (break the fill, the click, the wiring, the hover, in turn, each test fails). That caught a false pass: a helper matching `Worktrees` finds the panel **title**, not the tab line, so one click test was passing while clicking a cell no tab occupies.

Clicked live in a real terminal at 120, 40 and 28 columns: the fill moves, the footer changes to the target view's keys, and the ladder still degrades to ` Worktrees   PRs · 13 ` without wrapping.

The demo video on #158 shows the old tab line and is now stale.


epilande:
Header folded into one line, following the same thread: the panel said `Worktrees` three times across two rows (title, first chip, and the title's `42 worktrees` suffix).

```
before          Worktrees · 13 repos · 42 worktrees
                 Worktrees   Pull Requests · 13

after           all repos   Worktrees 42   Pull Requests 13
```

The duplication was structural rather than careless. Tabs name the views, so a title standing over them can only repeat whichever one is showing. What the title carried that the tabs did not is the **scope**, so that is the half kept, and the header is now three zones: the scope lead, the chips, and a muted tail for the removal notice and the scan spinner.

Two decisions worth calling out:

**The lead never comes from the loaded rows.** It reads the scope flag and `props.repo`, so it is right in the first frame and flips in the same frame as the `Tab` that changed it. A lead that arrived with the data would slide both chips rightwards mid-open, and a chip is a click target. The one exception is a lone repo, which `showsGroupHeaders` draws no header for, so an unscoped panel that got one repo back names it here or nowhere.

**The `·` between a label and its count is gone.** In this TUI that dot divides peers (`9 untracked · 1 waiting`); gluing a count to the label it belongs to made one fact read as two, and it is where two of the three `Worktrees` came from. It survives in the tail, where the things either side of it really are peers.

The ladder gives up the counts **before** the scope, since `42` means something very different across thirteen repos than inside one. Walked live at 200, 60, 46, 36, 30, 24, 20, 18 and 16 columns:

```
 46   ccmux   Worktrees 16   Pull Requests 2
 36   ccmux   Worktrees 16   PRs 2
 30   ccmux   Worktrees   PRs
 24    Worktrees   PRs
 18    Worktrees
```

That walk found a real defect: at 18 columns the old rule kept a chip whose label had been fitted away to a bare `…`, which on the active fill is a block of colour saying only that something was here. `fitTabs` now requires a real character of the label.

5220 pass / 0 fail; typecheck, knip, build clean. The four new assertions were mutation-checked (break the lead, restore the dot, reorder the ladder, reorder the tail); the tail-order one was added because the first pass proved nothing asserted it.

epilande:
Fresh recording of the new header, replacing the earlier one that still showed the two-line title + tab line.

Screen recording, lightly sped up. **Surface demo (shown, not driven):** the panel's rendering and its mouse and keyboard handling are exercised live in the clip, but no agent lifecycle is driven in it.

What to watch:

- The header is **one line**: the repo in bold, then a chip per view, each carrying its own count. No word is repeated across it.
- Clicking `Pull Requests` moves the fill to that chip and swaps the body to the repo's open PRs. Clicking `Worktrees` moves it back.
- `l` and `h` do the same thing from the keyboard, which is the parity the panel now enforces in one place rather than at each key site.

Recorded against a daemon scoped to this repo alone, so the panel lists only ccmux worktrees.

https://github.com/user-attachments/assets/0c2e90fc-ff13-4e09-9379-b17330069813

Issue #160, opened 2026-08-24
pressKey("escape") types six characters, so four escape assertions test nothing


epilande:
Found while adversarially reviewing #158. Pre-existing and unrelated to that PR.

`createMockKeys().pressKey("escape")` does not send an escape key. `pressKey` resolves its argument as key *input*, and its escape-sequence lookup is keyed on `"\x1B"`, not on the word `"escape"`. So the string falls through as literal text and the handler receives six separate events: `e`, `s`, `c`, `a`, `p`, `e`.

Proven with a throwaway component logging `event.name`:

```
SEEN: ["right","left","e","s","c","a","p","e","x","x+shift"]
```

Four tests assert escape behavior through it and are therefore asserting something else:

- `src/tui/App.test.tsx:775`
- `src/tui/App.test.tsx:4736`
- `src/tui/App.test.tsx:5292`
- `src/tui/App.test.tsx:5344`

They pass, but not for the reason they claim, and they would keep passing if escape handling regressed.

`pressEscape()` is the intended API, but it emits a bare `\x1B` that the parser swallows entirely, producing zero events. So escape is currently untestable in this harness and has to be driven live.

**Scope:** `WorktreesPanel.test.tsx` has zero occurrences, so the panel's own coverage is unaffected.

**What a fix needs:** make `pressEscape()` deliver an event the parser surfaces (likely a timing or lone-ESC disambiguation issue in the key parser), then repoint the four call sites. Worth a lint rule or a type-level guard so `pressKey` cannot silently accept a key *name* where it expects input.

Also worth knowing, same probe: `pressKey("X", {shift: true})` arrives as name `"x"` with `shift: true`, so a bare `case "X"` never fires from a real terminal.

epilande:
Escape is testable after all, and the fix is two parts rather than one.

`createMockKeys` exposes `pressEscape()` (also `pressEnter()`, `pressTab()`, `pressArrow()`), so the six-characters problem is just the wrong call. But swapping to `pressEscape()` alone still asserts nothing: a bare `\u001b` is the prefix of every escape SEQUENCE, so the parser buffers it until either more bytes arrive or its timeout elapses.

Probed against a bare `useKeyboard` component:

- `pressEscape()` then an immediate repaint delivers NO event.
- `pressEscape(); pressKey("q")` delivers ONE event, name `q`, raw `\u001bq` -- the escape was absorbed into meta-q.
[...]

PR #14, opened 2026-07-10
feat(review): Send hunk review notes back to the agent

epilande:
## Overview

Adds review hand-back to the <kbd>d</kbd> hunk integration: after annotating a session's diff in hunk, ccmux offers to send the review notes back to the agent that authored the changes as a single prompt. Follow-up to #11.

## Motivation

The `d` review from #11 is read-only: notes written in hunk die with the hunk session, so feedback still had to be retyped into the agent by hand. This closes the loop. Annotate in hunk, quit, confirm, and the agent starts addressing the comments while the picker keeps watching its state.

Hunk brokers comments only in the live session's memory (nothing is written to disk), so ccmux polls the session while hunk runs, keyed on the picker's own tmux pane id, and takes a final read the moment hunk exits.

### Daemon send path

- **src/daemon/server.ts** - Routes `POST /sessions/:id/send` through the pane-io helpers: multiline text uses tmux bracketed paste (`sendPromptToPane`), single-line keeps `sendLiteralToPane`, and `enter: false` threads through as paste-without-submit
- **src/daemon/pane-io.ts** - `sendPromptToPane` gains a `pressEnter` param for fill mode
- **src/daemon/invokers/claude-invoker.ts** - Passes `pressEnter: true` at existing call sites

This is a general `ccmux send` improvement riding along: multiline text now arrives as one message instead of one submit per line, and the 150ms paste-to-Enter gap fixes the Codex >= 0.124 batching bug where sent text sat in the composer without submitting.

### Review poller

- **src/tui/utils/review.ts** - Polls `hunk session list` / `comment list --type user` sequentially while hunk runs, discovers the session by tmux pane id, retries a final read at exit against hunk's ~2s teardown window, extracts working-tree snippets per note (6 lines / 300 chars caps), and formats the prompt (truncated at the daemon's 10k cap). All hunk-facing calls sit behind one injectable `runHunkJson` seam that degrades to no-offer on any drift. Also adds the empty-diff guard so `d` on a clean tree reports "no changes to review" instead of opening a blank hunk

### Picker hand-off

- **src/tui/App.tsx** - On hunk exit with notes: paneless sessions get a "captured, no pane" toast; otherwise the `reviewHandback` mode decides between the confirm dialog, immediate send, or composer fill. Delivery posts to the daemon's send endpoint and toasts the outcome
- **src/tui/store.ts**, **src/tui/components/ConfirmationDialog.tsx** - `send-review` confirm action ("Send N comments to <agent>?")
[...]

PR #164, opened 2026-08-25
feat(tui): Source picker over open PRs and issues

epilande:
The picker half of #151. **Stacked on #162** (`feat/source-picker`), which has the daemon endpoint and the shared row code this builds on; retarget to `main` after that merges.

`N` in the picker and the sidebar, `n` inside the Worktrees panel: one filterable list of a repo's open pull requests and open issues, whose only verb is Enter. `N` claims the key `App.tsx`'s `case "n"` had explicitly reserved for "some later feature", and the mnemonic is the pair, `n` starts something new from nothing and `N` starts it from something that already exists on GitHub.

## One list, not two tabs

Typing `notif` should reach a PR and an issue at once, because a user remembers the words and not whether the thing they remember was filed as one or the other. Behind a tab boundary a match on the far side reads as "nothing matches", which is this surface's job failing quietly.

Tabs were weighed against the design that saves them (one global filter, filtered counts on both chips so nothing hides) and still lost: the rows are two lines each and the `⊙` / `○` markers plus the section headers already say which kind is which, so a tab buys separation you can see anyway at the price of the cross-kind search.

The section counts are FILTERED, so `Pull requests 0 / Issues 1` says where the matches went. They sit against their own labels with no `·`, since in this TUI that dot divides peers.

## Keys

NAV mode by default, `/` starts the filter. One key means one thing on every surface here, so `j`/`k` move and `q` closes exactly as they do in the panel. The filter row is drawn only while filtering, and Esc out of the filter drops the query along with it, which is what `exitSearchMode` already does for the session picker's search. Those two halves are one decision: a row that hid while the query stayed applied would leave the list narrowed with nothing on screen saying why. A second Esc closes, or reopens the Worktrees panel where that is what opened the picker. In filter mode only the keys this handler acts on are `preventDefault`ed, because the filter is an `<input>` and the panel's unconditional default at the top of its own handler would kill typing dead.

## Enter routes three ways

- A PR no checkout holds opens the dialog in **PR mode**.
- An issue opens it in the mirroring **issue mode** (new here): `NewSessionDraft.issue`, the daemon deriving the name with `slugForIssue` and cutting from the repo's default branch, since there is no head to fetch. Keeps Agent/Placement/Prompt, drops Where and Name, exactly as PR mode does and for the same reason.
- A source **already checked out** goes to that checkout, through the same live-session revalidation `spawnInWorktree` does, because a worktree holds one agent.

A cancelled dialog returns to the picker with its cursor *and* its filter restored, and in the filter, since under the Esc rule a non-empty query can only have been typed there.

## The checked-out proof differs by kind, deliberately

A PR is matched by SHA and never by branch name, because a fork chooses its own names. An issue is matched by NAME, and that is safe precisely because the name is ccmux's own: `slugForIssue` derives `issue-<n>` or `issue-<n>-<slug>`, so a match proves this tool made that directory for that issue. The prefix is family-exact, since a bare `startsWith` makes issue #14 claim `issue-144-foo`; where several exist (a second spawn derives `-2` rather than opening the first) the shortest name wins and the rest are counted in the phrase, because there is no SHA here to break the tie and silently choosing one of two live checkouts must be visible.

This was specced as a later PR and pulled forward: in one list, Enter on a PR routing to its checkout while Enter on an issue silently derived `issue-144-foo-2` would be the same key with opposite behaviour.

## Layout

Section and repo headers are lines the cursor never stops on, but lines the layout COUNTS: a scroll target computed without them drifts further out of true the further down the list you go, which is the bug the panel already paid for over its removable divider. When no repo has a row, the list is replaced by a real empty state rather than a scrollbox of headers, which is also what keeps every row reachable from the keyboard. The empty state blames the filter first, before any report about GitHub, because the query is what the user would change.

## Verification

- 5352 pass / 0 fail; typecheck, knip and build clean.
- **Live, end to end** in an isolated daemon: `N` opens it, `/` reveals the filter row and narrows as the counts follow, Esc clears the query and takes the row with it, Enter on an issue opens the dialog in issue mode with the right note and no worktree rows, Cancel returns to the picker with filter and cursor intact, and panel `n` → picker → Esc lands back on the panel.
- Eleven mutations across the two new test files, each confirmed to kill a test.

One harness finding worth recording: a bare ESC is the prefix of every escape *sequence*, so the parser holds it until its timeout elapses. `pressEscape()` followed by an immediate repaint reports nothing at all, and the next keypress arrives as meta-that-key instead. The tests wait it out. (`pressKey("escape")` is a different trap, and is #160.)


epilande:
Note on the empty check list: `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]` only, so a PR based on another branch runs nothing. Not a failure, and not skipped tests -- the same suite ran locally at 5352 pass / 0 fail with typecheck, knip and build clean. CI will run once this is retargeted to `main` after #162 merges.

epilande:
Surface demo (screen recording, sped up) of the new SourcePicker (N) added by this PR. Shows: opening the picker, filtering PRs/issues live (Pull requests/Issues counts narrow), navigating rows, pressing Enter on an issue to open the New Session dialog seeded from it, then Escape back out with the filter and cursor preserved before closing.

This is a navigation walkthrough, not a driven idle->working->waiting agent run - it demos the surface, it doesn't verify agent state transitions.

https://github.com/user-attachments/assets/2a064c89-f68a-478b-8fa2-79207ff4bd6f

PR #169, opened 2026-08-29
feat(cli): Shell completions for zsh, bash, and fish

### LINKED ISSUE #153: feat(cli): Shell completions
## Problem

No completions for commands, flags, or session refs; every ref gets typed from memory.

## Proposal

Completions for zsh/bash (fish if cheap): commands, flags, and where feasible dynamic values (session refs, agent types). Commander has no built-in generator, so step one is picking the approach (ecosystem package vs. generated script).

## Delivery

Single PR, including install instructions in the README.

epilande:
## Overview

Adds `ccmux completion <shell>` for zsh, bash, and fish: commands, flags, and live values (session refs, agent names, invocation ids, config keys and values) completed from one source of truth.

Resolves #153

## Motivation

Every session ref currently gets typed from memory, and there is no completion for commands or flags at all. Commander has no completion generator, so the approach question came first: a static generated script drifts from the CLI the moment a flag changes, and it cannot know about live sessions. Instead the wrapper scripts are tiny and every Tab press asks the binary.

### Completion engine

- **src/commands/completion.ts** - `ccmux completion <shell>` prints the wrapper script. A hidden `ccmux __complete <shell> -- <words...>` walks the live Commander tree (`createHelp().visibleCommands/visibleOptions`, argument `.choices()`) and a value-source registry for dynamic values: session ids and `%pane` ids on every session-taking argument and on `--resume`/`--fork`/`--target`, `self` where a full ref is accepted, native ids for `--session`, agent names, invocation ids for `invoke cancel|result`, config keys and their enumerable values. Daemon reads use a 300ms timeout and never start a daemon. `--cwd`/`--repo` emit a `:dir` directive so the shell completes paths in the user's directory (`bin/ccmux` cd's into the package before running).
- **src/commands/completion.test.ts** - Walk, option-value, `--`, directive, and rendering cases against the real command tree, plus a guard that finds every `session-id`/`session-ref`/`from`/`to` argument in the tree and asserts each completes sessions, so a new command cannot ship without it.
- **src/commands/config.ts** - `KNOWN_KEYS` entries gain `choices` where the value set is enumerable; `completableConfigKeys()` and `configValueChoices()` expose keys and values to the completer.

### CLI wiring

- **src/program.ts** - `createProgram()` builds the command tree (moved from `src/index.ts`) so tests can walk the real tree. Registers `completion` and the hidden `__complete`.
- **src/index.ts** - Now a shim that parses `createProgram()`.

### Install paths

- **README.md** - `Shell Completions` section under Installation with the zsh/bash/fish snippets, plus a CLI table row.
- **scripts/update-homebrew-formula.sh** - `generate_completions_from_executable(bin/"ccmux", "completion")`, so the formula installs completions for all three shells.
- **docs/architecture.md** - Code-map row.

## Notes

- The bash script copies `COMP_WORDS` with an index loop instead of `"${COMP_WORDS[@]:1:COMP_CWORD}"`: bash 3.2 (macOS default) joins the quoted slice into a single word, which made every multi-word completion return nothing.
- Not in this PR: `--opt=value` splitting, gh-backed `--pr`/`--issue` numbers, `--base` branch names (would run in the wrong repo), and the open-ended `columns.*`/`breakpoints.*` config keys.
- Per-Tab cost is ~150ms (one binary launch).

## Breaking Changes

None

## Testing

- [x] `bun run typecheck`, `bun run knip`, full `bun test` (5438 pass)
- [x] zsh live in a detached tmux session: root listing with descriptions, prefix, `switch <TAB>` sessions, `--untracked` values, `--cwd` directories, `config set theme <TAB>`
- [x] bash 3.2 live: same set
- [x] Compiled binary (`bun build --compile`) `completion zsh|fish` and `__complete` under a throwaway `HOME`
- [ ] fish: code-reviewed only (not installed locally)


epilande:
Demo of the zsh completions running against a live daemon: the command list with descriptions, live session refs and %pane ids on `switch`, option values, and the directory fallback on `--cwd`.

https://github.com/user-attachments/assets/70a82434-7efc-4040-a033-e8e4a367562e

PR #170, opened 2026-08-30
fix(tui): Long input values scroll instead of widening the dialog

epilande:
## Summary

Typing a value longer than the control in the Hand off dialog's **Note** field widened the input straight through the dialog's right border instead of scrolling. The New session dialog's **Prompt** and worktree **Name** fields had the same defect (a comment in `HandoffDialog.tsx` even called it out as long-standing).

Cause: an OpenTUI `<input>` measures its own width from its text, and Yoga's default `flexShrink` is `0`, so once the value outgrows the available columns the measured width becomes the flex basis, the input can't shrink, and it (plus the `flexGrow` wrapper box painting the control background) overflows the box.

Fix: `flexShrink={1}` on the input and its wrapper in all three fields, so the input gets the width the row actually has and scrolls the value inside it. The stale comment in `HandoffDialog.tsx` is rewritten to explain the mechanism.

## Testing

- [x] `bun test` passes (5466 pass, 0 fail)
- [ ] Verified end to end with a real agent session (if this touches detection, state, the daemon, or the CLI) — N/A, rendering only

Regression tests added on both dialogs: render with a long value at 60 columns and assert the right border column survives on that row, the value's tail is visible, and its head is scrolled off. Both fail on the previous code and pass with the fix. The worktree Name field has no dedicated test; it is the identical mechanism, covered by the Prompt test.

Verified live in the picker against two real Claude sessions (row menu → Hand off → pick target → Tab → long note; and `n` → Tab to Prompt → long prompt).

## Notes for reviewers

Before/after screen recordings are in https://github.com/epilande/ccmux/pull/170#issuecomment-5467596164 (before: the Note control widens through the border; after: the same flow, plus the New session Prompt field).

### DIFF
--- a/src/tui/components/HandoffDialog.test.tsx
+++ b/src/tui/components/HandoffDialog.test.tsx
@@ -170,6 +170,27 @@ describe("HandoffDialog", () => {
     expect(frame).toContain("CancelSend");
   });
 
+  it("scrolls a long note inside the control instead of widening it", async () => {
+    // An OpenTUI input measures its own width from its text, and Yoga's
+    // default `flexShrink` is 0, so a value longer than the control used to
+    // widen it (and the wrapper box behind it) straight through the right
+    // border. The border column must survive on the Note row, with the
+    // value's tail shown and its head scrolled off.
+    const note =
+      "default very long text what happening here and beyond the border";
+    const lines = (await render({ note, field: "note", width: 60 })).split(
+      "\n",
+    );
+    const noteLine = lines.find((line) => line.includes("Note"));
+    expect(noteLine).toBeDefined();
+    const borderLine = lines.find((line) => line.includes("Hand off"));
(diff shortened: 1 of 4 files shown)


epilande:
Screen recordings of the repro and the fix (sped up in places). Both takes end with Escape; nothing was sent.

**Before** — row menu → Hand off → pick target → Tab to Note → type a long note. The control widens through the dialog's right border:

https://github.com/user-attachments/assets/eeda4478-dac7-46b9-9d3a-31496bf75405

**After** — same flow. The value scrolls inside the control and the border stays put:

https://github.com/user-attachments/assets/2cbe4110-898a-4205-82d0-9dea5ed9d99c

**After, New session dialog** — `n` → Tab to Prompt → long prompt, same fix:

https://github.com/user-attachments/assets/2ebb6d8a-d2f0-4d27-8e16-80f9f701ca5f

PR #174, opened 2026-08-30
feat(spawn): Name windows after the worktree and add a --model passthrough

### LINKED ISSUE #171: feat(spawn): Name the window after the worktree and add a --model passthrough
## Problem / motivation

`ccmux spawn` is the natural way to dispatch one agent per task into its own tmux window and worktree (`ccmux spawn claude --issue 163 --detach`), but two gaps make a batch of spawns hard to use:

1. **Windows are indistinguishable.** The new window is left on `automatic-rename`, so it takes the agent's pane title. For Claude Code that is the version string, and two spawns show up as `2.1.251 | 6` and `2.1.251 | 7`. Telling them apart means running `tmux rename-window` by hand after each spawn.
2. **No way to pick the model.** `spawn` has no `--model`, and the agent starts with whatever the user's global settings say (for Claude, `~/.claude/settings.json` `model`). Spawning a session on a different model currently means pre-creating the worktree and dropping a `.claude/settings.local.json` into it before spawning, or overriding `promptCommand` in `ccmux.json`.

## Proposed solution

- Name the window at spawn time: the worktree name when there is one (`issue-163`, `pr-155`, the `--worktree` name), otherwise the agent name. Setting the name turns `automatic-rename` off for that window, which is the point. `--split` panes are unaffected.
- Add `--model <name>`, mapped onto each agent's own flag where one exists (`claude --model`, `codex -m`, etc., per `AgentDef`), and refused with a clear message for agents that have none. This is the same shape as `promptCommand`: per-agent and declared in `src/lib/agents.ts`, so a custom agent can declare its own.
- Update the `dispatch` skill's spawn section in the same PR (`--issue`/`--pr`/`--worktree`/`--detach`, window naming, `--model`) and bump `plugins/ccmux/.claude-plugin/plugin.json` so the installed plugin cache refreshes; the cache is still serving the pre-#128 text at `1.0.0`.

## Alternatives considered

- A generic `--agent-args` passthrough instead of `--model`. Simpler to implement but leaks each agent's CLI into the ccmux surface; `--model` is the one option people actually reach for.
- Leaving window naming to the user. It is one tmux command, but it has to happen after every spawn and the picker's own `n` dialog has no place to run it.

## Relevant agent(s)

- Not agent-specific (model flag mapping is per agent)

epilande:
## Overview

Names the tmux window `ccmux spawn` creates after the worktree (or the agent) and adds `--model <name>`, passed through as each agent's own model flag.

Resolves #171

## Motivation

A batch of `ccmux spawn claude --issue <n> --detach` calls produced windows that all read as Claude's version string, so telling them apart meant a manual `tmux rename-window` after every spawn. And there was no way to pick the model at spawn time short of pre-creating the worktree with a `settings.local.json` in it or overriding `promptCommand` in `ccmux.json`.

### Window naming

- **src/daemon/spawn-command.ts** - `TmuxSpawnArgvInput.windowName`, passed as `-n` on `new-window` only. Setting it pins `automatic-rename` off for that window, which is the point. A split pane lives in someone else's window and never gets one
- **src/daemon/server.ts** - Name is `worktreeInfo?.name ?? agent.name`: the `--worktree` name, `issue-<n>-...`, `pr-<n>-...`, a reused checkout, else the resolved agent

### `--model` passthrough

- **src/lib/agents.ts** - `AgentDef.modelFlag`, the same per-agent shape as `promptCommand`. Declared as `--model` on all nine built-ins, each read from the CLI's own `--help` (Claude, Cursor, Pi, omp, Copilot, and Antigravity's `agy` take `--model`; Codex, OpenCode, and Gemini take `-m, --model`; `codex resume` carries it too). Carried through both config merge paths
- **src/lib/preferences.ts** - `AgentConfig.modelFlag`, so a custom agent declares its own
- **src/daemon/spawn-command.ts** - `MODEL_PATTERN` and `normalizeModel`. The value is spliced unquoted after the launcher binary, so it must be shell-inert: letters, digits, `.`, `_`, `:`, `/`, `-`, never a leading `-` (wide enough for `provider/model` and `sonnet:high`). Spliced into `{bin}` for bare, prompt, and fork spawns; appended on a resume, whose templates carry no `{bin}`. An agent without a `modelFlag` is refused with a message naming the config key
- **src/daemon/server.ts** - `POST /spawn` accepts `model`, validates it (400), and hands it to both command builders
- **src/commands/spawn.ts** - `--model <name>`, validated in the option parser so a typo is rejected before `ensureDaemon`
- **src/daemon/spawn-command.ts** (follow-up, 54367fb) - `buildAgentForkCommand`'s id-only branch substituted `bin: binary` where the prompt and `{path}` branches both use `launcher`, so `--model` was silently dropped (no error, the agent came up on its default) for the `{bin} --resume {id} --fork-session` shape. No built-in uses that form, but it is the fallback the `{path}`-resolution refusal recommends, and the `forkAgent` test fixture uses it

### Tests

- **src/daemon/spawn-command.test.ts** - `-n` placement and absence on split; `normalizeModel` accept/reject cases; model on the bare, prompt, resume, and fork paths; refusal without a flag and on a non-string flag
- **src/lib/agents.test.ts** - Every built-in declares `--model`; a custom agent's `modelFlag` survives the merge
- **src/commands/spawn.test.ts**, **src/daemon/server.test.ts** - CLI value validation; the exact `new-window` argv now carries the name
- **src/daemon/server.test.ts**, **src/daemon/spawn-command.test.ts** (follow-up, 54367fb) - the `--model` wire path through `POST /spawn` had no coverage, so dropping `model` from either builder call left the suite green. Three server tests (spawn, a 400 refusal with nothing created, fork) plus one builder test for the id form; both fork tests fail without the fix above

### Docs and plugin

- **README.md**, **docs/agent-adapters.md** - `--model` example, the per-agent flag findings, the value pattern, and the window naming rule
- **plugins/ccmux/skills/dispatch/SKILL.md** - Spawn section covers `--issue`/`--pr`/`--worktree`/`--detach`, window naming, and `--model`
- **plugins/ccmux/.claude-plugin/plugin.json** - 1.0.0 to 1.1.0 so the installed cache refreshes

Left out on purpose: the picker's new-session dialog has no model field (CLI-only for now), and shell completion offers `--model` as a flag but has no value source, since the values are per agent and dynamic.

## Breaking Changes

None. A window created by `ccmux spawn` now has a fixed name instead of following the pane title; `tmux rename-window` still works on it.

## Testing

- [x] `bun run typecheck`, `bun run knip`, `bun test` (5533 pass, 0 fail)
- [x] Isolated daemon (own tmux socket, `CCMUX_PORT`, `CCMUX_HOME`) on the worktree build: `spawn claude --worktree demo-171 --model opus --detach` gives a window named `demo-171` with `automatic-rename` off and a pane running `claude --model opus` (Claude's banner reported Opus)
- [x] `spawn codex --model gpt-5 --detach` gives a window named `codex` running `codex --model gpt-5`
- [x] `spawn pi --split --model sonnet` leaves the host window's name and `automatic-rename` untouched; the pane runs `pi --model sonnet`
- [x] `spawn claude --model '-x'` is refused at the CLI with nothing started

`--issue` and `--pr` were not spawned live (they need a GitHub-backed repo to cut a worktree in); their window name is the worktree name, covered by the unit tests on `worktreeInfo.name` and the existing slug tests.


epilande:
Demo: spawn names the window after the worktree and passes --model through.

Recorded on the branch tip 2039926 against an isolated daemon running that build. The clip is sped up between steps, nothing else is edited.


https://github.com/user-attachments/assets/12517df4-9487-45dc-a106-291ff96e0fe3

PR #176, opened 2026-08-31
feat(prune): Offer merged-branch worktrees with an idle agent, ending the session

### LINKED ISSUE #175: feat(prune): Offer merged-branch worktrees with an idle agent, ending the session
## Problem / motivation

The end of a dispatched-agent cycle has leftover friction. After a spawned agent's PR merges, the worktree, local branch, tmux window, and the idle agent sitting in it all remain. `ccmux worktree prune` refuses to touch any of it: classification skips a worktree with ANY bound session, idle included ("an agent is idle here"), so cleanup today means manually finding each window, closing the agent, and only then pruning.

The skip is right in general: a bound session is the one signal that survives the fresh-vs-merged ambiguity (a worktree branched from base reads as "already merged" to every local check). But when the branch's PR is verifiably merged, that ambiguity is gone, and an idle agent on a merged branch is provably leftover.

## Proposed solution

Let a worktree with an idle session become a prune candidate when all three hold:

- the session is `idle` (`working` and `waiting` stay absolute skips; `waiting` is mid-turn by definition)
- the branch's PR state is MERGED from the gh lookup, not merely `isMergedInto` locally
- the working tree is clean (the existing dirty opt-in continues to apply on its own axis)

Offered, not silent: the candidate row is labeled so selecting it is consent ("closes the idle agent session"), the same shape as the existing dirty opt-in. On the CLI, an explicit flag (e.g. `--end-idle`) includes these candidates. Removal reuses the existing machinery: SIGTERM then SIGKILL on the agent, `kill-pane` (tmux closes the window with its last pane), then directory, branch, and state cleanup.

Cost of a wrong kill is low by design: the agent's transcript persists, so the session stays resumable after the pane dies.

Also in this PR: a short end-of-cycle cleanup note in the dispatch skill ("merge, then prune"), plus the plugin version bump so the installed cache refreshes.

## Alternatives considered

- Fully automatic (no selection, no flag): least friction, but prune's design is refusals over guesses, and idle-vs-waiting classification has misfired historically; one visible consent step is the right ceremony.
- Status quo plus documentation: keeps the two-step "kill the window, then prune" recipe forever; the friction is the complaint.

## Relevant agent(s)

- Not agent-specific (session status and PR state gate it, per-agent only in how the process is terminated)

epilande:
## Overview

Offers a worktree whose only occupant is an idle agent as a prune candidate once GitHub reports its branch's PR merged, and ends that agent (SIGTERM, then SIGKILL, then the pane) as part of the removal. The offer is explicit on both surfaces: the Worktrees panel labels the row and selecting it is the consent, and `ccmux worktree prune` needs `--end-idle`.

Resolves #175

## Motivation

The end of a dispatched-agent cycle left friction: after a spawned agent's PR merged, the worktree, local branch, tmux window, and the idle agent sitting in it all stayed behind, and `ccmux worktree prune` refused to touch any of it because classification skipped a worktree with any bound session. Cleanup meant finding each window, closing the agent by hand, and only then pruning.

The wide skip exists because local evidence cannot tell a fresh worktree from a finished one (a worktree just branched from base reads as "already merged" to every local check). A gh-verified MERGED state is what dissolves that ambiguity, so an idle agent on such a branch is provably leftover.

### Daemon

- **worktree-prune.ts** - A worktree whose bound sessions are all `idle` (and none `background`) goes through a new `classifyEndIdle` arm: only the SHA-proven `pr-merged` may offer it; `merged-locally`, `upstream-gone`, `pr-closed`, no PR, and a failed lookup all keep the `an agent is idle here` skip. `working`, `waiting`, and background sessions (subagent entries included) hold the worktree exactly as before and still spend no `gh` call; the idle arm spends one only on an open-PR cache miss. `runPrune` gains `allowEndIdlePaths`, an independent axis next to `allowDirtyPaths` (a dirty worktree an idle agent occupies needs both, and each is refused separately), plus a `liveSession` seam it re-reads right before signalling, so consent never overrides an agent that went back to work AND the pid it signals is the one that re-read returned rather than the one the scan recorded. Dry runs report the sessions they would end.
- **server.ts** - `POST /worktrees/prune` accepts `allowEndIdle` with the same over-cap check and normalization as `allowDirty`, and backs `liveSession` with the live session manager. The endpoint's per-request re-scan is unchanged, so a caller-supplied path still has to re-classify as removable.
- **worktree-prune.test.ts**, **server.test.ts** - Cover the gates: idle+merged+clean offered (one lookup), working/waiting refused (zero lookups), merged-but-dirty double-gated, no gh proof skipped, background and subagent sessions hold, consent without the opt-in refused before anything is signalled, `sessionStatus` reporting `working` refused despite consent, and the endpoint's refused outcome and over-cap 400.

### CLI

- **worktree.ts** - Adds `--end-idle` to `worktree prune`. Without it, an offered candidate an idle agent occupies is withheld under "Not offered" with `an agent is idle here; pass --end-idle to include it (removal ends the session)` and never enters the count, the dry run, or the selection. With it, the row says `removal ends this session`, the confirmation adds `It will also end N idle agent session(s). Transcripts persist on disk, but agents that resume by directory (opencode, pi, omp) lose the way back once the worktree is gone.`, and the request carries `allowEndIdle`. The command's daemon-facing edges are injectable (`PruneCommandDeps`) so the flag's effect on the request body is tested without a daemon.
- **worktree.test.ts** - Flag off/on through `planPrune`, the dry-run body, the interactive confirm line, and the posted consent list.

### TUI

- **WorktreesPanel.tsx** - A removable row with idle sessions reads `claude idle (x ends it)` and `claude idle (will be ended)` once selected, yellow rather than red because the transcript persists. In compact (sidebar) mode the consequence leads the detail line and the reason takes the truncating tail, mirroring the dirty rule. <kbd>a</kbd> skips such rows (a bulk key is never the consent) and says so; the confirm adds `ending N idle agent session(s)`; the request sends `allowEndIdle` for the selected ones. One predicate, `endsSessions`, gates all four so they cannot disagree, and an older daemon (empty `sessions`) leaves every one of them as it was.
- **WorktreesPanel.test.tsx** - Consent wording in both states and widths, kept-row phrase unchanged, `a` gating plus flash, confirm line, single-target <kbd>x</kbd>, and the posted body.

### Docs and plugin

- **README.md** - `--end-idle` in the command table and examples; the prune prose and safety rules now describe the working/waiting skip and the idle-on-merged opt-in.
- **docs/architecture.md** - The session-gate paragraph in "Worktree pruning" describes the new rule, the two consent axes, and the live-status re-check; run phase 1 notes the re-check precedes the SIGTERM.
- **plugins/ccmux/skills/dispatch/SKILL.md** - "Cleaning up after a merge": `ccmux worktree prune --end-idle` (or the panel) takes the worktree, branch, window, and idle agent in one step.
- **plugins/ccmux/.claude-plugin/plugin.json** - 1.1.0 to 1.2.0 so the installed cache refreshes.

## Breaking Changes

None. A worktree with an idle agent that was a skip before is now an offered candidate, but nothing removes it without the new consent: an older TUI or a hand-written POST that omits `allowEndIdle` gets a refused outcome, not a removal.

## Testing

- [x] `bun run typecheck` clean, `bun test` 5580 pass / 0 fail
- [x] Live, on an isolated tmux socket and daemon port from this branch's build, against a throwaway clone whose `issue-171` worktree tip equals PR #174's `headRefOid` (a real merged PR, so the gh proof is genuine) with a pane-tracked stand-in agent idle inside it:
  - `ccmux worktree prune --dry-run` withholds the row with the flag-naming reason; `--end-idle --dry-run` offers it and reports `would stop agent`
  - `ccmux worktree prune --end-idle` interactively: `stop agent: gemini pid exited`, `close pane`, directory moved aside and deleted, `git branch -D issue-171`
  - Picker Worktrees panel: row reads `PR #174 merged · gemini idle (x ends it)`, `(will be ended)` once selected, confirm shows `ending 1 idle agent session`, removal takes pane, process, directory, and branch
  - `POST /worktrees/prune` without `allowEndIdle` answers 200 with a refused outcome (`has an idle agent session and was not opted in`), nothing touched
  - Flipping the stand-in to `waiting` (gemini's terminal rules) withholds the worktree even with `--end-idle`
### Review follow-up (cb11869)

A three-lane review of the branch turned up no Must findings and three Shoulds, all fixed here:

- The re-check verified the LIVE session row and then signalled the SCAN's pid. The reconciler rewrites a row's pid when a new process takes over its pane and clears it on a soft evict, so the status that authorized the kill could describe one process while the SIGTERM landed on another — the mismatch `handleKillSession` keeps its own rule against. The seam (`sessionStatus` → `liveSession`) now returns the pid with the status, resolved once and used for both the refusal and the signal; an `undefined` answer proceeds with a null pid instead of the scan's, so it closes the pane and signals nothing.
- The endpoint's own supplier had no test and fails OPEN. Covered through a dry run, which reports the pid the real run would signal and touches nothing.
- `WorktreesPanel.test.tsx`'s row-height test passed `compact` where it meant `selected`, so it asserted nothing about the invariant its comment named. It now pins what actually keeps the height stable: selecting a row changes the phrases' text and never their count, at both widths.

- [x] Test-verified only (no live path reaches them): the removal-time `sessionStatus` re-check refusing an agent that went `working` after the scan, the background/subagent hold, the merged-but-dirty double gate, and the `allowEndIdle` over-cap 400

The live stand-in was a gemini-named process (pane-tracked) rather than a Claude one, because a machine with Claude hooks installed tracks Claude natively and never binds a fake `claude` pane; the end-idle logic is agent-independent, and the kill path is the existing `stopSessions`.


epilande:
Screen recording (sped up) of the Worktrees panel on this branch, against a real merged-PR worktree with two idle agents in it.

**Surface demo (shown, not driven):** it walks the consent surface and stops at the confirmation, then escapes out. Nothing is removed, so the removal path itself is not exercised on camera.

What to look at:

- `issue-160` is offered as `PR #172 merged · 2 agents idle (x ends them)`, while `issue-175` (`PR #176 open · claude idle`) and `issue-163` (`branch gone · 3 agents idle`) sit unoffered right above it. Only a gh-verified MERGED PR opens the gate; an open PR and local-only evidence both keep the skip.
- Selecting the row flips the note to `(will be ended)` and colours it yellow rather than red: selection is the consent, and the transcript file survives the removal.
- The confirmation restates the consequence: `Delete 1 worktree and its branch?` / `ending 2 idle agent sessions`.
- `a` skips the row and the footer drops to `x remove 9`: a bulk key is never the consent.

https://github.com/user-attachments/assets/738a5a78-e71d-43c1-877b-9b8b2ab41c03

PR #178, opened 2026-09-01
fix(tui): Keep list cursor on the previous row after kill

epilande:
## Summary

Keeps the picker cursor on the previous living row after `x` kills the selected session, instead of jumping back to row 1.

Killing a mid-list or bottom session cleared `selectedSessionId`, so `selectedIndex()` fell back to 0. After a kill, selection now lands on the last pre-kill row that still exists in the **rebuilt** flat list. Killing the top row stays on the new 0. Empty list is fine. Does not wrap.

`killedIndex-1` is not used: that indexes the post-rebuild list with the pre-kill index. Killing the last member of a middle group also drops that group's header (`buildFlatItems` omits empty groups), so `i-1` jumped **forward** onto the next project's header and the next `x` became kill-group on the wrong project.

## Related issue

—

## Testing

- [x] `bun test src/tui/store.test.ts` — 255 pass, including ungrouped mash-`x` plus grouped last-of-middle-group and first-of-many / header
- [x] Isolated picker (prior commit): four Gemini stand-ins on `-L ccmux-kill-verify`, `G` to last row (`agents:3.0`), `x` `y` lands preview on `agents:2.0`; mashing `x` walked 3.0 → 2.0 → 1.0 → 0.0 → empty. Did not wrap.

## Notes for reviewers

`removeSession` is the SSE `session_removed` path the confirm-`x` flow already uses. Selection stays by session id (or header key). After the list filter, the last still-living predecessor from the pre-kill prefix is selected.

The dead id is cleared first inside the same `batch` as the list filter. Clearing first is load-bearing: `selectedIndex()` falls back to 0 when the id is missing, which is the jump this fixes.

Grouped default (`groupBy: project`): last member of a middle group lands on the previous group's last session, not the next group's header. First-of-many lands on the living group header (literal previous row).

<div><a href="https://cursor.com/agents/bc-0869a07a-69ab-4ffd-8acc-e33baa21d565?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0869a07a-69ab-4ffd-8acc-e33baa21d565&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


epilande:
Screen recording at d81f628, real time, no speedup.

Scenario A: killing the last of three sessions in `beta-api` leaves the cursor on the row above, not at the top of the list. Scenario B: killing `alpha-app`'s only session, whose group header disappears with it, lands the cursor on `beta-api`'s first session row rather than on that group's header, where the next `x` would have been kill-group on a project nobody moved to. Scenario C is also captured: killing the middle group `delta-lib`'s only session falls back to the previous group's last session.

Rows are stand-in agent processes in throwaway project directories, run against an isolated tmux server and daemon.

https://github.com/user-attachments/assets/49ff883b-78dc-4cd1-9da9-462fa985d1d5

PR #179, opened 2026-09-02
fix(opencode): Keep one server's markers off a sibling window's row

### LINKED ISSUE #177: [Bug]: OpenCode last prompt leaks across windows in one dir
### Preflight

### ccmux version

1.3.2

### Agent

OpenCode

### Are hooks installed?

Yes

### What happened?

I run two OpenCode instances in the same tmux session, in the same working directory. Each instance is in its own tmux window (the status-bar “tab”, not a pane split).

ccmux often shows the same last prompt on both rows, even when the OpenCode conversations are different.

I expected each row to show the last prompt from that window’s OpenCode session.

Instead, both `working` rows under one project share a subtitle. Example: window A was prompted `fix the flaky test` and window B `refactor the parser`, but both ccmux rows showed `fix the flaky test`. A third idle pane in that directory had no prompt. This happens often, not on every turn.

### Steps to reproduce

1. In one tmux session, `cd` to the same git repo in two windows.
2. Start `opencode` in each window.
3. Send different prompts in each OpenCode conversation.
4. Open `ccmux picker` (or the sidebar).

### OS and tmux version

macOS 26.5.2, tmux 3.7c

Likely ccmux-side: the OpenCode plugin writes `opencode-<session_id>.json` (no pid in the name). `eagerSeed` lists every session for that directory. The reconciler applies a marker by `nativeSessionId` without checking that `marker.pid` still belongs to that pane. Two processes then clobber one marker file, and the newest `lastPrompt` lands on both rows.

OpenCode also shares `~/.local/share/opencode/opencode.db` per project ([anomalyco/opencode#31307](https://github.com/anomalyco/opencode/issues/31307)). That can make session ids collide. It does not make painting one prompt onto two ccmux rows correct.

When it happens, compare per row from `ccmux show --json`: `id`, `tmuxPane`, `pid`, `nativeSessionId`, `lastPrompt`.

epilande:
## What happens

Two `opencode` processes in one directory share OpenCode's SQLite db (anomalyco/opencode#31307), so each sees the other's sessions. Reproduced live with real `opencode` instances; the bug is broader than the title:

- The plugin's boot seed wrote a marker for **every** session `session.list` returned, stamped with its own pid. Window B's boot therefore rewrote window A's live `opencode-<sid>.json`: pid flipped to B, `last_prompt` dropped, state reset to idle.
- The daemon's tick looked the marker up by the row's `nativeSessionId` and folded every marker sharing that marker's pid, with no check that the pid lives in the row's pane. Row A painted B's prompt **and B's `working` status**.
- It never self-heals: pane A no longer has a marker claiming its pid, so the link pass has nothing to relink from. "Often, not every turn" is really "every sibling boot in the same cwd, until the older window takes a turn." A third boot put the third window's prompt and status on all three rows.

## Fix

Two halves, each sufficient on its own to stop the paint.

**Plugin** (`src/plugins/opencode/plugin.js`): the seed writes markers only for sessions this server's `client.session.status` lists. OpenCode's `SessionStatus.list()` is a per-process in-memory map with idle entries deleted, so membership is the only evidence the server hosts a session. `session.list` is project-wide over the shared db and never was. Side effect worth having: the old seed also wrote every historical session in the same millisecond, so a fresh idle row's `nativeSessionId` was an arbitrary old session and `ccmux last` read a random transcript.

**Daemon** (`src/daemon/state-reconciler.ts`, wired in `index.ts`): a new `paneIdHostingPid` dep resolves a marker's pid to its pane by process ancestry (sync `findPaneHostingPid` over the pane cache and last process tree). `ownsOpenCodeMarker` drops a marker hosted by a different pane and warns once per (row, marker, pid). Unknown hosting (boot window, or a pid newer than the last snapshot) fails open.

Docs: `docs/agent-adapters.md` OpenCode section and `docs/architecture.md#opencode-aggregation` state both rules.

Closes #177.

## Verification

- `bun test` 5541/0, typecheck clean; 8 new tests (seed gate, sibling marker byte-identical, foreign-pane drop + warn-once, own-pane apply, both fail-open branches, non-opencode untouched).
- Live, isolated harness (own tmux server, daemon, `CCMUX_HOME`, `OPENCODE_DB`), three real `opencode` instances in one cwd:
  - **new plugin + new daemon**: marker A byte-identical through both sibling boots; every row only ever showed its own prompt and status.
  - **old plugin + new daemon**: the clobber still lands on disk, but no row painted a foreign prompt or status; the warn line fired as designed.

Before/after recording: https://github.com/epilande/ccmux/pull/179#issuecomment-5507374232 (the "after" take runs e51f5d7, the fix plus a comments-only follow-up).

## Caveats

- The plugin half reaches a machine only after `ccmux setup --agent opencode`. The daemon half ships with the upgrade and stops the paint on its own, with one measured sub-second window (689ms): a sibling's fresh pid is not in the last process snapshot yet, so the guard fails open for one fold. Only a stale plugin can trigger it; failing closed instead would blind every opencode restart in an existing pane for a scan interval.
- Two old-plugin residuals the daemon guard cannot reach, both gone with the new plugin: on a clobber the session id migrates to the newest pane for ~600ms (`ccmux last`/handoff on that row would read the wrong transcript in the window), and a clobbered row's stored prompt can go `null`.
- Two processes genuinely hosting one session id (the 31307 shape) is unresolvable with one marker per session id; that is OpenCode's bug and is documented as such.
- A boot where `session.status` itself fails now seeds nothing. Intended: a failed status call is not evidence of hosting, and the first bus event writes the marker anyway.


epilande:
Before/after recording (56s), two real `opencode` instances (GPT-5.4 mini) in one directory, picker on the left. Panes rather than windows so both TUIs are on camera; the mechanism is cwd-based. Before (6289fcb, ~0:21): the row for pane `demo:1.2` reads `work · BRAVO…` while that pane sits idle at its finished ALPHA reply, and after both settle every row reads BRAVO. After (e51f5d7, the fix from 72d1da1 plus a comments-only follow-up, ~0:46): `demo:1.3` reads `work · BRAVO…`, `demo:1.2` reads `done · ALPHA…`. Model latency and boots are sped up; the proof frames are held at 1x. The provenance still at the start of each take is a real frame from that take, prepended because the picker paints over it within half a second.

https://github.com/user-attachments/assets/be12a194-cd5a-4171-a188-d6b78e9cb93f

deferred:
Awesome, thanks for your work! May I ask you to prep a release to try it out? :)

PR #180, opened 2026-09-03
fix: target picker switches to invoking tmux client

TwegZhang:
## Summary

- Capture the tmux client tty that opens the picker popup and pass it into the picker.
- Pin pane selection to that client with `switch-client -c`, while retaining the current-client fallback for direct and legacy launches.
- Refuse the switch with a visible toast when no valid client can be identified, instead of risking another attached client.

## Related issue

Refs #75

## Testing

- [x] `bun test` passes (5602 passed, 2 skipped, 0 failed)
- [x] Verified the picker in a fully isolated tmux server with two attached clients
- [ ] Verified end to end with a real agent session (not required for session detection; the switch path was exercised with a process-matched stand-in)

## Notes for reviewers

Reproduced the regression with clients on `alpha` and `beta`: a bare `switch-client -t gamma` moved the wrong client. With this patch, opening the picker from `alpha` and selecting `gamma` moved only the initiating client; the client on `beta` remained in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ The zsh `eval` has to come after `compinit`, since the script registers itself w
 > bind-key -n M-p run-shell -C 'display-popup -c "#{client_tty}" -e "CCMUX_CLIENT_TTY=#{client_tty}" -E -w 80% -h 75% "ccmux"'
 > ```
 >
-> The wrapper captures the client that opened the popup, so selecting a session never switches another attached client. The picker exits after selection, so the popup closes itself and drops you straight into that pane. (`display-popup` requires tmux 3.2+.)
+> The wrapper captures the client that opened the popup, so selecting a session never switches another attached client. The picker exits after selection, so the popup closes itself and drops you straight into that pane. (This binding needs tmux 3.3+, since `display-popup -e` arrived in 3.3.)
 
 ## 🎮 Usage
 
--- a/README.md
+++ b/README.md
@@ -102,13 +102,13 @@ The zsh `eval` has to come after `compinit`, since the script registers itself w
 >
 > ```tmux
 > # Prefix + C-p: open ccmux in a centered popup
-> bind-key C-p display-popup -E -w 80% -h 75% "ccmux"
+> bind-key C-p run-shell -C 'display-popup -c "#{client_tty}" -e "CCMUX_CLIENT_TTY=#{client_tty}" -E -w 80% -h 75% "ccmux"'
 >
 > # Or skip the prefix entirely (Alt+p from any pane)
-> bind-key -n M-p display-popup -E -w 80% -h 75% "ccmux"
+> bind-key -n M-p run-shell -C 'display-popup -c "#{client_tty}" -e "CCMUX_CLIENT_TTY=#{client_tty}" -E -w 80% -h 75% "ccmux"'
 > ```
 >
-> The picker exits after you select a session, so the popup closes itself and drops you straight into that pane. (`display-popup` requires tmux 3.2+.)
+> The wrapper captures the client that opened the popup, so selecting a session never switches another attached client. The picker exits after selection, so the popup closes itself and drops you straight into that pane. (`display-popup` requires tmux 3.2+.)
 
 ## 🎮 Usage
 
--- a/src/daemon/spawn-command.ts
+++ b/src/daemon/spawn-command.ts
@@ -1,11 +1,14 @@
 import { accessSync, constants, statSync } from "node:fs";
 import { isAbsolute } from "node:path";
 import type { AgentDef } from "../lib/agents";
+import { CLIENT_TTY_PATTERN } from "../lib/tmux-client";
 import {
   isUntrackedMode,
   UNTRACKED_MODES,
   type UntrackedMode,
 } from "./worktree-move-changes";
+
+export { CLIENT_TTY_PATTERN } from "../lib/tmux-client";
 
 /**
  * Requested split direction, in tmux's own vocabulary: `"h"` splits the
@@ -95,8 +98,6 @@ export function normalizeTarget(
  * anything outside that shape is a caller mistake worth naming rather than a
  * flag to hand to tmux.
  */
-export const CLIENT_TTY_PATTERN = /^\/dev\/[A-Za-z0-9._/-]{1,64}$/;
-
 /** Validate a wire tmux-client-tty field (`callerTty`). */
 export function normalizeClientTty(
   value: unknown,
--- a/src/daemon/spawn-command.ts
+++ b/src/daemon/spawn-command.ts
@@ -8,8 +8,6 @@ import {
   type UntrackedMode,
 } from "./worktree-move-changes";
 
-export { CLIENT_TTY_PATTERN } from "../lib/tmux-client";
-
 /**
  * Requested split direction, in tmux's own vocabulary: `"h"` splits the
  * target pane left/right (tmux `-h`), `"v"` splits it top/bottom (tmux
@@ -92,13 +90,10 @@ export function normalizeTarget(
 }
 
 /**
- * The only accepted shape for a tmux client tty (`callerTty`). tmux reports
- * `#{client_tty}` as an absolute device path (`/dev/ttys004` on macOS,
- * `/dev/pts/3` on Linux), and the value travels straight into a tmux argv, so
- * anything outside that shape is a caller mistake worth naming rather than a
- * flag to hand to tmux.
+ * Validate a wire tmux-client-tty field (`callerTty`). Anything outside the
+ * accepted shape is a caller mistake worth naming rather than a flag to hand
+ * to tmux.
  */
-/** Validate a wire tmux-client-tty field (`callerTty`). */
 export function normalizeClientTty(
   value: unknown,
   field = "callerTty",
--- a/src/tui/App.test.tsx
+++ b/src/tui/App.test.tsx
@@ -42,7 +42,10 @@ mock.module("./utils/sse", () => ({
   },
 }));
 
-const switchToPaneSpy = mock(async (_target: string): Promise<boolean> => true);
+type SwitchToPaneResult = boolean | "client-unavailable";
+const switchToPaneSpy = mock(
+  async (_target: string): Promise<SwitchToPaneResult> => true,
+);
 const sendKeysSpy = mock(
   async (
     _target: string,
@@ -1593,6 +1596,23 @@ describe("App pane-switch feedback and server scoping", () => {
       await renderWithSession();
       await selectFirstRowAndEnter();
       expect(setup.captureCharFrame()).toContain("Failed to switch");
+      expect(exitSpy).not.toHaveBeenCalled();
+    } finally {
+      restoreExit();
+      restoreFetch();
+    }
+  });
+
+  it("one-shot picker: an unknown caller client is refused with a specific toast", async () => {
+    switchToPaneSpy.mockImplementation(async () => "client-unavailable");
+    const restoreFetch = withServerInfo(null);
+    const { exitSpy, restore: restoreExit } = withExitSpy();
+    try {
+      await renderWithSession();
+      await selectFirstRowAndEnter();
+      const frame = setup.captureCharFrame();
+      expect(frame).toContain("Cannot switch: unable to identify");
+      expect(frame).toContain("this picker's tmux client");
       expect(exitSpy).not.toHaveBeenCalled();
     } finally {
       restoreExit();
--- a/src/tui/App.test.tsx
+++ b/src/tui/App.test.tsx
@@ -22,6 +22,7 @@ import { MAX_TURNS, renderTurns } from "../daemon/transcript-read";
 import { realpathSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
+import type { SwitchToPaneResult } from "./utils/client-switch";
 
 // Capture SSE callbacks so tests can fire events
 let sseCallbacks: SSECallbacks | null = null;
@@ -42,7 +43,6 @@ mock.module("./utils/sse", () => ({
   },
 }));
 
-type SwitchToPaneResult = boolean | "client-unavailable";
 const switchToPaneSpy = mock(
   async (_target: string): Promise<SwitchToPaneResult> => true,
 );
@@ -1589,13 +1589,15 @@ describe("App pane-switch feedback and server scoping", () => {
   }
 
   it("one-shot picker: a failed pane switch shows a toast and does not exit", async () => {
-    switchToPaneSpy.mockImplementation(async () => false);
+    switchToPaneSpy.mockImplementation(async () => "switch-failed");
     const restoreFetch = withServerInfo(null); // fail-open: same-server guard passes
     const { exitSpy, restore: restoreExit } = withExitSpy();
     try {
       await renderWithSession();
       await selectFirstRowAndEnter();
-      expect(setup.captureCharFrame()).toContain("Failed to switch");
+      expect(squish(setup.captureCharFrame())).toContain(
+        squish("Failed to switch: pane or client unavailable"),
+      );
       expect(exitSpy).not.toHaveBeenCalled();
     } finally {
       restoreExit();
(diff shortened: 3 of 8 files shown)


src/tui/utils/client-switch.ts
```diff
@@ -0,0 +1,35 @@
...
+  target: string,
+): Promise<SwitchToPaneResult> {
+  const capturedClientTty = process.env.CCMUX_CLIENT_TTY;
+  const clientTty =
+    capturedClientTty === undefined
+      ? await resolveCurrentTmuxClientTty()
+      : capturedClientTty;
+  if (!clientTty || !CLIENT_TTY_PATTERN.test(clientTty)) {
+    return "client-unavailable";
+  }
```
epilande: This fallback is the half of the bug the PR doesn't close.

`resolveCurrentTmuxClientTty()` states its own contract at `tmux-client.ts:54-61`: anything that moves the returned client "has to verify membership itself with `listTmuxClientTtys`, otherwise it yanks a client that was never involved." Here it goes straight into `switch-client -c`.

Not theoretical. From a pane with two clients on one session, the moved client tracks activity ordering and flips with it:

```
B1: ttys100 newest   -> ttys100 moved
B2: ordering flipped -> ttys099 moved
```

`resolveSpawnFocusArgv` (`spawn-command.ts:992-1011`) already does this, probing `listTmuxClientTtys` first. Deriving from `$TMUX`'s session via `list-clients -t` works too, and has no best-effort fallback.

One dead end already ruled out: inside a popup, `display-message -t "$TMUX_PANE" -p '#{client_tty}'` returns empty.

Fine as a follow-up. Raising it while the file is new.

src/tui/App.tsx
```diff
@@ -275,11 +275,14 @@ export function App(props: AppProps) {
...
-        // Pane is gone (daemon holds the stale row until its liveness sweep).
-        // Surface it instead of exiting the one-shot picker as if it worked.
-        store.actions.showToast("Failed to switch: pane is gone");
+    switchToPane(pane).then((result) => {
+      if (result !== true) {
+        // Surface refusal/failure instead of exiting the picker as if it worked.
+        store.actions.showToast(
+          result === "client-unavailable"
+            ? "Cannot switch: unable to identify this picker's tmux client"
+            : "Failed to switch: pane is gone",
```
epilande: nit: The refusal never names `CCMUX_CLIENT_TTY`. That's what a user needs to debug a broken binding, since dropping the `run-shell -C` wrapper puts the literal `#{client_tty}` in the env and every switch then fails. And `"switch-failed"` no longer means only a dead pane now that a client tty is in the argv.

```suggestion
        store.actions.showToast(
          result === "client-unavailable"
            ? "Cannot switch: CCMUX_CLIENT_TTY is unset or malformed (check your tmux binding)"
            : "Failed to switch: pane or client unavailable",
        );
```

src/tui/utils/client-switch.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const spawn = withSpawn([]);
+    try {
+      const result = await withClientTty("", () => switchToPane("%8"));
+
+      expect(result).toBe("client-unavailable");
+      expect(spawn.calls).toEqual([]);
+    } finally {
+      spawn.restore();
+    }
+  });
```
epilande: nit: `CLIENT_TTY_PATTERN.test` has no coverage. This case passes `""`, which short-circuits on `!clientTty` before reaching the regex, and the other four pass well-formed values. Deleting `|| !CLIENT_TTY_PATTERN.test(clientTty)` leaves all five green.

```suggestion
  });

  it("refuses a client tty that is not a device path", async () => {
    const spawn = withSpawn([]);
    try {
      const result = await withClientTty("ttys005", () => switchToPane("%8"));

      expect(result).toBe("client-unavailable");
      expect(spawn.calls).toEqual([]);
    } finally {
      spawn.restore();
    }
  });
```

src/tui/App.test.tsx
```diff
@@ -42,7 +42,10 @@ mock.module("./utils/sse", () => ({
   },
 }));
 
-const switchToPaneSpy = mock(async (_target: string): Promise<boolean> => true);
+type SwitchToPaneResult = boolean | "client-unavailable";
```
epilande: nit: Drifted from the real type in both directions. `client-switch.ts:7` exports `true | "client-unavailable" | "switch-failed"`, so this admits `false` (no longer producible) and omits `"switch-failed"`. Lines 1592 and 1700 mock `async () => false` and pass only because `App.tsx` branches on `!== true`.

```suggestion
type SwitchToPaneResult = import("./utils/client-switch").SwitchToPaneResult;
```

`mockImplementation` is typed, so this makes those two `false` mocks a compile error. Switching them to `"switch-failed"` also gives the real failure value its first App-level coverage.

src/daemon/spawn-command.ts
```diff
@@ -1,12 +1,15 @@
...
 import { isAbsolute } from "node:path";
 import type { AgentDef } from "../lib/agents";
+import { CLIENT_TTY_PATTERN } from "../lib/tmux-client";
 import {
   isUntrackedMode,
   UNTRACKED_MODES,
   type UntrackedMode,
 } from "./worktree-move-changes";
 
+export { CLIENT_TTY_PATTERN } from "../lib/tmux-client";
```
epilande: nit: No consumers. Nothing imports `CLIENT_TTY_PATTERN` from `spawn-command`, and nothing did on `main` either, so there's nothing to preserve. With the import on line 4, it also implies this module owns the pattern when `lib/tmux-client` does. Let's drop it.

src/tui/App.tsx
epilande: nit: Name `CCMUX_CLIENT_TTY` in the refusal. Dropping the `run-shell -C` wrapper puts the literal `#{client_tty}` in the env and every switch fails with no hint why. `"switch-failed"` can also mean the client now, not only the pane.

```suggestion
        store.actions.showToast(
          result === "client-unavailable"
            ? "Cannot switch: CCMUX_CLIENT_TTY is unset or malformed (check your tmux binding)"
            : "Failed to switch: pane or client unavailable",
        );
```

src/tui/App.test.tsx
epilande: nit: Drifted from the real type. It admits `false` (no longer producible) and omits `"switch-failed"`. Lines 1592 and 1700 mock `async () => false` and pass only because `App.tsx` checks `!== true`.

```suggestion
type SwitchToPaneResult = import("./utils/client-switch").SwitchToPaneResult;
```

`mockImplementation` is typed, so those two mocks become compile errors. Switching them to `"switch-failed"` covers the real value.

README.md
```diff
@@ -102,13 +102,13 @@ The zsh `eval` has to come after `compinit`, since the script registers itself w
...
-> bind-key C-p display-popup -E -w 80% -h 75% "ccmux"
+> bind-key C-p run-shell -C 'display-popup -c "#{client_tty}" -e "CCMUX_CLIENT_TTY=#{client_tty}" -E -w 80% -h 75% "ccmux"'
 >
 > # Or skip the prefix entirely (Alt+p from any pane)
-> bind-key -n M-p display-popup -E -w 80% -h 75% "ccmux"
+> bind-key -n M-p run-shell -C 'display-popup -c "#{client_tty}" -e "CCMUX_CLIENT_TTY=#{client_tty}" -E -w 80% -h 75% "ccmux"'
 > ```
 >
-> The picker exits after you select a session, so the popup closes itself and drops you straight into that pane. (`display-popup` requires tmux 3.2+.)
+> The wrapper captures the client that opened the popup, so selecting a session never switches another attached client. The picker exits after selection, so the popup closes itself and drops you straight into that pane. (`display-popup` requires tmux 3.2+.)
```
epilande: Needs tmux 3.3, not 3.2. `display-popup -e` arrived in 3.3 (tmux `CHANGES`, `3.2a TO 3.3` block); `-c` and `run-shell -C` are both fine on 3.2. On 3.2 the config loads without complaint and the key press then fails with an unknown-flag error instead of opening the popup. Ubuntu 22.04 LTS ships 3.2a and is supported into 2027.

```suggestion
> The wrapper captures the client that opened the popup, so selecting a session never switches another attached client. The picker exits after selection, so the popup closes itself and drops you straight into that pane. (This binding needs tmux 3.3+, since `display-popup -e` arrived in 3.3.)
```

src/tui/utils/client-switch.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  try {
+    const proc = Bun.spawn(
+      tmuxArgv("switch-client", "-c", clientTty, "-t", target),
```
epilande, line 23: This fallback is fine for the sidebar and a pane-launched picker: the Enter keystroke bumps the presser's activity before the switch runs, so `#{client_tty}` resolves to the right client. It is still exposed where the picker runs in a popup without `-e`: `notify-jump.ts:52-62` opens one with `-c` only, and so does anyone still on the old binding. There the picker resolves inside the popup, where its own keystrokes no longer bump activity, and one keystroke on any other attached client steers the switch.

Adding `-e CCMUX_CLIENT_TTY=<tty>` there closes it, at the cost of raising the tmux floor to 3.3 on a path every user with notifications hits (and that spawn ignores stderr, so a 3.2 user would see nothing). Your call whether that trade-off belongs in this PR or a follow-up.

Dead end to skip: inside a popup, `display-message -t "$TMUX_PANE" -p '#{client_tty}'` is empty, because the popup's pane belongs to no window.

src/tui/App.tsx
epilande: nit: Name `CCMUX_CLIENT_TTY` in the refusal. A binding that keeps `-e` but drops the `run-shell -C` wrapper (and `-c`) puts the literal `#{client_tty}` in the env, and every switch then fails with no hint why. `"switch-failed"` can also mean the client now, not only the pane (`switch-client -c /dev/ttys999` exits 1 with a live pane).

```suggestion
        store.actions.showToast(
          result === "client-unavailable"
            ? "Cannot switch: CCMUX_CLIENT_TTY is malformed or no tmux client was found (check your tmux binding)"
            : "Failed to switch: pane or client unavailable",
        );
```

The assertion at `App.test.tsx:1614-1615` matches the old string, so it moves with this.

src/tui/utils/client-switch.test.ts
epilande, line 94: nit: Nothing covers `CLIENT_TTY_PATTERN.test`. `""` and the failed-fallback case both short-circuit on `!clientTty`, and the other three pass valid ttys. Deleting the regex check leaves all five green.

```suggestion
  });

  it("refuses a client tty that is not a device path", async () => {
    const spawn = withSpawn([]);
    try {
      const result = await withClientTty("ttys005", () => switchToPane("%8"));

      expect(result).toBe("client-unavailable");
      expect(spawn.calls).toEqual([]);
    } finally {
      spawn.restore();
    }
  });
```

src/tui/App.test.tsx
epilande: nit: Drifted from the real type. It admits `false` (no longer producible) and omits `"switch-failed"`. Lines 1592 and 1700 mock `async () => false` and pass only because `App.tsx` checks `!== true`.

```suggestion
type SwitchToPaneResult = import("./utils/client-switch").SwitchToPaneResult;
```

`mockImplementation` is typed, so those two mocks become compile errors. Switching them to `"switch-failed"` covers the real value.

src/daemon/spawn-command.ts
```diff
@@ -95,8 +98,6 @@ export function normalizeTarget(
  * anything outside that shape is a caller mistake worth naming rather than a
  * flag to hand to tmux.
  */
-export const CLIENT_TTY_PATTERN = /^\/dev\/[A-Za-z0-9._/-]{1,64}$/;
```
epilande, line 98: nit: Orphaned docstring. The constant moved to `lib/tmux-client.ts` but its comment stayed, so `normalizeClientTty` has two stacked doc comments and only the nearest attaches. The shape and argv sentences belong on the constant at its new home (its one-liner there drops them); the "caller mistake worth naming" sentence belongs on `normalizeClientTty` itself.

src/daemon/spawn-command.ts
epilande: nit: No consumers, here or on `main`. Let's drop the re-export.

epilande:
Thanks for tracking this down.

I tested this locally and confirmed the approach. Capturing `#{client_tty}` at bind time is necessary, not just safer. Read inside the popup it follows activity order, and a client's activity freezes while its popup is open, so one sibling keystroke overtakes it. Two clients on one session, popup held open, sibling bumped mid-flight:

```
before:    ttys099 alpha 1788420549 | ttys100 alpha 1788420542   (presser = 099)
mid-popup: ttys099 alpha 1788420549 | ttys100 alpha 1788420551
after:     ttys099 alpha 1788420549 | ttys100 delta  1788420551   <- sibling moved
```

`run-shell -C` is required too. `display-popup -e` does not format-expand (tmux puts the value straight into the environment), so without it the env holds the literal `#{client_tty}`.

One thing outside the diff: `tmux.ts:310` and `:340` still run bare `switch-client -t`, and the comment at `:338-339` ("same approach as switchToPane") is false now that `switchToPane` targets a client. Fine as a separate issue if you'd rather keep this PR small.

README.md
TwegZhang: Verified against tmux `CHANGES` for 3.2a to 3.3: `display-popup -e` was added in 3.3. I will update the documented floor and use the suggested wording.

src/tui/utils/client-switch.ts
TwegZhang: I will keep the notification-opened popup out of this PR. Adding `-e` there would raise a path used by all notification users to tmux 3.3, while its ignored stderr makes the 3.2 failure silent. That compatibility change needs a separate treatment rather than expanding this fix.

src/tui/App.tsx
TwegZhang: I will take the suggested diagnostics and move the matching assertion. Naming `CCMUX_CLIENT_TTY` makes the malformed literal binding actionable, and `pane or client unavailable` matches both `switch-client` failure modes.

src/tui/utils/client-switch.test.ts
TwegZhang: Confirmed: the current empty-string case short-circuits before the regex. I will add the non-device-path case so removing `CLIENT_TTY_PATTERN.test` would fail the test.

src/tui/App.test.tsx
TwegZhang: I will import the real `SwitchToPaneResult` type and replace the two impossible `false` mocks with `"switch-failed"`.

src/daemon/spawn-command.ts
TwegZhang: I will move the shape and argv documentation with `CLIENT_TTY_PATTERN` and leave the caller-error sentence on `normalizeClientTty`.

src/daemon/spawn-command.ts
TwegZhang: I will drop the unused re-export.

TwegZhang:
The activity-order explanation matches the observed cross-client jump and clarifies why resolving `#{client_tty}` inside the popup is insufficient. I will address the in-diff review items: document the tmux 3.3 floor, improve the refusal messages, add regex coverage, use the real result type in tests, and clean up the moved comment/re-export. I will keep the notification popup and the two bare `switch-client -t` calls in `openDedupedCommandWindow` out of this PR because they involve separate compatibility/scope decisions, especially the silent tmux 3.2 notification failure.

README.md
TwegZhang: Fixed in 47ee5ef: README now states that this binding requires tmux 3.3+, since display-popup -e was added in 3.3.

src/tui/App.tsx
TwegZhang: Fixed in 47ee5ef: the client-unavailable toast now names CCMUX_CLIENT_TTY and the tmux binding, while switch-failed reports that either the pane or client may be unavailable. Component assertions cover both messages.

src/tui/utils/client-switch.test.ts
TwegZhang: Fixed in 47ee5ef: added a ttys005 non-device-path case that expects client-unavailable and verifies that no tmux command is spawned.

src/tui/App.test.tsx
TwegZhang: Fixed in 47ee5ef: App.test.tsx now imports the real SwitchToPaneResult type, and both impossible false mocks use switch-failed.

src/daemon/spawn-command.ts
TwegZhang: Fixed in 47ee5ef: the device-path and argv details now document CLIENT_TTY_PATTERN in tmux-client.ts, while normalizeClientTty keeps the caller-mistake explanation.

src/daemon/spawn-command.ts
TwegZhang: Fixed in 47ee5ef: removed the unused CLIENT_TTY_PATTERN re-export from daemon/spawn-command.ts.

TwegZhang:
Addressed the six in-scope review items in 47ee5ef. Verification: typecheck, targeted TUI/client-switch tests (296/296), full suite (5603 pass, 2 skip, 0 fail), build, and knip. Kept the notification-popup compatibility trade-off and the two unrelated bare switch-client sites out of scope as discussed. Ready for re-review.

PR #185, opened 2026-09-06
feat(columns): Add a `summary` column showing the agent's pane-title summary, else the prompt

### LINKED ISSUE #183: feat(columns): Add a per-agent summary column from the pane title
## Problem / motivation

Across 35 live sessions, the agent's pane title identified the session on 31 of them and the `prompt` cell on 20. The prompt loses to task-notification payloads, relayed messages from other agent sessions, post-compaction stdout, and short continuations like "commit and push". Claude, Copilot, Cursor and omp all write a generated summary or session title into `pane_title` on every turn, and ccmux shows none of it.

This is the second half of #159. The `title` column was split out of that PR (original implementation in aiwen324's addb769 and d6e66a1) so `promptLines` could land first, with the column promised as a follow-up.

## Proposed solution

A `summary` column key that renders the agent's pane title when its per-agent rule yields one, and falls back to the last prompt otherwise. One cell, one or the other; never both on a line. "Summary" rather than "title": next to `status`, `project`, `branch` and `agent`, "title" reads like something the user sets, and it is the natural name if ccmux ever grows user-assigned session names. The `EnrichedSession` field stays `paneTitle`, since that is what it holds.

### Two keys, defaults switch, columns config is the opt-out

- `prompt` stays exactly what it is today: the raw last prompt.
- `summary` is the new fallback cell.
- The default picker and sidebar layouts switch from `prompt` to `summary`, so a stock config gets the smarter cell. A hand-tuned `columns` layout that names `prompt` keeps the raw prompt until its owner changes it.
- Opting out is the column config that already moves `pane`, `time` and `pr` around, no new preference:

```bash
ccmux config set columns.row2.left prompt
ccmux config set sidebar.columns.row2.left prompt
```

- <kbd>p</kbd> (`promptDisplay`) cycles both keys to off, as it does the prompt cell today.

### With `promptLines`

The block stays the prompt; it exists to show long text, and a summary is a short title. The #159 yield rule becomes "drop the cell when it would print the prompt": a `summary` cell showing a real summary stays on the identity line with the prompt block below it, while a `prompt` cell, or a `summary` cell falling back, yields to the block as today. That is the one way to see both at once.

### Search

`summary` is a search field with highlights like `project` and `prompt`, since it is now what the default cell shows.

### Per-agent rules

Normalization is a per-agent rule on `AgentDef`, not a filter. Survey of every built-in on an isolated tmux server, one minimal turn each:

| agent    | idle                         | while working                                 | after a turn                                | what it is                                        |
| -------- | ---------------------------- | --------------------------------------------- | ------------------------------------------- | ------------------------------------------------- |
| claude   | `✳ Claude Code`             | `✳ Claude Code`                              | `✳ Say hello`                              | generated summary                                 |
| copilot  | `GitHub Copilot`             | `say hello and nothing else - GitHub Copilot` | `Implement Hello Function - GitHub Copilot` | generated summary, app name suffixed              |
| cursor   | `Cursor Agent`               | `Cursor Agent`                                | `Hello Bot`                                 | generated summary, app name before the first turn |
| omp      | `π > probe-omp-w5`           | `π ⠧ Say hello and nothing else`              | `π > Say hello and nothing else`            | session title                                     |
| codex    | `probe-codex-x7`             | `⠏ probe-codex-x7`                            | `probe-codex-x7`                            | cwd basename                                      |
| pi       | `π - probe-pi-z3`            | unchanged                                     | unchanged                                   | cwd basename                                      |
| gemini   | `◇  Ready (probe-gemini-t6)` | `✦  Working… (probe-gemini-t6)`               | `◇  Ready (…)`                              | state plus cwd, padded to 82 chars                |
| opencode | `OpenCode`                   | `OpenCode`                                    | `OpenCode`                                  | static app name                                   |

Four set something worth showing; the rest echo the cwd (already the `project` column), their own state (already `status`), or a static app name. So the rule per agent:

- claude: strip the leading `✳`; `Claude Code` reads as empty
- copilot: strip the ` - GitHub Copilot` suffix; the bare app name reads as empty
- cursor: `Cursor Agent` reads as empty
- omp: strip the `π >` / `π ⠧` prefix
- codex, pi, gemini, opencode, and custom agents: no summary, so the cell is the prompt for them. Built-ins only for now; a rule shape on `AgentConfig` for custom agents can come when someone asks.

A filter would need hostname detection (tmux seeds `pane_title` to the hostname, so today the cell shows the machine name for anything that sets nothing, and `pane-title.test.ts` locked that in as content), a cwd comparison, an app-name comparison and empty handling, and would still miss gemini's `◇  Ready (dir)`. Once one agent needs its own pattern there is a per-agent table either way, matching how `terminalRules`, `errorRules` and `readyPattern` already work.

### Still applies from the #159 review

1. Empty titles read as absent; gemini emits `""` during boot.
2. `paneTitle` is enrichment-only and does not trigger an SSE broadcast on its own, so the long-lived sidebar holds a stale summary on an idle row. Add it to the tracked-field comparison in `sessions.ts`, comparing the normalized value: codex and omp rewrite the title on every spinner frame, and the raw string would broadcast every scan tick.

The budget split and per-field `maxPromptLen` row selection from that review drop out: summary and prompt never share a line.

### Credit

The starting point is aiwen324's `title` commits (addb769, d6e66a1). Keep the credit through the squash with a `Co-authored-by` trailer.

## Alternatives considered

- A generic filter over `pane_title` (hostname, cwd, app name): see above; it misses gemini and grows a per-agent table anyway.
- Show the raw title: the hostname leak and the spinner glyphs make it noise for half the agents.
- Summary and prompt side by side on one line with a split budget: more information per row, but it needs a budget split and per-field row selection in `SessionItem`, and the block already gives the prompt room when both matter. Decided against.
- A new preference to toggle the behavior: the column config already is that knob.

## Relevant agent(s)

- Claude Code, GitHub Copilot, Cursor, oh-my-pi (rules); every other built-in defaults to no summary

## Additional context

- #159 review comment with the original survey and design: https://github.com/epilande/ccmux/pull/159#pullrequestreview-5006441250
- Starting point: aiwen324's `title` commits addb769 and d6e66a1, dropped from #159 in db037e7
- Companion cleanup of the prompt cell's own noise: #184

epilande:
## Overview

Adds a `summary` column that shows the agent's own pane-title summary when its per-agent rule yields one, and the last prompt otherwise. The default picker and sidebar layouts switch from `prompt` to `summary`; `prompt` itself is unchanged.

Resolves #183

## Motivation

Across 35 live sessions the pane title identified the session on 31 of them and the prompt cell on 20. The prompt loses to task-notification payloads, relayed messages, post-compaction stdout and short continuations like "commit and push". Claude Code, Copilot, Cursor, OpenCode and oh-my-pi all write a generated summary into `pane_title` on every turn, and ccmux showed none of it.

This is the second half of #159. The `title` column was split out of that PR so `promptLines` could land first; the starting point here is @aiwen324's two `title` commits, carried on the branch with their authorship.

### Rule (daemon-side)

- **src/lib/pane-summary.ts** - `summaryFromPaneTitle(agentType, paneTitle, paneCwd)`, the pure per-agent rule. Runs in the daemon only; the TUI reads the shipped field. Normalizes the title (ANSI stripped, whitespace collapsed), applies the agent's `match`, and reads a title equal to the machine's hostname as no summary, since tmux seeds every pane title with it
- **src/lib/agents.ts** - `AgentDef.summaryTitle`: one `match` regex per agent whose capture group is the summary, plus `empty` titles and, for omp, a cwd-basename check. Set on claude, copilot, cursor, opencode and omp. The issue's survey recorded OpenCode as a static `OpenCode`; live on 1.18.29 it writes `OC | <session title>`. Built-ins only; custom agents get no summary yet

### Column

- **src/lib/preferences.ts** - `summary` joins `COLUMN_FIELDS`
- **src/tui/components/session-columns.ts** - default layouts switch to `summary`; `withoutFlexText` beside `withoutPrompt`; `flattenToRow1` drops row 2's flex cell when row 1 already carries one, so a hand-placed `prompt` on row 1 never shares the line with the default `summary`
- **src/tui/components/SessionItem.tsx** - the prompt cell is extracted into `PromptCell`; the `summary` cell falls back to it, so older-prompt matches, transcript snippets and the `[pane]`/`[transcript]`/`[cwd]` tag still render under search. Under the `prompt` opt-out, a match that hit only the summary shows a `[summary]` tag. The `promptLines` block now renders BELOW row 2 rather than between the two rows: row 2 is the short identity line (summary, PR, time) and the block is a paragraph, so the row reads shortest to longest. Height is a sum, so nothing in the row math changes
- **src/tui/components/SessionList.tsx** - the `promptLines` yield is per row: a session with a summary is measured and drawn by the layout that keeps its cell above the block, a session without one by the layout that drops it. Both the height math and the row read the same object

### Search and daemon

- **src/tui/store.ts** - `summary` is a substring search field with `highlights.summary`; it shares the prompt's score contribution rather than adding a sixth match source, keeping the ranking tier gaps intact
- **src/daemon/server.ts**, **src/daemon/index.ts** - `enrichSession` computes `summary` and ships it beside the raw `paneTitle`. `syncPaneSummaries()` re-broadcasts a visible session whose summary changed, once per scan, and every all-client send records what it sent through `recordBroadcast`, so a turn that changes status and title together broadcasts once. Comparing the raw title would broadcast on every codex/omp spinner frame

### Review fixes (50f3cac)

- Cursor's rule reads shell-written titles as empty (`:/abs`, `/abs`, `:~/x`, `~`, bash's `user@host:~/x`). The shell writes the cwd into `pane_title` on every prompt, and a cursor pane that had not retitled yet showed it as the summary
- `lastPaneSummary` states one invariant (recorded equals what last went on the wire) and names its three writers; `onBranchPRsChanged` re-checks visibility after its await; `syncPaneSummaries` returns early with no SSE client; `normalizeTitle` strips the C0/C1 controls `stripAnsi` leaves behind
- `flattenToRow1` keeps one flexible cell even when row 2 carries both `summary` and `prompt`; the flex-text rename is finished (`maxFlexLen`, `flexOnRow`, `hasFlexText`)

### Docs

- **README.md** - `summary` row in the columns table, the opt-out (`ccmux config set columns.row2.left prompt` / `sidebar.columns.row2.left prompt`), and the per-session `promptLines` yield
- **docs/architecture.md** - summary as a fifth search source and why it shares the prompt's score

## Breaking Changes

None. A hand-tuned `columns` layout that names `prompt` keeps the raw prompt. Stock configs get the smarter cell; the two `config set` lines above restore the old subtitle.

## Testing

- [x] `bun run typecheck`, `bun run check:unused`, `bun test` (5781 pass / 0 fail)
- [x] Isolated daemon (own tmux socket, port, `CCMUX_HOME`): stock config shows the summary on a claude row and the prompt on a row with no rule, with the hostname kept off both
- [x] `config set columns.row2.left prompt` restores the raw prompt with no new preference
- [x] `promptLines: 3`: the summary stays on the identity line with the block below; the fallback row drops its cell
- [x] Search on an older prompt still renders the match line in the cell; a hit on the summary text renders with highlights
- [x] Retitling a pane from outside the agent updates the row over SSE with no other session change, and a turn that changes status and title together broadcasts once
- [x] The demo video below predates the internal rework (rule shape, daemon-side normalization, per-row layout); the behavior it shows is unchanged


epilande:
Surface demo of the `summary` column on an isolated daemon with three throwaway Claude sessions (sped up). Shown, not driven through agent states: the stock layout with the summary on the Claude row and the prompt fallback on a row whose pane title is tmux's hostname seed, the `p` cycle (inline, own row, off), a search hit on an older prompt replacing the summary with the match line, `promptLines: 3` keeping the summary on the identity line above the wrapped block while the fallback row yields its cell, and the `columns.row2.left prompt` opt-out restoring the raw prompt. Claude is the representative agent for the per-agent rule; copilot, cursor and omp rules are covered by unit tests only.

https://github.com/user-attachments/assets/66d19d37-ea36-4ba4-981d-9a792ab3c7d0

PR #188, opened 2026-09-06
fix(search): Keep the space before a windowed highlight

### LINKED ISSUE #186: fix(search): Highlight window drops the space before the matched span
### ccmux version

1.3.2 (main, 0a2945d)

### Agent

Claude Code (not agent-specific: it is a picker render bug)

### Are hooks installed?

Yes

### What happened?

In the picker's search mode, when the matched prompt is long enough that the prompt cell windows with a leading `…`, the space immediately before the highlighted span is dropped. The match reads as one word glued to the preceding one.

Observed on a throwaway Claude session whose prompt was:

> Without using any tools, write exactly one sentence explaining what a terminal multiplexer is to someone who has never opened a terminal before, keeping it plain and avoiding examples of any kind.

Searching `multiplexer` renders the row as:

```
 ● Sessions (1/1)
 / multiplexer
────────────────────────────────────────────────────────────────────────────────
 ▼ hl186-a (1)
 1 ● idle    tmp/hl186-a:HEAD  …laining what a terminalmultiplexer is to someone who has never opened a terminal before,…   Claude   v2.1.263        a:1.1  39s
```

`terminalmultiplexer` should be `terminal multiplexer`.

### Steps to reproduce

1. Start a session whose last prompt is long, with the word you will search for sitting deep enough into it that the prompt cell has to window (a leading `…` appears).
2. Run `ccmux picker` at a width where the prompt cell is the flexible filler (160 cols here).
3. Press `/` and type that word.
4. The row shows the match with the preceding space missing.

It is not a capture artifact: the same one-column loss reproduces in a headless `testRender` of `HighlightedText` inside a fixed-width row box (see Cause).

### Cause

Not `truncateHighlighted`. That function keeps the space: probing it directly on the prompt above, the character before `<b>` is `" "` at every budget (60, 80, 100, 120 columns), and its 100-column output matches the rendered cell exactly apart from the missing space.

The space is lost at render, in the interaction between three places:

- `src/tui/components/HighlightedText.tsx:40-47` emits one sibling `<text>` node per markup segment, so the pre-match context, the bold span, and the trailing context are three separate flex children.
- `src/tui/components/SessionItem.tsx:594` puts them in the prompt cell's `flexGrow={1} flexShrink={1} flexDirection="row"` box.
- `src/tui/components/SessionItem.tsx:623` and `:634` window the markup to `ctx.maxPromptLen`, which is a layout BUDGET (`SessionItem.tsx:900`), not the box's real width. `truncateHighlighted` (`src/tui/utils/format.ts:146`) fills that budget exactly whenever the match is deep enough to need a leading `…`.

So the row's content ends up a column or two wider than the box actually is. Yoga takes those columns off the widest shrinkable sibling, which is the pre-match `<text>`, and that node then re-wraps at its new width. The whitespace at the wrap point is consumed, and the word after it moves to a line the height-1 row never shows.

Rendering `HighlightedText` inside a fixed-width row box makes the mechanism visible. The markup is `…laining what a terminal <b>multiplexer</b> is to someone` (50 visible columns):

```
box=50: ["…laining what a terminal multiplexer is to someone", ""]
box=49: ["…laining what a terminalmultiplexer is to someone",  ""]
box=48: ["…laining what a         multiplexeris to someone",  "terminal"]
```

At 49 the shrink is exactly one column, the trailing space is the wrap point, and it vanishes while `terminal` still fits. At 48 the same wrap moves `terminal` to a second line and the gap is plainly a wrap, not a clip.

### Expected

`…laining what a terminal multiplexer is to someone…`, with the space before the highlighted span intact. One fix direction: render the whole highlighted line as a single `<text>` with inline `<b>` spans, so any shrink happens once, on one node, at the end of the line.

### OS and tmux version

macOS 26.5.2, tmux 3.6a

Predates #185: seen in that PR's demo video, but neither the per-segment render nor the windowing is touched there. The per-segment `HighlightedText` render dates to the initial release (d946239); the highlight windowing that makes the content almost-but-not-quite fit arrived with the whole-session search in #3 (a84da12). Before windowing, an overflowing highlight shrank hard and rendered visibly garbled, so the one-column case is what the windowing exposed.

Reproduced on main (0a2945d) against an isolated daemon and a single throwaway Claude session.

epilande:
## Summary

Keeps the space before a search highlight when the prompt cell windows with a leading `…`.

A windowed match used to render `terminalmultiplexer` because `HighlightedText` emitted one Text node per markup segment and Yoga shrank the pre-match sibling.

## Related issue

Closes #186

## Testing

- [x] `bun test` full suite: 5694 pass, 3 skip, 0 fail
- [x] Regression: `HighlightedText` in a 49-col `flexGrow`/`flexShrink` row keeps `terminal multiplexer` (fails on main)
- [x] Verified picker still renders in a detached tmux session (`ccmux-verify` at 200x50 and 60x30). Daemon was stopped, so there was no live long-prompt search row to type `/ multiplexer` against. The glued-word path is the headless row-box case above.
- [x] CI / ci required check passed

## Notes for reviewers

The glued-word case is the regression test: `HighlightedText` in a 49-col `flexGrow`/`flexShrink` row with the 50-col windowed markup from the issue (`…laining what a terminal <b>multiplexer</b> is to someone`).

On this branch, shrink clips the tail:

```
box=50: …laining what a terminal multiplexer is to someone
box=49: …laining what a terminal multiplexer is to
```

On main, box=49 was `terminalmultiplexer`.

`span` `fg` is applied through a small `as unknown as` helper. `@opentui/solid` 0.1.97 types `SpanProps` as `ComponentProps<{}, TextNodeRenderable>`, which drops `fg` even though `TextNodeRenderable` accepts it.

<div><a href="https://cursor.com/agents/bc-f6d5b86f-1957-4ab2-9b05-3e1ac7f762b5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f6d5b86f-1957-4ab2-9b05-3e1ac7f762b5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


epilande:
Live before/after in the picker: same session, same 120-column terminal, same search for `multiplexer is`. Playback is sped up.

First take is main at a6d7265. The windowed prompt renders `terminalmultiplexer is`, with the space before the match eaten. Second take is this branch at 211dd1a: `terminal multiplexer is`, space intact, and the match bold in the theme's yellow.

One visible side effect of the single Text node: the tail now clips at the end of the line, so the trailing `…` main printed at this width is gone.

https://github.com/user-attachments/assets/a72fab75-e051-45f6-8b17-d6581ca99920

PR #189, opened 2026-09-07
fix(columns): Strip task notifications, handoffs and ANSI from the prompt cell

### LINKED ISSUE #184: fix(columns): Strip task notifications, handoffs and ANSI from the prompt cell
## Problem / motivation

`stripCommandMarkup` in `session-columns.ts` handles two shapes of Claude's structured prompt: `<command-name>` + `<command-args>`, and `<local-command-stdout>`. Everything else passes through raw, so the `prompt` cell and the new `promptLines` block paint:

- task-notification payloads (`<task-notification> <task-id>…`) as the row's prompt
- relayed `[ccmux handoff]` messages from other agent sessions
- post-compaction `<local-command-stdout>` content with its ANSI escapes intact, so a row shows a raw escape byte

This was the biggest source of noise in the 35-session comparison on #159 (the prompt cell identified the session on 20 rows against the pane title's 31), and the wrapped block makes it more visible rather than less.

## Proposed solution

Extend the stripper to:

- strip ANSI from the `<local-command-stdout>` branch (`strip-ansi.ts` already exists in `src/lib`)
- reduce a task-notification payload to its `<summary>` when present, else drop it as a prompt
- treat a relayed handoff message by its frozen provenance header, showing the relayed text rather than the header

With a test per shape, since the first two are exact strings Claude emits today.

## Alternatives considered

- Leave it to the `summary` column (#183): that helps only the four agents that set a pane title, and the prompt cell stays the fallback for the rest.

## Relevant agent(s)

- Claude Code (the structured prompt shapes); handoff messages reach every agent

## Additional context

- Noted in the #159 review as an "our side" fix: https://github.com/epilande/ccmux/pull/159#pullrequestreview-5006441250

epilande:
## Overview

Reduces the noisy shapes Claude writes into the transcript (task notifications, relayed handoffs, local-command and `!` shell output) to the text the user actually cares about before they reach the prompt, promptLines and summary-fallback cells, and strips real terminal escapes from them.

Resolves #184

## Motivation

Rows after a background task, a handoff, or a shell command painted raw XML tags, provenance headers, and escape bytes like `[?25l` or `]8;;https://…` in the prompt cell. The prompt cell exists to show intent at a glance, and these shapes hid it behind machinery.

### Prompt reduction

- **session-columns.ts** - `stripCommandMarkup` handles three more shapes: a `<task-notification>` reduces to its `<summary>` (dropped entirely with none), a `[ccmux handoff]` message shows the relayed payload instead of its frozen header, and `<bash-stdout>` (with `<bash-stderr>` as fallback, `<bash-input>` shown as `! <cmd>`) is reduced like `<local-command-stdout>`. Claude Code 2.1.x records a `!` command as the `<bash-*>` pair, not `<local-command-stdout>`, which only built-ins like `/model` still write. `<task-notification>` is tested first because it is the outermost wrapper and its `<result>` can quote any tag below it. Both stdout shapes go through `stripTerminalNoise`: `stripAnsi`, a control sweep that turns the residue into spaces, then a whitespace re-collapse.

### Escape stripping

- **strip-ansi.ts** - `ANSI_PATTERN` matched CSI with numeric parameters only, so the escapes a spinner actually emits (`\x1b[?25l` cursor hide, OSC 8 hyperlinks, title sets) survived as printable text. It now matches private-parameter CSI and BEL/ST-terminated string sequences. Two-byte `ESC x` forms stay unmatched on purpose so a lone ESC never eats the next character, and an unterminated string sequence matches nothing rather than the rest of the line. Every other caller gets strictly more removal.
- **strip-ansi.ts** - Adds `stripTerminalNoise`, the strip + control sweep + collapse chain the prompt cell and the pane-title summary both need.
- **pane-summary.ts** - `normalizeTitle` calls `stripTerminalNoise` instead of carrying its own copy of the chain.

### Tests

- **session-columns.test.ts** - One case per shape: ANSI in local-command stdout, task notification with and without a summary, handoff with and without a payload, the exact `<bash-stdout>` shape observed live, the stderr fallback, the both-empty drop, `<bash-input>`, stdout winning over input, and a notification whose `<result>` quotes a `<bash-stdout>` tag still reducing to its summary.
- **strip-ansi.test.ts** - Private-mode CSI, OSC by BEL and by ST, lone ESC preserved, unterminated OSC preserved, and `stripTerminalNoise` reduction, stray-byte separators and noise-only input.

## Breaking Changes

None

## Testing

- [x] `bun test src/tui/components/session-columns.test.ts` (154 pass)
- [x] `bun test` (full suite, 5808 pass)
- [x] `bun run typecheck`
- [x] Live picker capture from this branch's build against a real Claude Code 2.1.263 session running `! printf` with an OSC 8 hyperlink, cursor hide/show and a BEL: the row went from `<bash-stdout> ]8;;https://example.com LINKED ]8;; …` to `LINKED green file`

### DIFF
--- a/src/lib/pane-summary.test.ts
+++ b/src/lib/pane-summary.test.ts
@@ -208,7 +208,8 @@ describe("summaryFromPaneTitle", () => {
     });
 
     it("strips the controls stripAnsi leaves behind", () => {
-      // `stripAnsi` handles CSI only. A BEL, a lone ESC and a C1 byte (here
+      // `stripAnsi` handles well-formed sequences. A BEL, a lone ESC and a
+      // C1 byte (here
       // U+0085 NEL, which JS `\s` does not match) would otherwise ride into a
       // rendered cell. Each becomes a space, so the words stay apart.
       expect(
--- a/src/lib/pane-summary.ts
+++ b/src/lib/pane-summary.ts
@@ -32,12 +32,12 @@ const HOSTNAME = hostname();
  * also unwraps Claude's `<command-name>` log markup, which is a property of
  * Claude's JSONL transcript and has no business being applied to a title.
  *
- * `stripAnsi` handles CSI only, so the control sweep after it is what catches
- * the rest: a lone ESC, a BEL, and the C1 block (U+0080-U+009F), which is the
- * practical residue once tmux's own title validation has had its say. Each
- * becomes a SPACE rather than nothing, so the `\s+` collapse behind it still
- * sees a separator: `\n` and `\t` fall under both rules, and deleting them
- * outright would weld two words together.
+ * `stripAnsi` handles well-formed sequences, so the control sweep after it is
+ * what catches the rest: a lone ESC, a BEL, and the C1 block (U+0080-U+009F),
+ * which is the practical residue once tmux's own title validation has had its
+ * say. Each becomes a SPACE rather than nothing, so the `\s+` collapse behind
+ * it still sees a separator: `\n` and `\t` fall under both rules, and
+ * deleting them outright would weld two words together.
  */
 function normalizeTitle(text: string): string {
   return stripAnsi(text)
--- a/src/lib/pane-summary.ts
+++ b/src/lib/pane-summary.ts
@@ -2,7 +2,7 @@ import { hostname } from "node:os";
 import { basename } from "node:path";
 import { BUILTIN_AGENTS } from "./agents";
 import type { SummaryTitleRule } from "./agents";
-import { stripAnsi } from "./strip-ansi";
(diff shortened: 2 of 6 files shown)


epilande:
Screen recording, sped up: the same `!` shell turn painted by the main build, then by this branch. Before, the prompt cell shows the raw `<bash-stdout>` wrapper plus the OSC 8 and SGR escape residue. After, it reads `LINKED green file`. The row is configured with `columns.row2.left: ["prompt"]` because the default row 2 renders the pane-title summary instead.

https://github.com/user-attachments/assets/5e5502bd-bf57-4adf-800b-3ba1414cadcb

Issue #181, opened 2026-09-04
fix(picker): Pin the notification popup to its client and drop the tmux 3.3 requirement


epilande:
## Summary

#180 stops the picker from switching the wrong tmux client, but only for users who update their keybinding to the new `run-shell -C` form. Three gaps remain.

### 1. The notification popup still targets the most-recently-active client

`src/daemon/notify-jump.ts` resolves the client tty itself and opens `display-popup -c <tty> -E ccmux`, but never passes that tty into the popup. The picker inside falls back to `display-message -p '#{client_tty}'` and inherits the exact bug #180 fixed. Deferred in #180 because adding `-e` would raise the notification path to tmux 3.3, and that spawn ignores stderr, so a 3.2 failure would be silent.

### 2. The tmux 3.3 floor is avoidable

Only `display-popup -e` needs 3.3. `-c`, `-E`, `-w`, `-h`, and `run-shell -C` all exist in 3.2 (checked against 3.2a `cmd-display-menu.c` and `cmd-run-shell.c`). Passing the tty as an argument instead of an environment variable works on 3.2:

```
bind-key C-p run-shell -C 'display-popup -c "#{client_tty}" -E -w 80% -h 75% "ccmux --client-tty #{client_tty}"'
```

Two details for whoever picks this up:

- `display-popup` does not format-expand its `-E` command string, but `run-shell -C` expands the whole line before parsing it, so `#{client_tty}` inside the quoted command is already substituted by the time `display-popup` sees it.
- `-c "#{client_tty}"` is redundant. `run-shell -C` hands the pressing client's command-queue state to the nested `display-popup`, so the popup already opens on the right client. Harmless to keep.

### 3. Users on the old binding get no signal

With `CCMUX_CLIENT_TTY` unset, `switchToPane` falls through to `resolveCurrentTmuxClientTty()`, which is the pre-#180 behavior. Single-client users are unaffected. Multi-client users still move the wrong client, and nothing tells them: the toast only fires when tmux reports no client at all.

## Proposal

- Add `ccmux --client-tty <tty>` on the picker entry and keep `CCMUX_CLIENT_TTY` as a fallback so the #180 binding keeps working.
- In `notify-jump.ts`, pass the already-resolved tty as `--client-tty` on the popup command. No version gate needed, since nothing in that line requires 3.3.
- README: document the argument form and restore the tmux 3.2+ note.
[...]

PR #190, opened 2026-09-07
fix(picker): Resolve a popup's client from its own session, not the binding

epilande:
## Summary

#182 made the picker refuse every uncaptured switch inside a `display-popup` on a server with two or more clients attached, because `#{client_tty}` there names whichever OTHER client typed last. Correct, but it broke the previously documented plain binding (`display-popup -E ... "ccmux"`) the moment a second terminal attached, and the fix was a `run-shell -C` binding everyone had to adopt.

The popup already knows where it came from. tmux sets `$TMUX` per popup as `<socket>,<pid>,<session id>`, and the third field is the session the popup command targeted, which is the pressing client's session for any binding without `-t`. The resolver now lists that session's clients:

- exactly one: that is the launcher, pin the switch to it (the case #182 refused)
- two or more: refuse with `shared-session-popup`, the one case that genuinely needs `--client-tty`
- lookup failed, timed out, unparseable `$TMUX`, or a socket override pointing elsewhere: refuse with `popup-client-unknown`. This is its own refusal on purpose: collapsing it into `no-client` would let `openDedupedCommandWindow` create the agents window untargeted, in whichever session typed last.

Captured ttys (`--client-tty`, `CCMUX_CLIENT_TTY`) still win before any probe. Plain panes, the sidebar, and a picker outside tmux keep the old current-client guess, at the same number of spawns as before.

- `src/lib/legacy-popup.ts` becomes `src/lib/popup-client.ts`: popup detection plus the `$TMUX` session lookup, facts only. Policy (one/many/none) stays in `resolvePinnedTmuxClientTty`.
- `popup-binding-check.ts` and the `ccmux setup` old-binding warning are removed. The old binding is no longer wrong.
- README's default binding is the plain `display-popup` form again. The `run-shell -C ... --client-tty` form stays documented for two terminals on the same session, or a binding that opens the popup on a session other than the pressing terminal's.
- New toast copy for the two popup refusals; picker `--help` for `--client-tty` reworded.

## Related issue

Follow-up to #182 (refs #181, #180). Unreleased, so no user upgrade note needed beyond README.

## Testing

- [x] `bun test` passes (5815 pass, 0 fail, 200 files)

`bun run check:unused` and `bun run build` are clean.

Live on an isolated tmux 3.6a server (`-L popfix`, isolated daemon on `CCMUX_PORT=2271`, two clients attached via `script -q /dev/null tmux attach`, a pane-tracked stand-in agent for rows):

- Plain binding shape, two clients on two different sessions, popup opened from the LESS recently active client: resolver returned the launching client's tty where `#{client_tty}` returned the other one. A real picker Enter moved only the launching client (`list-clients` before: `ttys097 alpha / ttys098 gamma`, after: `ttys097 beta / ttys098 gamma`).
- Two clients on the SAME session: `shared-session-popup`, the new toast rendered, nobody moved.
- `--client-tty` on the popup command: the capture beat the popup lookup, which would have named a different client.
- Server started with `TMUX=/tmp/bogus,1,7` exported in its environment: the popup's `$TMUX` still named its own socket and the launching session, so the value is set per session, not inherited (the `TMUX_PANE` lesson from #182 does not apply here).
- Sidebar in a plain pane: switched as before.

Not verified live: tmux 3.2 (`list-clients -t <session>` and `display-popup` both exist there per 3.2a source, same caveat as #182), and after the review round's reorder only the resolver was re-probed rather than a full picker Enter (the success path is unchanged and the new toast has a renderer test).

## Notes for reviewers

- One shape where this is strictly worse than #182: a binding that opens the popup with `-t` on a session other than the pressing terminal's. `$TMUX` then names that target session, and the resolver moves ITS client, silently, where #182 refused. Nothing inside the popup can detect it, so README tells that user to use `--client-tty`. Judged rare enough to accept.
- The socket-override mismatch arm (`popup-client-unknown` via `CCMUX_TMUX_SOCKET` pointing at another server) is unreachable from a picker or sidebar today, since `activeTmuxSocketOverride()` ignores the override while `$TMUX` is set. It is covered by unit tests and kept as the guard for a caller that is daemon-marked.
- `src/lib/tty.ts` still says `TMUX_PANE` is unset inside a popup. Measured: it is set to a phantom pane id that `list-panes -a` does not contain. Out of scope here, nothing reads it.


epilande:
End-to-end check on tmux 3.6a. Screen recording, sped up.

Isolated tmux server, three sessions, two attached terminals: one on `alpha`, one on `bravo`. The popup is opened from the `alpha` terminal with the plain binding (`display-popup -E ... ccmux picker`, no `--client-tty`). While the picker is up the other terminal is typed into, so tmux's own `#{client_tty}` resolves to `/dev/ttys104`, the terminal that did not open the popup.

1. `origin/main` at 3cfc14d: Enter refuses with "this popup was given no client tty and several clients are attached". Nothing moves.
2. This PR at 5403a95: Enter switches only the launching terminal. `list-clients` afterwards shows `/dev/ttys105` on `charlie` and `/dev/ttys104` still on `bravo`.
3. A third terminal attaches to `alpha`, so the launching session has two. Enter refuses with the `shared-session-popup` message, the case that stays genuinely ambiguous.

Rows come from real pane-tracked detection against stand-in agent panes, with a daemon on its own port and tmux socket. Suite at 5403a95: 5817 pass, 0 fail; typecheck, knip and build green.

https://github.com/user-attachments/assets/3df17d8f-9299-404f-95a3-6dba30b16f75

PR #192, opened 2026-09-08
feat(preview): Show the session summary in the preview header

florian-schneider-cohere:
## Summary

- The picker preview header showed the project, cwd, and branch/version/target metadata — but not the **agent's own summary of the session**, so two sessions in one project were indistinguishable in the preview.
- Agents with a `summaryTitle` rule (Claude, OpenCode, Cursor, oh-my-pi, Copilot) already ship that summary as `session.summary` (derived daemon-side from the pane title), and the `summary` column renders it in list rows. The preview header simply never used it.
- Now the summary renders as its own header line between the project row and the cwd; the header budget grows from four rows to five when one is present. Sessions without one render exactly as before.
- The summary is pre-truncated to the header's inner width (`truncateText(title(), separatorWidth())`, the idiom `ConfirmationDialog`/`ContextMenu`/`DropdownField` use), so a summary wider than the preview clips with an ellipsis instead of wrapping inside the fixed-height box.
- The line sits in an always-mounted wrapper box: the summary can land over SSE after first mount, and a bare `<Show>` mounted late would append to the END of the parent box — the same opentui insertion-anchor trap the transcript slot in `BackgroundPeek` guards against. (Note: the late-arriving test passes with the wrapper removed too, so it pins the layout but does not by itself distinguish wrapper vs bare `<Show>`; the wrapper is kept for the reason above.)

## Test plan

- [x] `bun run typecheck`
- [x] `bun test` — 200 files / 5841 tests pass, including four new `Preview.test.tsx` cases:
  - summary on its own header line, separator pinned at `metadataLine + 1`
  - four-row header preserved without a summary, separator pinned at `metadataLine + 1`
  - late-arriving summary renders in the header, separator pinned at `metadataLine + 1`
  - a summary wider than the header truncates to one line (no wrapped continuation, rows below keep their slots) — verified this test fails without the `truncateText` fix
- [x] Live render per AGENTS.md: isolated tmux server + daemon (`CCMUX_TMUX_SOCKET`, `CCMUX_PORT`, `CCMUX_HOME`), fake OpenCode session with pane title `OC | Fix flaky auth tests across suites`, `ccmux picker --preview`:

```
 ● Sessions (1)
 ▼ mling_ir (1)                                                     │ mling_ir                                   ● idle
 1 ● idle  embed-flo-sch…mling_ir  Fix flaky auth tes…  OpenCode    │ Fix flaky auth tests across suites
                                                                    │ /data/1d/embed-flo-schneider/mling_ir
                                                                    │ v9.4 · work:1.1
                                                                    │ ───────────────────────────────────────────
```

Also re-captured at 100 cols; header layout holds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,8 @@ The daemon starts automatically the first time you run a ccmux command (picker, 
 
 Press <kbd>P</kbd> to split the picker and preview the highlighted session's live pane content side by side. Press <kbd>Tab</kbd> to focus the preview and act in place: your keystrokes go straight to that agent's pane, so you can approve a permission, answer a question, or type a follow-up without ever leaving ccmux.
 
+The preview header names the session: the project in bold, then the agent's own summary of the session on agents that publish one (Claude Code, OpenCode, Cursor, oh-my-pi, and Copilot do; on the rest the line is omitted), then the working directory and the branch/version/target metadata.
+
 When the session has agents running, an **Agents** section lists each one with its runtime. Finished agents drop off the list.
 
 https://github.com/user-attachments/assets/7e0d42b3-4e7b-43b8-8d06-72a2d69dd694
--- a/src/tui/components/Preview.tsx
+++ b/src/tui/components/Preview.tsx
@@ -34,6 +34,7 @@ import {
   formatSubagentName,
   formatVersion,
   shortenCwd,
+  truncateText,
 } from "../utils/format";
 
 /**
@@ -413,6 +414,14 @@ export const Preview: Component<PreviewProps> = (props) => {
   const AGENTS_SHOWN_MAX = 4;
   const liveSubagents = () => props.session?.subagents ?? [];
 
+  /**
+   * The agent's own summary of what this session is doing
+   * (`EnrichedSession.summary`, derived daemon-side from the pane title),
+   * on the agents that publish one. Renders as its own header line so the
+   * preview names the session's work, not just the project it sits in.
+   */
+  const title = () => props.session?.summary ?? null;
+
   const metadataLine = () => {
     const s = props.session;
     if (!s) return "";
@@ -446,7 +455,7 @@ export const Preview: Component<PreviewProps> = (props) => {
         when={props.session}
         fallback={<text fg={theme.overlay}>Select a session to preview</text>}
       >
-        <box height={4} flexDirection="column">
+        <box height={title() ? 5 : 4} flexDirection="column">
           <box flexDirection="row">
             <box flexGrow={1}>
               <text fg={theme.text}>
@@ -456,6 +465,17 @@ export const Preview: Component<PreviewProps> = (props) => {
             <text fg={statusColor()}>
               {statusIcon()} {statusText()}
             </text>
+          </box>
+          {/* Stable wrapper: a summary can land after first render (enrich
+              arrives over SSE), and a bare <Show> mounted late would append
+              to the END of the parent box (opentui insertion anchor),
+              landing the title under the separator. The always-mounted box
+              pins its slot in the column order. Same fix pattern as the
+              transcript slot in BackgroundPeek. */}
+          <box flexDirection="column">
+            <Show when={title()}>
+              <text fg={theme.text}>{truncateText(title() ?? "", separatorWidth())}</text>
+            </Show>
           </box>
           <text fg={theme.subtext}>
             {shortenCwd(props.session!.paneCwd ?? props.session!.cwd)}
--- a/src/tui/components/Preview.test.tsx
+++ b/src/tui/components/Preview.test.tsx
@@ -77,6 +77,119 @@ describe("Preview", () => {
     expect(frame).toContain("~/Code/myapp");
   });
 
+  it("shows the session summary on its own header line", async () => {
+    const home = process.env.HOME || "";
+    const frame = await renderPreview(
+      mockEnrichedSession({
+        project: "myapp",
+        cwd: `${home}/Code/myapp`,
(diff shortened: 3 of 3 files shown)


src/tui/components/Preview.tsx
```diff
@@ -457,6 +465,17 @@ export const Preview: Component<PreviewProps> = (props) => {
...
           </box>
+          {/* Stable wrapper: a summary can land after first render (enrich
+              arrives over SSE), and a bare <Show> mounted late would append
+              to the END of the parent box (opentui insertion anchor),
+              landing the title under the separator. The always-mounted box
+              pins its slot in the column order. Same fix pattern as the
+              transcript slot in BackgroundPeek. */}
+          <box flexDirection="column">
+            <Show when={title()}>
+              <text fg={theme.text}>{title()}</text>
```
epilande: Let's pre-truncate to the header's inner width, the idiom `ConfirmationDialog`, `ContextMenu`, and `DropdownField` use. `truncateText` needs adding to the `../utils/format` import at the top of the file.

```suggestion
              <text fg={theme.text}>{truncateText(title(), separatorWidth())}</text>
```

README.md
```diff
@@ -169,6 +169,8 @@ The daemon starts automatically the first time you run a ccmux command (picker,
 
 Press <kbd>P</kbd> to split the picker and preview the highlighted session's live pane content side by side. Press <kbd>Tab</kbd> to focus the preview and act in place: your keystrokes go straight to that agent's pane, so you can approve a permission, answer a question, or type a follow-up without ever leaving ccmux.
 
+The preview header names the session: the project in bold, then the agent's own session title on agents that publish one (Claude Code, OpenCode, Cursor, oh-my-pi, and Copilot do; the rest fall back to the prompt), then the working directory and the branch/version/target metadata.
```
epilande:
```suggestion
The preview header names the session: the project in bold, then the agent's own summary of the session on agents that publish one (Claude Code, OpenCode, Cursor, oh-my-pi, and Copilot do; on the rest the line is omitted), then the working directory and the branch/version/target metadata.
```

epilande (changes requested):
Thanks for your contribution, this is a good addition. There's one thing to fix before merging, plus some naming and wording changes needed to align with the rest of the codebase.

1. **A long summary wraps out of the header box.** The new `<text>` has no width budget, and OpenTUI wraps rather than clips, so a summary wider than the preview wraps to two or three lines inside the fixed `height={5}` box and evicts the rows below it. I rendered it at 100 cols with the default 40% preview and a 92-char summary: the cwd line disappears and the metadata line overprints the third summary line:

   ```
   │ myapp                          ● idle
   │ Investigate why the flaky auth
   │ integration tests fail
   │ myapp:maino·ev1.2.3ss all suites on
   │ ─────────────────────────────────────
   ```

   This is reachable at ordinary widths (a 120-col terminal under the 80% popup leaves ~35 columns, and your own example title is 34 chars), and Cursor's `summaryTitle` rule passes the whole pane title through, so there's no upper bound on the string. Let's pre-truncate the way the other dialogs do. `separatorWidth()` is already the header's inner width, so it's a one-liner (suggestion inline).

2. **Let's call this the session summary, not the session title.** The field is `summary`, the list column is `summary`, and what the agents publish is their own generated one-line summary of what the session is doing. The pane title is just how it reaches us. "Session title" reads as a new field when it's the existing one in a new place. Suggest renaming in the PR title, the doc comment, and the README sentence:

   `feat(preview): Show the session summary in the preview header`

   While you're in the doc comment, the `OC | <title>` example never renders: OpenCode's rule captures group 1 and the daemon returns the capture, so the prefix is stripped before the TUI sees it.

3. **README:** "the rest fall back to the prompt" describes the list's summary column, not the preview header. In the header the line is omitted (suggestion inline).

4. **Tests:** let's have all three new tests assert the separator sits at `metadataLine + 1`. That pins the 4-row / 5-row budget directly, and it would have caught the wrap in (1). Minor note on the late-arriving test: it passes with the wrapper `<box>` removed too (I checked, 28/28), so it doesn't currently protect the insertion-anchor guard the description credits it with. Fine to keep the wrapper, it matches `BackgroundPeek`, but the test name and the PR description overstate what it checks.

florian-schneider-cohere:
Thanks for the careful review — all four points addressed in 1bdd48a (rebased on v1.4.0):

1. **Wrapping**: the summary now pre-truncates to `separatorWidth()` via `truncateText`, per your inline suggestion. Added a regression test (92-char summary at a 30%-wide preview): it asserts the tail renders nowhere and the rows below keep their slots — confirmed it fails without the fix.
2. **Naming**: renamed to "session summary" in the commit/PR title, the doc comment, and the test names. Dropped the `OC | ` example from the doc comment — you're right, the daemon returns the capture.
3. **README**: applied your suggested sentence ("on the rest the line is omitted").
4. **Tests**: all three (plus the new truncation test) now assert the separator sits at `metadataLine + 1`, with `tmuxTarget: "dev:1"` in the fixtures so the metadata row is locatable. Renamed the late-arriving test to "renders a late-arriving summary in the header" and corrected the PR description — it pins the layout but doesn't distinguish wrapper vs bare `<Show>`; the wrapper stays for the `BackgroundPeek` reason.

One note: force-pushed over the "Update branch" merge — the branch is now rebased on `b2a7057` (v1.4.0) with the same content, so the diff is just the three files.

epilande (approved):
LGTM! Thanks!

epilande (approved):
LGTM! Thanks!

epilande (approved):
LGTM! Thanks!

PR #198, opened 2026-09-11
feat(tui): Surface waiting sessions and improve list readability

epilande:
## Summary

Surfaces waiting sessions in a pinned, oldest-first `needs you` band across every grouping. Sessions appear once and return to their ordinary group when the wait ends. The existing row information remains visible: checkout and branch, status and completion cues, invoke outcomes, task text, and PR badges.

- Brightens task text and adds configurable idle-age fading (`ageFade.after`, default 24 hours; `0` disables). Status, agent, PR, and handoff colors retain their meaning.
- Shows local worktree counts on unambiguous project headers, without `gh` or prune classification. Collapsed activity counts take priority when space is tight.
- Adds the attention band to the sidebar, including pane targets for waiting sessions.
- Preserves responsive column modes and icon styles. `b` changes grouping without suppressing row identity; tmux names shorten only under session/window headers.

## Breaking changes

None. Existing keys, column modes, and icon styles remain supported.

## Testing

- [x] Full suite: 5,918 pass, 0 fail, 204 files (commit signing disabled for git fixtures)
- [x] Typecheck, build, unused-code checks, and diff whitespace checks
- [x] Detached picker at 160×45 and 96×30; sidebar at 30×30
- [x] All five grouping modes, invoke outcomes, background/completed rows, shared branches, badges, and collapse/expand captured with fixture data
- [x] Actual CLI rendered against the existing main daemon at all three sizes; daemon left running unchanged

The updated 34-second walkthrough is a surface demo using fixture data, not newly driven real-agent transitions. The original attention-band implementation was also exercised with a real Claude permission prompt.

## Stack

PR 2 (`tui-views`) is stacked on this branch. **Retarget PR 2 to `main` BEFORE merging PR 1**: this repository auto-deletes merged branches, which would otherwise cause GitHub to close PR 2. Neither PR is being merged as part of this update.

## Summary by CodeRabbit

- **New Features**
  - Added a pinned **“needs you”** section for waiting sessions, sorted by age.
  - Preserved project paths, branches, worktree indicators, and pane addresses across grouping modes.
  - Added responsive status badges for completion and invocation outcomes.
  - Added configurable idle-row fading via `ageFade.after` (default: 24 hours; set to 0 to disable).
  - Added repository worktree counts and improved fitting for narrow layouts.
  - Waiting sessions remain visible, with selected groups expanding when needed.

- **Documentation**
  - Updated guidance for session identity, status badges, waiting sessions, worktrees, and idle-row fading.

- **Tests**
  - Expanded coverage for redesigned session lists and configuration behavior.


epilande:
End-to-end verification at 9362c1a. Screen recording of the redesigned rows on `ccmux picker`, filtered to a throwaway repo with one main checkout and two worktrees. It shows the one-cell status glyph, the shared branch identity slot with the worktree `+` marker, the `main + 2 worktrees` header facts, and the `needs you` band lifting a session out of its group on a Bash permission prompt then folding it back on reject. Playback is sped up.

https://github.com/user-attachments/assets/4f212e7c-5033-4540-af51-78d35323ee50

Verdict: PASS.

| Tier | Result |
| --- | --- |
| Static | `typecheck`, `knip`, `build` green; `bun test` 5902 pass / 0 fail (one non-reproducing failure on the first of three runs) |
| TUI render | Picker at 160x45 and 96x30, sidebar at 30, against the live 28-session roster: 1-cell glyph, aligned identity slot, header facts, age fade, separators at `width - 5` with no bleed |
| Detection | Recorded driven run, Claude, hooks mode, idle to working to waiting to idle |
| CLI / HTTP | Read-only smoke on `/health`, `/sessions`, `/worktrees`, `ccmux show --json` |
| Config | `ageFade.after` rejects non-numeric and negative values with exit 1, writes `{"ageFade":{"after":48}}` on the positive path |

Also checked: the `needs you` header refuses to collapse on both `space` and `h` while a control group collapses normally; the waiting row gains its third line in the sidebar; the row is moved into the band, not duplicated.

Not exercised: the `/worktrees` `localOnly` trim on a branch daemon. The response projection carries no PR fields, so it is a wire-compatible change that only drops a `gh`-spawning cache warm. Verified by inspection and typecheck.

Observation, low: a detached checkout surfaces as `.claude (1)   HEAD` in the header, because the shared-branch suppression only filters `main`.

Representative run on Claude; this change affects all agents.

epilande:
Verified live against Claude Code 2.1.278 under tmux at 77e50a5, on an isolated daemon built from the branch over two throwaway repos, so no real sessions appear.

Screen recording, sped up. A real permission prompt drives one session to `waiting`: the row moves into the `needs you` band at the top and `repoa` drops from (2) to (1) while keeping its `main + 2 worktrees` fact. On the band header, `m` opens a menu with no Kill group and `x` does nothing. `J` on the waiting row moves its home group below `repob`; `K` moves it back. Rejecting the prompt empties the band and the row returns to `repoa (2)`. With the picker still open, a `git worktree add` for a third checkout leaves the header at `main + 2 worktrees` until the 30s refresh lands and it reads `main + 3 worktrees`.

Replaces the earlier recording from 9362c1a, which showed the one-cell glyph rows reversed in f70bfc0. Not shown: the sidebar band and the 24-hour idle fade.

https://github.com/user-attachments/assets/7251f3ff-03e8-4211-9963-9877160367e6

epilande:
Band fix in 4299447, rendered in the picker from fixture data at 110x24 with grouping off.

With grouping off, the rows after the `needs you` band used to run straight on from it, so `working`, `done`, and `idle` sessions read like members of the band. Now `▼ sessions (8)` closes the band after the two `waiting` rows. `x`, the row menu, and collapse do nothing on that header, same as on the band header, and nothing is written to state. Grouped modes are unchanged because the next group header already ends the band.

Before, then after:

Before: rows after the needs-you band run on without a header

After: a sessions (8) header ends the band

PR #205, opened 2026-09-17
fix(daemon): Treat static ✳ pane titles as unknown

### LINKED ISSUE #204: [Bug]: Multiple false "Finished" notifications per Claude Code turn under tmux (static ✳ title since Claude Code 2.1.236)
### Preflight

- [x] I searched existing issues and this has not already been reported.
- [x] I am on the latest release of ccmux (1.4.1) and confirmed the bug still happens there.

### ccmux version

1.4.1 (first noticed on 1.3.0; `src/daemon/pane-classify.ts` and the native-Claude stale check are identical between the two)

### Agent

Claude Code 2.1.273, macOS, inside tmux

### Are hooks installed?

Yes (`ccmux setup`; SessionStart / SessionEnd / Notification hooks, all marker PIDs alive)

### What happened?

A single Claude Code turn produces **several "Finished" notifications while Claude is still working**, and only the last one corresponds to the turn actually ending. The earlier ones carry the *user's prompt* as the body (e.g. "可以, 继续") instead of Claude's answer, because at that moment there is no assistant text yet.

What I expect: exactly one "Finished" per turn, delivered when the turn has really completed.

### Root cause (traced through the source)

1. Claude sessions are `native`-tracked. Status comes from the JSONL; when the JSONL has been silent for more than `PANE_IDLE_THRESHOLD_MS` (30 s), `resolveNativeClaudeStates` calls `detectPaneState` on the pane.
2. `classifyPaneTitle` only recognises a Braille spinner (U+2800–U+28FF) as "working". A leading `✳` is "not_working", which falls through to `classifyPaneContent`; when no waiting/plan pattern matches, the result is `active`, and `active && wasWorking` is written as **idle**.
3. Claude Code no longer produces the title ccmux is looking for:
   - 2.1.228 changed the busy-title glyphs from Braille to `◐`/`◑` ("Updated terminal title busy-spinner glyphs to reduce tab-bar jitter").
   - 2.1.236 stopped animating the title under a multiplexer entirely ("Fixed terminal tab titles jumping in tmux: the title is now written only when its text changes instead of animating every 960ms"). In the bundle this is `tengu_static_title_under_mux`, default on, gated on `TMUX`/`STY`/`ZELLIJ`. Under tmux the pane title is a **static** `✳ <summary>` for the whole turn. I sampled `#{pane_title}` 30× at 100 ms while Claude was mid-turn: 30/30 were `✳ …`.
4. So every JSONL gap > 30 s (long thinking, a long Bash/test/push command) flips the session `working → idle`, `decideNotification` yields `finished`, and after `delayMs` it fires. The 60 s cooldown then lets it fire again on the next gap. The finished body falls back to `lastPrompt` because `buildFinishedContext` finds no assistant text after the prompt.

Real-world gaps inside one turn from the JSONL that produced the screenshot (`assistant`/`user` entries after the prompt): 89 s, 52 s, 132 s, 40 s. Each one became a "Finished".

### Why the obvious knobs don't help

- `notifications.delayMs` only filters gaps shorter than `30 s + delayMs`, and delays genuine notifications by the same amount. It can't cover 90–130 s gaps without making real notifications useless.
- The hook marker only ever carries `idle` / `waiting_permission` (Notification hook), so it cannot assert "working".

### Suggested fixes

Any one of these would restore "notify once, when the turn is really over":

- **Use a positive end-of-turn signal instead of inferring it from silence.** Claude Code's `Stop` hook fires exactly when a turn ends and `UserPromptSubmit` when one starts; writing `working` / `idle` into the marker from those two events would make the pane heuristics unnecessary for hooked sessions.
- **Fix the title classifier**: treat `◐`/`◑` as working, and under a multiplexer (static `✳` title) treat the title as *no information* rather than "not working".
- **Detect working from pane content**: while Claude works, the bottom of the pane always shows a spinner-glyph status line such as `· Mustering… (9m 19s · ↓ 23.1k tokens)` (glyphs `·✢✳✶✻✽`, elapsed time in parentheses). A working session should never be downgraded to idle while that line is present.
- At minimum, don't map `active && wasWorking → idle` when the title gives no evidence; leave the session `working` until the log or a hook says otherwise.

### Environment

- ccmux 1.4.1 (Homebrew), daemon restarted after upgrade
- Claude Code 2.1.273 (bug reproducible on any ≥ 2.1.236 under tmux)
- macOS, tmux 3.x, `allow-rename off`
- `notifications: { enabled: true, delayMs: 3000 }`

epilande:
## Overview

Stops false Finished notifications mid-turn under tmux: a static ✳ pane title is no longer treated as idle, and a visible Claude spinner status line keeps the session working through JSONL gaps.

Resolves #204

## Motivation

Claude Code ≥2.1.236 writes a static ✳ title for the whole turn under a multiplexer. ccmux treated that glyph as "not working", and `detectPaneState` mapped active pane content to idle. Any JSONL silence past ~30s then fired Finished, often with `lastPrompt` as the body, while Claude was still going.

### Pane classifier

- **pane-classify.ts** - Title can prove working (Braille, ◐◑◓◒) but never idle. A leading ✳ is unknown. Pane content with a `·✢✳✶✻✽` spinner plus elapsed time in parentheses returns working, so a stale working session is not downgraded while that line is present.
- **pane-classify.test.ts** - Regression for the ✳ + spinner / idle-gap case: static ✳ is unknown, the Mustering status line is working, and `detectPaneState` stays working instead of idle.
- **state-reconciler.test.ts** - Native stale-pane path keeps `working` when detection still says working.

### Docs

- **architecture.md** - Updates the pane-classify concern line to match the new title/content rules.

## Breaking Changes

None

## Testing

- [x] Classifier and reconciler tests pass (`pane-classify`, `state-reconciler`, `index.resolve-working`)
- [x] Full `bun test` (5899 pass, 3 skip, 0 fail) with `commit.gpgsign=false`
- [x] Verified live against Claude Code 2.1.274 under tmux on both tracking arms (recordings in the comments; real pane captures are now test fixtures)

## Notes for reviewers

Stay on the M path from #204. No Stop / UserPromptSubmit hook-marker redesign.

The native reconciler still maps `active && wasWorking` to idle when the pane shows no spinner and no wait. That is the remaining genuine-finish safety net. What it must not do is treat a static ✳ title as that evidence.

<div><a href="https://cursor.com/agents/bc-2a3bf74c-9804-5b24-a4bc-d0d4b476c917?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2a3bf74c-9804-5b24-a4bc-d0d4b476c917&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of Claude’s working state across newer spinner symbols and status lines.
  - Prevented static ✳ symbols from incorrectly marking sessions as not working.
  - Correctly preserves working sessions when an active pane shows a valid working indicator.
  - Improved fallback inspection when pane titles do not provide reliable state information.
  - Correctly transitions stale working or waiting sessions to idle when pane content is active.

- **Documentation**
  - Updated architecture guidance for current pane state indicators.


epilande:
Verified live against Claude Code 2.1.274 under tmux on both tracking arms, at 6b18250.

**Hooked (native) arm, production daemon.** The row holds `working` through 106s of JSONL silence while the pane title stays the stale static `✳ Scan CoreServices directory` from the previous turn, then flips to `idle` exactly once, 5s after `⏺ done`.

https://github.com/user-attachments/assets/1f02a3ce-d17d-4384-bee7-e76e22bb9484

**No-hooks (pane-tracked) arm, isolated daemon.** `working` through 108s of silence, `idle` 3s after the turn ends (the regression the review caught: this arm had lost its only idle source), then a permission prompt in manual mode goes `waiting` and clears to `idle` 14s after approval.

https://github.com/user-attachments/assets/47a24bd7-875f-4428-ac05-a9faf2c974d7

Poll logs sampled `ccmux show --json` every 5s alongside both recordings. Transitions were idle → working → idle (A) and working → idle → waiting → idle (B), with no flapping.

PR #199, opened 2026-09-12
feat(tui)!: add views, scope, and consistent keys

epilande:
Adds Sessions, Worktrees and Start views with shared repo scope and consistent keys. Help comes from one action registry.

A daemon cache supplies lightweight local worktree inventory, exact source totals and PR check/review state. Timer refreshes do not scan dirty files in every checkout. Never-successful GitHub lookups stay absent; failed refreshes retain successful facts with a stale marker. Branch PR associations use the upstream repository and ref, with exact local-tip matching when upstream identity is unavailable; local-ahead commits and the repository list cap do not break proven associations. Header reads never classify worktree removal.

Session rows retain PR 1’s corrected identity and status information, with needs-you pinned in every grouping. Worktrees and Start keep one-line rows with compact metadata. Removal consequences wrap during confirmation, including in the narrow sidebar. The sidebar keeps compact overlays without a strip. Per-view previews and the palette remain PR 3.

Confirmed session actions use a frozen set of visible targets. Scope changes reconcile hidden selections, removed IDs lose their marks, and older polls cannot overwrite newer applied facts. Same-view `W`/`N` updates both scope and rows; marked Start targets survive collapsed groups. Panel regression coverage includes asynchronous reads, removal outcomes, copy/browser feedback and modal guards. See [the regression audit](https://github.com/epilande/ccmux/blob/tui-views/docs/tui-views-regression-audit.md).

## Breaking Changes

`h`/`l` cycle views; `W`/`N` jump narrowed to the cursor’s repo. Space marks, `r` restarts, `R` refreshes, and `d`/`D` review working-tree/branch changes. Dirty deletion is a question during confirmation. See [the key migration note](https://github.com/epilande/ccmux/blob/tui-views/docs/tui-views-migration.md).

## Demo

[CodeRabbit follow-up MP4](https://github.com/epilande/ccmux/pull/199#issuecomment-5644867508): help-prefix handling and returning to Sessions in the cursor repo. Surface demo (shown, not driven).

[Watch the correction MP4](https://github.com/epilande/ccmux/pull/199#issuecomment-5644708687): repo scope, same-view W/N and removal confirmation. [Earlier view/grouping demo](https://github.com/epilande/ccmux/pull/199#issuecomment-5644326962). Both are surface demos (shown, not driven).

## Testing

[...]

PR #206, opened 2026-09-20
feat(tui): column header, titled-rule strip, and one-line group headers

epilande:
## Overview

Redesigns the top structure of the picker: the view strip becomes a titled rule, an optional column-header line labels the row columns, and group headers carry their own rule instead of a separate divider row.

Stacked on #199 (base `tui-views`).

## Motivation

- It is not obvious what data the picker's columns show, and since columns are user-configurable the answer differs per setup.
- A plain header row read as another data row, and the filled tab chip competed with it for attention.
- Every non-first group cost a divider row plus a label row, and vertical space is the picker's scarcest resource.

## Screenshots

Live sessions at 150 columns, before (this branch's base) and after.

**Flat list.** The strip becomes a titled rule and the column header is on by default for the `none` grouping.

Flat list, before and after

**Grouped list.** Each group header carries its own rule; the divider row above later groups is gone. The header is off by default here (`auto`), and `ccmux config set columnHeader always` brings it back.

Grouped list, before and after, plus a collapsed group

**Worktrees view.** Same header component, rule sized from the scrollbox viewport and drawn flush with the rows.

Worktrees view, before and after

**100 columns.** Labels that would not fit their cell are left blank rather than truncated.

Flat list at 100 columns

### View strip

- **ViewStrip.tsx** - Sets the view names into a border-colored rule. The active view gets a blue `━━` stub and bold text, counts ride after the name as a dim `·N`, and the scope sits at the right behind a `scope:` prefix. Clips instead of wrapping below its natural width
- **ViewStrip.test.tsx** - Rewritten for the rule layout, including the scope budget at narrow widths

### Column header

- **preferences.ts** - `ColumnHeaderMode` (`auto` | `always` | `never`) and its default
- **config.ts** - Registers `columnHeader` so `ccmux config set columnHeader always` works
- **session-columns.ts** - `columnHeaderCells` projects labels from the same resolved layout the rows use, so only fixed-position cells get one and user column config is honored; `showColumnHeader` resolves `auto` to the flat `none` grouping
- **SessionList.tsx** - Renders the lowercase, overlay-colored head above the scrollbox and pads it by the measured viewport inset so its cells end on the rows' columns with or without a scrollbar. Hidden below the md breakpoint and in the sidebar
- **App.tsx**, **store.ts**, **index.tsx**, **picker.ts** - Thread the preference through
- **README.md** - Config table row and a Column header section

### Group headers

- **GroupHeader.tsx** - Fills the rest of the line with a rule, dropped when the label leaves no room. New `flushRight` option for lists whose rows have no right padding
- **SessionList.tsx** - Removes the divider row and sizes the header from the scrollbox viewport
- **grouping.ts** - Every header is one visual line; the row-menu anchor loses its divider offset
- **WorktreesPanel.tsx** - Sizes its headers from the scrollbox viewport, flush right
- **architecture.md** - Describes the strip, the head, and the group rule

## Breaking Changes

None

## Testing

- [x] `bun test` (5802 pass) and `bun run typecheck`
- [x] Picker at 150, 100, and 80 columns, flat and grouped, header selected and collapsed, with and without a scrollbar
- [x] Sidebar at 45 columns, and the Worktrees panel in both the picker view and the sidebar overlay
- [ ] Start view repo headers (share `GroupHeader`, not rendered: `gh` is missing from the capture environment)


epilande:
Merged the two fixes from below (the `sessions` header that ends the band in #198, the strip status cues in #199) up into f74e992 and moved the cues into this branch's titled rule. Rendered in the picker from fixture data at 110x24 with grouping off.

The cues sit at the rule's right end, before `scope:`: `▲ degraded · 1 invoking · scope: all repos`. Pressing `f` shows `Sessions ·8/10` and a dim `active`. With the daemon killed, a yellow `● reconnecting` joins them. A healthy daemon leaves the rule exactly as this PR drew it. Under the column header, `needs you (2)` and `sessions (8)` are both one-line headers with their own rule and no divider row between them. When space runs short, the scope name clips first, then `invoking`, then `active`; the connection and degraded glyphs always stay.

Degraded and 1 invoking inside the titled rule, sessions header ends the band

After f: Sessions ·8/10 and active

Daemon killed: reconnecting

Healthy: strip unchanged

